PR #3, opened 2019-05-24
Feature flags

pmdartus:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/feature-flag/text/0111-feature-flags.md)

Originally posted by @caridy at https://github.com/salesforce/lwc/pull/1197

### DIFF
--- a/text/0111-feature-flags.md
+++ b/text/0111-feature-flags.md
@@ -1,0 +1,164 @@
+# RFC: Feature Flag
+
+## Status
+
+- Start Date: 2019-03-01
+- RFC PR: TBD
+
+## Summary
+
+This RFC defines the infrastructure pieces to support feature flag to enable experimentation and progressive development inside LWC core packages.
+
+## Motivation
+
+* It is becoming more difficult to add new features to engine and compiler due to the potential breaking hazard.
+* Some features need to period of adaptation and testing before we can allow users-of-LWC to consume them, but today the only option is to keep them in a separate branch forever.
+* Performance implications of a change is sometimes difficult to assess without hitting production servers, but at that point, it is too late since everyone has access to the same set of features today.
+* Branching code logic for certain features (e.g.: slotchange event) is becoming more complicated over time, and having that in the actual code makes the code less readable.
+
+## Goals of this proposal
+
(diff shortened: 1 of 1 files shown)


text/0111-feature-flags.md
```diff
@@ -0,0 +1,167 @@
...
+It is also important to notice that only features that are configured as `null` on the LWC multi-package repository can be enabled or disabled on the client side. If the feature is marked as `true`, it will be always present independently of the `LWC_config` value, while those marked as `false` are just not present at runtime, which means attempting to configure those via `LWC_config` will do nothing.
+
+#### Testing
+
+During the test phase, all features will be compiled out as runtime optional (equivalent to setting them to `null` value). This guarantees that we can have different tests for the different branching logic. With that in mind, you can use a test helper function to enable certain features per test, e.g.:
+
+```js
+import { enableFeature } from '@salesforce/lwc-test-support'; // bikeshed required here
+
+enableFeature('foo'); // enabling feature foo for this test (we don't need a way to disable them)
```
caridy: I suspect this is only useful if we never use a feature flag during the initialization of the engine, e.g.:

```js
const something = FeatureFoo ? () => 1 : () => 2;
```

but it could be used as:

```js
const something = () => FeatureFoo ? 1 : 2;
```

that's a small detail, but probably should be mentioned, otherwise you have no way to test that feature flag.

text/0111-feature-flags.md
```diff
@@ -0,0 +1,188 @@
...
+Code behind these flags can be conditionally enabled (or completely removed) based on your project configuration. This allows newly developed features to be selectively released when the LWC platform considers them ready for production use.
+
+### Flagging Details
+
+The flag status in the generated build is controlled by the `@lwc/features` package. This package exports a list of all available features and their current status.
+
+A feature can have one of a three flags:
+
+* `true` - The feature is present and enabled: the code behind the flag is always enabled in the generated build.
+* `null` - The feature is present but disabled in the build output. It must be enabled at runtime.
```
tariqrafique: @Gr8Gatsby this is the feature flag option that could work for us. the code is deployed to the browser in a disabled manner and is available to toggle at runtime

PR #4, opened 2019-05-24
Track reform

pmdartus:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/track-reform/text/0110-track-reform.md)

Originally posted by @caridy at salesforce/lwc#1216

### DIFF
--- a/text/0110-track-reform.md
+++ b/text/0110-track-reform.md
@@ -3,7 +3,7 @@
 ## Status
 
 - Start Date: 2019-05-08
-- RFC PR: TBD
+- RFC PR: https://github.com/salesforce/lwc-rfcs/pull/4
 
 ## Summary
 
--- a/text/0110-track-reform.md
+++ b/text/0110-track-reform.md
@@ -1,0 +1,67 @@
+# RFC: Track Reform
+
+## Status
+
+- Start Date: 2019-05-08
+- RFC PR: TBD
+
+## Summary
+
+This RFC describes the reform for the tracking mechanism and the `@track` decorator.
+
+## Goal
+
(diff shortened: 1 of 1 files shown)


mpsenn:
I believe this RFC to be incorrect. This is a breaking change, because currently components can set non-tracked fields in getter methods. But after this change those getters will be breaking read-only invariants like this one:
https://github.com/salesforce/lwc/blob/f6844f28894b8bad1336dd1753314e41b1737ac5/packages/%40lwc/engine/src/framework/decorators/track.ts#L77

caridy:
@mpsenn the tracked fields will remain the same... and this invariant will work the same, while for non-tracked fields, the values mutated during rendering might have issues. I believe your assumption is correct. Specifically, about this part: https://github.com/salesforce/lwc/blob/f6844f28894b8bad1336dd1753314e41b1737ac5/packages/%40lwc/engine/src/framework/observed-fields.ts#L39-L44

@jodarove can you double check this? Maybe the solution here is to not apply such restrictive invariant, and instead just skip the rehydration mark if needed. We can brainstorm and update this.

jodarove:
@caridy we already addressed @mpsenn concern. Just to verify, @mpsenn you are referring to this case right?

```html
<template>
     {computedProp}
</template>
```

```js
class Cmp extends LightningElement {
   foo;

  get computedProp() {
    if (this.foo === undefined) {
        this.foo = 'some calculated value';
    }

    return this.foo;
  }
}
```

mpsenn:
Correct.

caridy:
Just to clarify, the mutation will be tracked there during the rendering phase, but since you're in dirty state, it doesn't imply another rendering, it just does nothing more. I hope that this is now clear.

PR #5, opened 2019-06-14
chore: update if-then-else branching

ekashida:


text/0111-feature-flags.md
```diff
@@ -69,38 +69,39 @@ Some changes do not require an RFC:
...
 ```
 
 which gets compiled to one of the following options:
 
 ```js
-import { feature } from "@lwc/config";
-{   // foo feature is configured as `true`
+import { ENABLE_FOO } from "@lwc/config";
+// foo feature is configured as `true`
+{ // this block is needed in case the if-condition declares some binding
```
caridy: ... the then block declares some ...

PR #6, opened 2019-06-14
Context Service in LWC

caridy:
* [x] api
* [x] examples
* [x] implementation details

### DIFF
--- a/text/0000-context-service.md
+++ b/text/0000-context-service.md
@@ -1,0 +1,200 @@
+- Start Date: 2019-06-01
+- RFC PR: (leave this empty)
+- Lightning Web Component Issue: (leave this empty)
+
+# Summary
+
+Context provides a way to pass data through the component tree without having to pass props down manually at every level.
+
+# Motivation
+
+In a web component, data is passed top-down (host to child elements on its shadowRoot) via attributes, but this can be cumbersome for certain types of attributes (e.g. UI theme) that are required by many components within an application. Context provides a way to share values like these between components without having to explicitly pass a prop through every level of the tree.
+
+* *Common use case*: receiving values from a component higher “up” the DOM tree
+when you don't own the components in between.
+* A context allows a provider component to provision a stream of immutable
+values to multiple components “lower” in the DOM.
+
+# Goals
+
+The primary goal of this RFC is to define how the to consume context, guarantee the right ergonomics and semantics for the consumer code while the provider can be experimental APIs.
+
+As a secondary goal is to provide ways for the lightning platform to control who can provide new context values, as a way to gate this feature.
+
+A third goal is to support the provision of context values on any element, whether it is LWC or not.
+
+# Prior Art
+
+* React Context: https://reactjs.org/docs/context.html
+
+# Detailed design
+
+Important Notes:
+
+* It is important for the consumer to be able to consume more than one context value
+from different context providers if needed.
+* It is important to guarantee that the provisioned object is reactive, so the UI can
+rely on it directly.
+
+## Proposal: Context Consumer
+
+This proposal provides a high-level API (an abstraction layer) for authors to consume a context value at any level in the DOM flatten tree. This must be a future-proof API, and must be easy to use and easy to maintain. Here is an example:
+
+```js
+import { LightningElement } from 'lwc';
+import { XFooContext } from 'x/fooContext';
+import { XBarContext } from 'x/barContext';
+
+export default class MyComponent extends LightningElement {
+    @wire(XFooContext) a; // context object provided by an element up in the flatten tree
+    @wire(XBarContext) b; // context object provided by element up in the flatten tree
+}
+```
+
+Pros:
+* it could be statically verified when wired to a field.
+* it works nicely with inheritance since it relies on field declarations.
+* it is extremely familiar since it relies on the `wire` decorator.
+
+Cons:
+* the wire decorator is often seen as more complicated, and often requires authors to be defensive in case the value is never resolved (which is not the case of for context).
+* it requires changes in the wire protocol to expand its capabilities to support wiring based on flatten DOM tree. (more on this below)
+
+_Note: This is very similar to how React does it, but extending it to access multiple contexts._
+
+_Note: since `@wire` supports wiring to a method, it will work the same way for context._
+
+## Proposal: Context Provider
+
+This proposal provides a low-level API for authors to provide a context value at any level in the flatten DOM tree that can be consume by any node in the subtree. This API is subject to changes as we understand more and more about how context is used.
+
+### Defining a new Context
+
+The definition of a new context is equivalent to create a new wire adapter. It has an identity, and it has some internal implementation details that are necessary to work as the first argument of the `@wire` decorator.
+
+This API must be experimental for the MVP:
+
+```js
+import { createContextExperimental } from 'lwc';
+
+export const XFooContext = createContextExperimental({
+    x: 0,
+    y: 0,
+});
+```
+
+Where the first argument to `createContextExperimental()` function receives the default payload to be provided to any node consuming this new context.
+
+_Note: we could add a second argument for a context value validation routine if needed._
+
+The consumer of this context is as described in the previous section of this document for `XFooContext`.
+
+## Providing a custom value for a context
+
+Setting the value of a context already defined must be an operation involving an element. Any element in the  flatten DOM subtree of that element or its shadowRoot, can consume the value set for a context.
+
+This API must be experimental for the MVP:
+
+```js
+import { setContextExperimental } from 'lwc';
+import { XFooContext } from 'x/fooContext';
+
+setContextExperimental(elm, XFooContext, {
+    x: 1,
+    y: 2,
+});
+```
+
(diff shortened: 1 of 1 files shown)


text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+
+# Adoption strategy
+
+This is a brand new feature, we just need to document it.
+
+# How we teach this
+
+* The consumer aspect of it is very straight forward considering that we have plenty of documentation about the `@wire` decorator.
+
+# Unresolved questions
```
kevinv11n: A couple more:
- Is the value always synchronous provisioned during the `wiring` stage of the component lifecycle?  
- Is there a formal error state that the wire can provision? Will it participate in the component error boundary?

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+
+* Should the context value be validated?
+* Should the context value be type any? or must be an object?
+* Should the context value be serializable to support locker?
```
kevinv11n: On Lightning Platform we've agreed to only provision inert data via a wire.

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+Where `<c-consumer>` and `<c-other-consumer>` or any of their descendent will be able to consume the context value by using the `@wire(XFooContext.Consumer)` decorator, e.g.:
+
+```js
+import { LightningElement } from 'lwc';
+import XFooContext from 'x/fooContext';
+
+export default class MyComponent extends LightningElement {
+    @wire(XFooContext.Consumer) foo; // context object provided by <x-foo-context> element
```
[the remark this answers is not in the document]
caridy: That is the whole point, we are not opinionated on how you share the adapter, it is up to you to decide how. In this case, `XFooContext` is a constructor for `<x-foo-context>` element that provisions the theme value, and `.Consumer` is the adapter that will be wired to the nearest `<x-foo-context>` element.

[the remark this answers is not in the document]
caridy: also, this is to demonstrate that with this infrastructure you can implement the exact same thing you do in react.

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+
+export default class MyComponent extends LightningElement {
+    @wire(XFooContext) a; // context object provided by an element up in the flatten tree
+    @wire(XBarContext) b; // context object provided by element up in the flatten tree
+}
+```
+
+Pros:
+* it could be statically verified when wired to a field.
+* it works nicely with inheritance since it relies on field declarations.
```
sfdciuie:
```suggestion
* it works well with inheritance since it relies on field declarations.
```

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+* it is extremely familiar since it relies on the `wire` decorator.
```
sfdciuie:
```suggestion
* it is familiar since it relies on the `wire` decorator.
```

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+* the wire decorator is often seen as more complicated, and often requires authors to be defensive in case the value is never resolved (which is not the case of for context).
+* it requires changes in the wire protocol to expand its capabilities to support wiring based on flatten DOM tree. (more on this below)
+
+_Note: This is very similar to how React does it, but extending it to access multiple contexts._
+
+_Note: since `@wire` supports wiring to a method, it will work the same way for context._
+
+## Proposal: Context Provider
+
+This proposal provides a low-level API for authors to provide a context value at any level in the flattened DOM tree that can be consume by any node in the subtree. This API is subject to changes as we understand more and more about how context is used.
```
sfdciuie:
```suggestion
This proposal provides a low-level API for authors to provide a context value at any level in the flattened DOM tree that can be consumed by any node in the subtree. This API is subject to change as we understand more and more about how context is used.
```

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Defining a new Context
+
+The definition of a new context is equivalent to create a new wire adapter. It has an identity, and it has some internal implementation details that are necessary to work as the first argument of the `@wire` decorator.
```
sfdciuie:
```suggestion
The definition of a new context is equivalent to creating a new wire adapter. It has an identity, and it has some internal implementation details that are necessary to work as the first argument of the `@wire` decorator.
```

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+    <x-foo-context theme="dark">
+        <c-consumer></c-consumer> 
+    </x-foo-context>
+    <x-foo-context value="lite">
+        <c-other-consumer></c-other-consumer> 
+    </x-foo-context>    
+</template>
+```
+
+Where `<c-consumer>` and `<c-other-consumer>` or any of their descendent will be able to consume the context value by using the `@wire(XFooContext.Consumer)` decorator, e.g.:
```
sfdciuie:
```suggestion
Where `<c-consumer>` and `<c-other-consumer>` or any of their descendants will be able to consume the context value by using the `@wire(XFooContext.Consumer)` decorator, e.g.:
```

text/0000-context-service.md
```diff
@@ -0,0 +1,200 @@
...
+import XFooContext from 'x/fooContext';
+
+export default class MyComponent extends LightningElement {
+    @wire(XFooContext.Consumer) foo; // context object provided by <x-foo-context> element
+}
+```
+
+# Other Alternatives
+
+* The two new APIs can be imported from `@lwc/context` or `@lwc/wire-service` rather than `lwc` directly. This might help to eventually allow others to use contexts with vanilla web components. It will also make it easier for Lightning Platform to restrict who has access to these APIs in platform.
```
sfdciuie:
```suggestion
* The two new APIs can be imported from `@lwc/context` or `@lwc/wire-service` rather than `lwc` directly. This might help to eventually allow others to use contexts with vanilla web components. It will also make it easier for Lightning Platform to restrict who has access to these APIs on platform.
```

text/0000-context-service.md
caridy: added

caridy:
@kevinv11n @pmdartus @sfdciuie @davidturissini this is ready for another pass. I think have covered almost everything now, including implementation details.

text/0000-context-service.md
```diff
@@ -0,0 +1,171 @@
...
+        <x-bar></x-bar>
+    </x-theme-context>
+</template>
+```
+
+It means that `<x-bar>` or any of its descendants can wire to `ThemeContext.Provider` to obtain the `theme` value.
+
+## Implementation Details
+
+In order to implement `LightningContext`, it requires a reform on the the wire protocol. This reforms has two folds:
```
jbenallen:
```suggestion
In order to implement `LightningContext`, it requires a reform on the wire protocol. This reform has two folds:
```

text/0000-context-service.md
```diff
@@ -0,0 +1,171 @@
...
+
+1. We need to introduce a new event that can be dispatched by a wire adapter to create a side channel between the `eventTarget` with a context provider installed on an element in the the composed path. The new event is implemented via a new constructor called `LinkContextEvent`, and it expects a unique `token`, and a `callback` to finalize the linking process. E.g.:
```
jbenallen:
```suggestion
1. We need to introduce a new event that can be dispatched by a wire adapter to create a side channel between the `eventTarget` with a context provider installed on an element in the composed path. The new event is implemented via a new constructor called `LinkContextEvent`, and it expects a unique `token`, and a `callback` to finalize the linking process. E.g.:
```

text/0000-context-service.md
```diff
@@ -0,0 +1,171 @@
...
+
+    function callback(data, unsubscribe) {
+        // provider calls this function at any given time
+        eventTarget.dispatchEvent(new ValueChangedEvent(data));
+        unsubscribeCallback = unsubscribe;
+    }
+
+    eventTarget.addEventListener('connect', () => {
+        // attempt to link to a context provider identified by token
+        const event = new LinkContextEvent(token, callback);
```
jbenallen: What is `token`? It's not defined in this example.

caridy: Excellent question. Maybe I should have a section under implementation details for just that.

`token` is the dom event name used to link to the parent element (maybe `token` is the wrong term here). I didn't want to call it `domEventName` because it might be confusing. The point of this event name is that it "should" be tokenized somehow... generated once during the definition of the adapter. In the case of `LightningContext` element, it uses that to create the adapter and to listen for it on the DOM, and it is created per adapter, as a unique event name.

jbenallen: So it's a [DOMString](https://developer.mozilla.org/en-US/docs/Web/API/DOMString) whose value is generated automatically by the system.

caridy: yes, it is a DOMString but whose value must be provided by the author of the adapter, and we don't do much with it... other than dispatching a dom event with a callback whenever the adapter emits a `LinkContextEvent` event.

caridy:
got the thumbs up from @kevinv11n @diervo and @jbenallen

caridy:
This RFC was partially rejected during implementation due to the intersection with Wire Reform (https://github.com/salesforce/lwc-rfcs/pull/14).

PR #7, opened 2019-07-05
first draft for LightningElement.CustomElement abstraction

caridy:


text/0000-custom-element.md
```diff
@@ -0,0 +1,59 @@
...
+1. Provide a dead simple way to register a custom element from a LWC Constructor.
+2. Preserve the web components semantics.
+
+# Proposal
+
+This proposal provides a high-level API (an abstraction layer) for authors to access the custom element associate to a LWC Constructor:
+
+```js
+import XFoo from 'x/foo';
+customElements.define('x-foo', XFoo.CustomElement);
```
caridy: this is the meat of this proposal... the `.CustomElement` static accessor on every LWC constructor

text/0000-custom-element.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. Provide a dead simple way to register a custom element from a LWC Constructor.
+2. Preserve the web components semantics.
+
+# Proposal
+
+This proposal provides a high-level API (an abstraction layer) for authors to access the custom element associate to a LWC Constructor:
+
+```js
+import XFoo from 'x/foo';
```
apapko: is allowing to import 'constructor' property an option?
```js
import { constructor as XFooConstructor } from 'x/foo';
```

text/0000-custom-element.md
```diff
@@ -0,0 +1,59 @@
...
+* it preserves the semantics of web components in a way that you can't bend them.
+
+Cons:
+* it can't be statically analyzed (but since we don't use this in platform, that should be ok).
+* it does not provide a way for you to control whether or not you want open or closed shadowRoot (this is fine because consumers should not be in control)
+
+## Semantics and Invariants
+
+* There is a 1-1 mapping between LWC Constructors and the corresponding Custom Element.
+* Claiming the Custom Element for the abstract `LightningElement` must be disallow.
```
apapko: what does this mean- Claiming the Custom Element for the abstract `LightningElement`?

text/0000-custom-element.md
```diff
@@ -0,0 +1,59 @@
+- Start Date: 2019-07-05
+- RFC PR: (leave this empty)
+- Lightning Web Component Issue: (leave this empty)
+
+# Summary
+
+Facilitate the registration of LWC Constructor as Custom Elements.
```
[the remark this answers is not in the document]
caridy: > registration of LWC Constructor as Custom Elements

text/0000-custom-element.md
pmdartus: The main issue that I have with `CustomElement`, is that the custom element represents a custom element instance while the custom elements registry accepts a constructor. I support @apapko into making this explicit.

text/0000-custom-element.md
```diff
@@ -0,0 +1,59 @@
...
+
+This is a brand new feature, we just need to document it.
+
+# How we teach this
+
+* The consumer aspect of it is very straight forward considering that they don't have to learn a new API, just a static field name in LWC Constructors.
+
+# Unresolved questions
+
+* How can users control the shadowRoot's mode?
```
pmdartus: As you stated above, the consumer should not decide what is the shadowRoot mode. I don't think this statement should be included in the _Unresolved questions_ section.

text/0000-custom-element.md
ekashida: I mentioned a similar concern in https://github.com/salesforce/lwc/pull/1395#pullrequestreview-259747574

text/0000-custom-element.md
jodarove: Agree with you, we should resolve it!

In today web components, you can decide the shadowRoot mode to use per component. In the current implementation we have it per component, passing it to the `buildCustomElementConstructor`.

As of [today implementation](https://github.com/salesforce/lwc/pull/1395) of this RFC, all LWC custom elements shadowRoot mode is [open](https://github.com/salesforce/lwc/pull/1395/files#diff-863e83caa80f932b8cd147e027dcc89bR18). It can't be set to close.

Lets decide:
- Should we enforce all LWC custom elements to be in open or close mode? Should we provide a mechanism on how to set it?
- Do we want an LWC custom element to specify the shadowRoot mode? Ex: via static property on the constructor.

cc/ @caridy

caridy: Ia agreed with @pmdartus this question can be removed from here. It is a decision of the developer or the platform, not the consumer of the component.

pmdartus:
Looking at this RFC again, I am now uncertain about attaching a `CustomElementConstructor` on the `LightningElement.prototype`.

We have started decoupling the LWC engine from the DOM (#23), what would the `CustomElementConstructor` getter returns when the engine is running in an environment without DOM. The current approach with the `buildCustomElementConstructor` method has the following benefits:
- it can be removed from the `@lwc/engine-core` and only exposed from the `@lwc/engine-dom`. 
- it can be tree-shaken if it is not used by the application.

mikebernard:
This would be a huge win for [nCino](https://www.ncino.com/)! We have a library of custom web components that we've built and use in many of our Salesforce applications. It would save us a lot of time and money if we could reuse these components.

gonzalocordero:
@jodarove This was already implemented right? should we change the status then?

jodarove:
thanks @pmdartus!

PR #10, opened 2019-07-09
Dynamic components

diervo:
This RFC introduces a set of principles and invariants required to preserve a reliable and predictable system while allowing lazy loading of dependencies.

## Basic example

```html
<template>
    <x-lazy lwc:dynamic={customConstructor}></x-lazy>
</template>
```

```js
import { LightningElement, track } from "lwc";
export default class DynamicCtor extends LightningElement {
    @track customConstructor;

    connectedCallback() {
        this.loadCtor();
    }

    async loadCtor() {
        const ctor = await import("c/customConstructor");
        this.customCtor = ctor;
    }
}
```


text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+
+By doing **dynamic imports** we are making the dependencies **soft** (not part of the dependency tree), meaning that will be loaded at runtime (a network call may be made). That being said, by forcing some invariants (described below) in the way we use those dynamic imports we could gather a lot of metadata at compile time that will allows us to optimize **when** those resource can be prefetched (see Annex for pivots proposal).
+
+## Guiding Principles
+
+We want to avoid ad-hoc solutions and proprietary non-standard ways to define dynamic component loading. We want to build an API that is future proof, whenever more things land into the spec for modules.
+
+## General Invariants
+
+Note that the invariants defined bellow are not necessarily constraints on LWC in general but rather explicitly defined for the UI Platform use cases:
```
caridy: -> "Lightning Platform use cases"

text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+    - No special handling in user-land is needed.
+- It must work regardless of the algorithm, even if nothing is preloaded at all.
+    - The decision for when a dynamic dependency is loaded is orthogonal to its definition.
+    - We might have a preload mechanism based on MRU or on a specific context (ex.mobile) but the component should work regardless of the algorithm even if nothing is preloaded at all.
+
+You will notice that this follows the same pattern as what we have done for `@wire`: Bring the ability to understand and resolve dependencies at compile, build and design time.
+
+## API
+
+Bellow is the API and its ergonomics for dynamically loading a module. It has very similar mental model to the `is` attribute proposal (which we can't use since Safari will not implement it in the foreseeable future): An arbitrary tag name on which you can assign any constructor.
```
caridy: -> HTML `is` attribute proposal...

text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+
+The biggest draback on this API is its historic abuse in the Lightning Platform: A lot of missuse of dynamic component creation. We will be adding a lot of guiderails to avoid developers shooting themselves in the foot.
+
+# Alternatives
+
+We have consider very early on non-standard ways to import modules, but we belive that will put us in the wrong path in the long term.
+
+# Adoption strategy
+
+This will be first allowed in OSS and depending on feedback and usage, we will be rolling it out on the Salesforce Lightning Platform as we see fit.
```
ekashida:
```suggestion
This feature will initially be implemented in OSS and will be rolled out to the Salesforce Lightning Platform depending on usage and feedback.
```

text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+
+# How we teach this
+
+Dynamic imports is a standard feature (in Stage 4), and we will be documenting `lwc:dynamic` directive along with a lot of documentation explaining the anti-patterns and and when to use it (and more importantly when to not use it).
```
ekashida:
```suggestion
Dynamic imports is a standard feature (in Stage 4). Our documentation around the `lwc:dynamic` directive should provide a comprehensive list of known anti-patterns along with details around when this feature should be used.

File: text/000-dynamic-components.md [hunk 5]
```diff
@@ -0,0 +1,338 @@
...
+
+> If we could snapshot, at a given point in time, the metadata of an app and precompute all server side generated components (if any); then the whole application would be a big monolithic and static component tree, on which routes and every other pivot are simply just if/else branches.
+
+Though idealistic it is conceptually sound (and even technically achievable). It allows us to then **decide when we really need those branches** and when to trim them. Each of those branches would be lazy loaded and dynamically created.
+
+## Proposal by example
+
+Let's imagine we need to build a chart component (`lightning-chart`).
+
+A developer can choose between a bunch of types of charts (lightning-chart-histogram, lightning-chart-bar, lightning-chart-pie, ...), but we only want to expose one and branch based on a type property. We could implement it today like something like this:
```
DNLZ: It may be worth being more explicit here that this scenario and following example are how things would be done pre RFC changes. I got a little confused when reading this for the first time.

btw, I love that you guys are doing these RFCs now!

text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+}
+</script>
+
+<!-- compiled code -->
+<script>
+  import { loader } from "@salesforce/loader";
+  // this can't be imported in user-land
+
+  function tmpl($api, $cmp) {
+     return [$cmp.lazyConstructor != null // falsy for null or undefined
```
DNLZ:
```suggestion
     return [$cmp.lazyConstructor != null ? // falsy for null or undefined
```

DNLZ: it looks like this is supposed to be a ternary but is missing the "`?`".

DNLZ:
Got a general question, but not sure if this is the right place to ask it. Apologies in advance if that is the case.

How will a developer in "user-land" handle the situation where dynamically imported components have a different attribute names?

For example, let's pretend we have a situation where we want to conditionally import one the components with the following structures.

`<c:compOne varOne={bind1} />`

`<c:compTwo varTwo=(bind2} />`

will the component with the `lwc:dynamic` need to have both `varOne` and `varTwo` as attributes?

e.g, `<x-myDynamicComp lwc:dynamic={constructor} varOne={p1} varTwo={p2} />`

Will it result in runtime errors if we try to pass an attribute to a dynamically instantiated component that doesn't have a specified `@api` property for that attribute? Is there a way to dynamically pass just the attributes that are needed for the component we choose to instantiate?

Thanks.

ekashida:
@diervo can correct me if I'm wrong but developers should still be responsible for setting the correct attributes, just as in the hard-dependency approach where you essentially have a big switch. You would just say something like:
```
<template if:true={foo}>
  <x-foo lwc:dynamic={fooConstructor} foo={foo} value={value}></x-foo>
</template>
<template if:true={bar}>
  <x-bar lwc:dynamic={barConstructor} bar={bar} value={value}></x-bar>
</template>
...
```

text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+    <section>
+      <lightning-chart-pie>
+         #shadow-root
+            <!-- fancy dom of lightning-chart-pie goes here -->
+      </lightning-chart-pie>
+    </section>
+</lightning-chart>
+```
+
+Now let's imagine we have a lot of charts, like thirty. The template will become really verbose if we have to do all this conditional logic.
```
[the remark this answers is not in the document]
diervo: Will read

text/000-dynamic-components.md
diervo: yeah that is psudo-code, will fix

caridy:
@DNLZ, as @ekashida explained, that's on the dev so solve it. In fact, this feature is aiming for pivoting/bucketing mechanism, where different constructors for the same element in the template MUST implement the same API, otherwise you will get a runtime error if the provided attribute (even if undefined/null) is not defined. The workaround is what @ekashida showed above, where you control that via IF conditions on top of the new directive.

diervo:
Yes a couple of people already asked for it but that means you are loading an arbitrary component with arbitrary shape, which we don't want to really support (if you want to do that off-platform you might as well call createElement).

You can technically also set your properties programmatically but that is again something that we want to avoid and discourage.

DNLZ:
@diervo @caridy @ekashida - Awesome, thanks for the clarifications!

text/000-dynamic-components.md
```diff
@@ -0,0 +1,338 @@
...
+
+**Dynamic instantiation**: creation of a new component instance where the underlying component constructor is not known until runtime.
+
+**Code splitting:** Capability to split the code base in multiple bundles, that can be loaded on demand
+
+# Basic example
+
+```html
+<template>
+    <x-lazy lwc:dynamic={customConstructor}></x-lazy>
```
jodarove:
```suggestion
    <x-lazy lwc:dynamic={customCtor}></x-lazy>
``` ?

**Comment by rapsacnz** [comment]
I know this will affect the speed of the framework, but there are several use cases that could really use some lazy load love - I have a wizard component written in Aura that I would dearly love to port to LWC, I also have a side scrolling carousel component that would really benefit from dynamism. I know we can use slots, but some cases this is not practical. I'm also fairly sure that with most of the UI static, as long as the number of dynamic components is kept low, performance won't take much of a hit.

File: text/000-dynamic-components.md [hunk 9]
```diff
@@ -0,0 +1,349 @@
...
+export default class DynamicCtor extends LightningElement {
+    @track customCtor;
+
+    connectedCallback() {
+        this.loadCtor();
+    }
+
+    async loadCtor() {
+        const ctor = await import("c/customConstructor");
+        this.customCtor = ctor;
```
KamilSzostak: `this.customCtor = ctor.default;`

KamilSzostak:
```suggestion
        this.customCtor = ctor.default;
```

pmdartus:
@diervo Shall we merge the RFC since the implementation is done?

diervo:
I literally did that on the plane, but apparently failed xD will do today

SuperPluck:
Hey, guys. My apologies if this is not the place to ask, but... 

Now that this has been accepted, what's the expected timeline to see it released?

diervo:
In OSS land you can use it already, in platform is just about prevent bad practices, so it's likely we will be surfacing first for ISVs and unmanaged packages first.
FYI @Gr8Gatsby

LukeKoz:
Is there any word on timelines for this yet?

We have a number of components which have large dependencies we really could do without loading until they are necessary. One basic example is building a wizard that has a number of steps of which all include heavy UI and processes. If we could load these steps as and when they are needed, that would dramatically improve performance and provide a better user experience.

As a side-note, we are an ISV developing managed packages.

Thanks for all your efforts on this! 😄

jordanhenderson:
Any updates on this? It would be a fantastic alternative to manually hardcoding components.

I can also see how internal SF teams could benefit from using this to load flow-embedded LWC into their runtime (I think Aura currently does this but is not publicly documented). We also have a custom flow/wizard style runtime (which interprets and renders Screen Flows) and have to currently hardcode each new custom component; this adds a possible point of failure/redundancy which could be avoided with dynamic LWC inclusion. Would be really great if anyone could provide a possible timeframe for this being released.

Gr8Gatsby:
@jordanhenderson what is the scenario you are looking to solve with dynamic components? Are you trying to distribute in a managed package, or just deploy in c namespace?

jordanhenderson:
@Gr8Gatsby we render flows based on FlowMetadata with our own runtime. As such, screen flows support adding LWCs to any screen - the SF native flow runtime loads these components in with Aura, but our own runtime can't achieve this without the ability to dynamically load LWC. We would probably want to load components from both the c and managed namespace, just like how flows can have any exposed LWC added to screens without hardcoding each.

PR #12, opened 2019-07-25
chore: update with note about initialization usage

ekashida:


text/0111-feature-flags.md
```diff
@@ -69,12 +69,12 @@ Examples of changes that would not require a feature flag:
 
 ### Using features in Code
 
-In your code, you must import bindings from the config package to branch your code. We have two options:
+In your code, you must import bindings from the features package to branch your code. We have two options:
```
apapko: should we stay consistent with the previous paragraph

> In your code, you must import bindings from the @lwc/features package

text/0111-feature-flags.md
```diff
@@ -154,9 +154,30 @@ describe('foo', () => {
 });
 ```
 
+Note that this will only work for features behind flags that are not used during the initialization of the engine.
```
apapko: are there any examples of initialization features that we can list?

text/0111-feature-flags.md
```diff
@@ -154,9 +154,30 @@ describe('foo', () => {
...
+};
+
+// Doesn't work when run as part of initialization
+let foo;
+if (ENABLE_FOO) {
+    foo = () => 1;
+} else {
+    foo = () => 2;
+}
+```
```
apapko: the examples above don't demonstrate their relationship to the engine initialization

text/0111-feature-flags.md
```diff
@@ -154,9 +154,30 @@ describe('foo', () => {
...
+
 ### Extensibility
 
-It is possible that `@lwc/config` can be used beyond this package, so other parts of the platform can implement a similar solution for their own features. This might include components, which means that the LWC compiler will have to understand the transformation as well.
+It is possible that `@lwc/features` can be used beyond this package, so other parts of the platform can implement a similar solution for their own features. This might include components, which means that the LWC compiler will have to understand the transformation as well.
```
apapko: is this a warning? Who's responsible in enduring that other entities understand the import?

ekashida: This is just saying that @lwc/features could be be used outside of our repository so we should keep code reuse in mind.

text/0111-feature-flags.md
ekashida: We don't have anything explicit like what you're asking for, but any code that is initially evaluated when the engine loads are subject to this restriction. This includes variable initializations and function declarations.

text/0111-feature-flags.md
ekashida: The former evaluates `ENABLE_FOO` when the function `foo` is invoked so it'll follow the right path if the flag is toggled during a test. The latter evaluates `ENABLE_FOO` during initialization so by the time we toggle the flag in a test, it's too late in that it won't make a difference.

PR #13, opened 2019-07-31
wip: observable fields

jodarove:


text/0110-track-reform.md
```diff
@@ -21,15 +22,16 @@ Another confusion is when for `@wire` decorator reactivity works fine even when
 
 ## Proposal
 
-### "Unlearn" the `@track` decorator
+### Learn the `@track` decorator when you really need it.
```
caridy: `### Use the...`

text/0110-track-reform.md
```diff
@@ -21,15 +22,16 @@ Another confusion is when for `@wire` decorator reactivity works fine even when
...
 
-We don't have the concept of deprecate, instead it is more accurate to call it "unlearn" because new users don't need to learn about it, and old users can just forget about it. Components that are using that decorator will just simply continue to function, and the compiler will completely ignore (remove) that decorator.
+The runtime/compiler will figure what is the list of fields that are defined by a class, and make the template react to mutations of the field at runtime.
```
caridy: figure -> "extract all fields"

react to mutations of the field at runtime -> observe mutations of any of those fields and update the UI accordingly.

text/0110-track-reform.md
```diff
@@ -21,15 +22,16 @@ Another confusion is when for `@wire` decorator reactivity works fine even when
...
+
+Mutations on fields (ex: `this.x = 5;`) is the majority of the use cases, but sometimes you need a field to hold a value other than a primitive, and make changes in its value ex: `this.address.city = 'San Francisco'`. Only in such cases you will need to use the `@track` decorator.
```
caridy: `is` -> `represent`

caridy: make changes in its value -> mutated, ex:

text/0110-track-reform.md
```diff
@@ -21,15 +22,16 @@ Another confusion is when for `@wire` decorator reactivity works fine even when
...
 
-Instead, the runtime/compiler will figure what is the list of fields that are defined by a class, and make them all trackable at runtime.
 
 ### Compiler changes
 
-When the compiler compiles the class, it can extract any field that is not decorated with `@api`, and pass the metadata through the registerComponent call, this information will be used as the same list of things to be tracked today. Not need to change the logic in the engine, just the error messages if any.
+When the compiler compiles the class, it can extract any field that is not decorated with `@api`, `@track` or `@wire`, and pass the metadata through the registerComponent call. For every class property that is passed to the registerComponent, we will create a getter and a setter in the prototype in order to observe mutations in to class property. No change is needed to the logic in the engine.
```
caridy: class property -> class field

let's not use property/properties that much, because that includes expandos, while fields are more clear.

text/0110-track-reform.md
```diff
@@ -60,8 +62,53 @@ export default class Foo extends LightningElement {
...
 In the example above, calling `foo()` and `bar()` will always render the latest on the next tick, while calling `baz()` might or might not update on the next tick, depending on others mutations in the current job. With this reform, they will always get the latest on the next tick, no matter what.
 
+Also, because of the way is going to be implemented, those non decorated class properties will maintain identity. Example:
+
+```js
+export default class Foo extends LightningElement {
+    _x;
+
+    get x() {
+        this.x = value;
```
caridy: `return this._x;`

text/0110-track-reform.md
```diff
@@ -60,8 +62,53 @@ export default class Foo extends LightningElement {
...
+    }
+
+    set x(value) {
+        this.x = value;
```
caridy: `this._x = value;`

text/0110-track-reform.md
```diff
@@ -60,8 +62,53 @@ export default class Foo extends LightningElement {
 
 In the example above, calling `foo()` and `bar()` will always render the latest on the next tick, while calling `baz()` might or might not update on the next tick, depending on others mutations in the current job. With this reform, they will always get the latest on the next tick, no matter what.
 
+Also, because of the way is going to be implemented, those non decorated class properties will maintain identity. Example:
+
+```js
+export default class Foo extends LightningElement {
```
caridy: probably this example is not really needed, it is weird, and by just explaining that observability of fields is not about proxies, and we don't add the proxy there, just when you `@track`, and the identity will be preserved, should be enough

text/0110-track-reform.md
```diff
@@ -60,8 +62,53 @@ export default class Foo extends LightningElement {
...
+        zipCode: '94102'
+    }
+
+    // this statament mutates the value of the field
+    // and it won't be reflected in the template.
+    // you need to use @track to properly reflect this in the template.
+    this.address.zipCode = '94104';
+```
+
+* Performance improvement: since there is no need to use proxies to observe primitive values (majority of use cases). Suppose we have a base component that is heavily used today (ex: icon) with a tracked property (ex: name, and it receives a string); by removing the track, which is not needed anymore for this field, we will remove the extra computations added by the proxy logic times how many instances of this component on the page.
```
caridy: "extra ..." -> extra checks to try to apply a proxy to the value.

text/0110-track-reform.md
```diff
@@ -60,8 +62,53 @@ export default class Foo extends LightningElement {
...
+
+* Deterministic update of the UI after mutating a field.
```
caridy: probably adding a little bit more here... saying that it is now deterministic because all fields are observed now.

text/0110-track-reform.md
```diff
@@ -60,8 +62,53 @@ export default class Foo extends LightningElement {
...
+
+* No need to ask whether a field should be tracked or not (we get this question a lot). It should come naturally to developers: You start by adding class properties until you mutate the value of a property and you don't see it reflected on the template, then you are going to the documentation and see that in such cases where you need to react to changes in the value of the field, `@track` is your friend.
```
caridy: class fields

caridy: then you are going to the -> then you read the

PR #14, opened 2019-08-20
`@wire()` protocol reform RFC

caridy:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/caridy/wire-reform/text/0000-wire-reform.md)

cc: @jbenallen @kevinv11n @davidturissini @diervo

### DIFF
--- a/text/0000-wire-reform.md
+++ b/text/0000-wire-reform.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-08-01
-- RFC PR: TBD
+- RFC PR: https://github.com/salesforce/lwc-rfcs/pull/14
 - Lightning Web Component Issue: (leave this empty)
 
 # Summary
@@ -35,10 +35,10 @@ This reform is focused on the refactor of the wire decorator code, and the wire 
 
 * To install a prototype descriptor to handle the wired field value from the `vm` of the component.
 * To create an instance of an adapter and link it to the host.
-* To signal to the adapter instance when the component is connected or disconnected.
-* To signal to the adapter instance when the config has changed by providing the new config object.
+* To signal to the adapter instance when the component is connected or disconnected via `connect()` and `disconnect()` APIs.
+* To signal to the adapter instance when the config has changed by providing the new config object via `update()` API.
 * To extract the config value from the host object by relying on the compiler's wire metadata.
-* To signal to the adapter instance when a context is available by providing the new context value. 
+* To signal to the adapter instance when a configured context is available by providing the new context value via `update()` API. 
 
 ## Responsibilities of the wire service
 
@@ -50,12 +50,16 @@ This reform is focused on the refactor of the wire decorator code, and the wire 
 
 * `@lwc/engine` does not know about `@lwc/wire-service` and vice versa.
 * `@lwc/engine` will install a descriptor on the prototype for every wired field during the decorators registration routine.
-* `@wire` decorator will create an instance of the provided `WireAdapter` during the component initialization routine by providing the data callback, and the context callback as arguments. 
+* `@wire` decorator will create an instance of the provided `WireAdapter` during the component initialization routine by providing the data callback as argument.
 * `@wire` decorator will invoke the adapter's `connect` and `disconnect` based on the internal hooks used per component.
 * `@wire` decorator will create a mutation tracking phase to track any access executed during the computation of the config before calling the `update` routine to be able to detect mutations on those values and issue another update on the adapter instance.
-* `@wire` decorator will implement the logic to provide contextual information when requested by the wire adapter.
+* `@wire` decorator will detect if the adapter is expecting contextual information, and if there is a context provider defined to carry on the hand-shaking protocol between the consumer and the provider.
 * `@lwc/compiler` will provide a config function per `@wire()` declaration to produce a new config object when invoked with the `component` as the first argument. The `@wire` adapter can rely on that config function to produce a new config object at will.
 * `@lwc/wire-service` becomes Lightning Platform-specific for the most part (`register()` method) since anyone can implement the wire adapter protocol.
+
+_Backwards Compatibility Notes:_
+
+* As today, the descriptor installed on the prototype of the component by the `@wire` decorator was identical to `track` when decorating a field. This means that the component author could change the value of the field, and such change will be tracked. This is not longer the case, and even though the author can still change the value, it will not be reactive, causing no side effects on the UI of the element. This is a non-backwards compatible change, but we believe this is a very low risk for something that was not working as expected.
 
 ### Wire Adapter Protocol
 
@@ -63,24 +67,57 @@ The formalization of the wire adapter protocol is important because that enables
 
 ```ts
 interface WireAdapter {
-    update(config: Record<string, any>);
-    context(uid: string, value: any);
+    update(config: ConfigValue, context?: ContextValue);
     connect();
     disconnect();
 }
-
 interface WireAdapterConstructor {
-    new (callback: EmitDataCallback, contextualizer?: RequestContextCallback): WireAdapter;
+    new (callback: DataCallback): WireAdapter;
+    configSchema?: Record<string, WireAdapterSchemaValue>;
+    contextSchema?: Record<string, WireAdapterSchemaValue>;
 }
-
-type EmitDataCallback = (value: any) => void;
-
-type RequestContextCallback = (uid: string) => void;
+type DataCallback = (value: any) => void;
+type ConfigValue = Record<string, any>;
+type ContextValue = Record<string, any>;
+type WireAdapterSchemaValue = 'optional' | 'required';
 ```
 
 _Notes:_
 
-* Not all environments will support or need contextualization (e.g.: preloading LDS data), the `requestContextCallback` is optional. All adapters should be able to work with and without a context provider.
+* Not all environments will support or need context (e.g.: preloading LDS data), but does supporting it can rely on the static field called `contextSchema` to provide the context value when available.
+* Some environments might choose to implement validation rules for `configSchema` and `contextSchema` to guarantee compliance.
+* we favored the `DataCallback` over a promised based on `update()` calls because the callback can be invoked sync and async, but most important, because update might never be called by the environment.
+
+### Semantic changes for `@wire` decorator IDL
+
+As today, there are few restrictions and ambiguities with the IDL for the config object in `@wire` decorator declarations. This sections will describe the new semantics, which are almost identical for most cases:
+
+* `$token` can only appear as the value of a top level member property, e.g.: `@wire(foo, { x: '$prop1' })` will continue to be valid, while `@wire(foo, { x: { y: '$prop1' } })` will throw a compiler error, while today it doesn't but the value is never transformed, and remains as a string value. This is a non-backwards compatible change, but we believe this is a very low risk for something that was not working as expected.
+* there will be no identity for inline JSON objects when assigned to a property in the config object, e.g.: `@wire(foo, { x: { y: 1 } })` where the value of `x` will be computed every time, instead of cached per instance or per class as today.
+* there will be identity preserved when assigning a reference values in the config object, e.g.: `@wire(foo, { x: someValue })` where the value of `x` will be a reference to `someValue` during the class declaration.
+* every time that `adapter.update()` is invoked, a new config object will be provided as a first argument, no identity is preserved in this case.
+
+### Context Provider for Wire Adapters
+
+TBD
+
+### Backwards Compatibility
+
+This RFC does introduce minor (or minimal) breaking changes:
+
+* Minor semantic changes for `@wire` decorator IDL as described above.
+* Removal of the experimental `LinkContextEvent` constructor in `@lwc/wire-service` in favor of the new `ContextProvider` API.
+* Minor semantic change on the identity of the first argument passed into `@wire`, it is now a wire adapter instead of a symbol.
+* Removal of LWC's `wire` services via `register()`, which was only needed for `@lwc/wire-services` to plug registered Wire Adapters.
+
+### Forward Compatible Changes
+
+* Lifting the restrictions around manual invocation of Wire Adapters.
+
+### Proposed Restrictions for Lightning Platform
+
+* `ContextProvider` should be gold-filed to prevent proliferation of contextual information in the first release.
+* Preserve the current restrictions to only allow certain wire adapters in `@wire` decorators declarations for one more release.
 
 ## Interop
 
@@ -132,7 +169,21 @@ class Foo extends React.Component {
 
 Since there is a need to support callable adapters that behave differently depending on who uses that (wire adapter vs user invoking the function directly), we have added a simple mechanism to support such feature via `adapter` property member expression on the callable. This opens the door to transition existing adapters to the new form. E.g.: APEX adapters are all callable objects.
 
-Additionally, those callable objects can implement a forking logic based on the type of argument, if there is a desire to avoid the `adapter` property member expression.
+Additionally, those callable objects can implement a forking logic based on the type of argument, if there is a desire to avoid the `adapter` property member expression. E.g.:
+
+```js
+export function invokeApex(...args) {
+    if (new.target) {
+        // invocation via new, return a WireAdapter instance
+        const [ dataCallback ] = args;
+        // ...
+    } else {
+        // standard function call, return a Promise of some Apex controller result
+        const [ apexControllerParams ] = args;
+        // ...
+    }
+}
+```
 
 # How we teach this
 
--- a/text/0000-wire-reform.md
+++ b/text/0000-wire-reform.md
@@ -8,7 +8,7 @@ This RFC describes the way to decouple the wire service from LWC entirely, and i
 
 # Motivations
 
-There is a dual-dependency between `@lwc/engine` and `@lwc/wire-service`, even though neither of those two packages are importing each other, it is the responsibility of the adapter author to connect them via `registerWireService(register)` where `registerWireService()` is provided by `@lwc/wire-service` and `register()` is provided by `@lwc/engine`. This is, by itself, complex and confusing. Additionally, there is another `register()` method from `@lwc/wire-service` that is used by authors to link their wire adapters with their adapter ID (identity). This process also posses a limitation, and unnecessary dependency, making adapters to tied to LWC.
+There is a dual-dependency between `@lwc/engine` and `@lwc/wire-service`, even though neither of those two packages are importing each other, it is the responsibility of the adapter author to connect them via `registerWireService(register)` where `registerWireService()` is provided by `@lwc/wire-service` and `register()` is provided by `@lwc/engine`. This is, by itself, complex and confusing. Additionally, there is another `register()` method from `@lwc/wire-service` that is used by authors to link their wire adapters with their adapter ID (identity). This process also poses a limitation, and unnecessary dependency, making adapters to tied to LWC.
 
 Additionally, there are various situations where the wired field or method is not behaving correctly, the most notable example is when a configuration value that uses a member expression might not trigger the update on the config. (e.g. `wire(foo, { x: '$foo.bar' }) data`, if `foo.bar` changes, config is not updated). This is because the wire decorator is not relying on the reactivity system used by LWC, and instead it relies on getter and setters that are slower, intrusive, complex and do not cover the whole spectrum of mutations that could occur on a component.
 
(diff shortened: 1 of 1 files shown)


text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+        return (
+            <div>{this.state.valueProducedByMyWireAdapter}</div>
+        );
+    }
+}
+```
+
+# Adoption strategy
+
+Since there is a need to support callable adapters that behave differently depending on who uses that (wire adapter vs user invoking the function directly), we have added a simple mechanism to support such feature via `adapter` property member expression on the callable. This opens the door to transition existing adapters to the new form. E.g.: APEX adapters are all callable objects.
```
pmdartus: If I understand correctly, the Apex adapter implementation would look like:

```ts
export function invokeApex(...args) {
  if (new.target) {
    // invocation via new, return a WireAdapter instance
    const [dataCallback, contextCallback] = args;
  } else {
     // standard function call, return a Promise of some Apex controller result
    const [apexControllerParams] = args;
  }
}
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+* `@wire` decorator will create an instance of the provided `WireAdapter` during the component initialization routine by providing the data callback, and the context callback as arguments. 
+* `@wire` decorator will invoke the adapter's `connect` and `disconnect` based on the internal hooks used per component.
+* `@wire` decorator will create a mutation tracking phase to track any access executed during the computation of the config before calling the `update` routine to be able to detect mutations on those values and issue another update on the adapter instance.
+* `@wire` decorator will implement the logic to provide contextual information when requested by the wire adapter.
+* `@lwc/compiler` will provide a config function per `@wire()` declaration to produce a new config object when invoked with the `component` as the first argument. The `@wire` adapter can rely on that config function to produce a new config object at will.
+* `@lwc/wire-service` becomes lightning platform specific for most part (`register()` method), since anyone can implement the wire adapter protocol.
```
pmdartus: Since this proposal drops the wire adapter registration, is the `@lwc/wire-service` package will only be used for legacy wires. Once all the wire adapters have been updated we should get rid of this package.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Adoption strategy
+
+Since there is a need to support callable adapters that behave differently depending on who uses that (wire adapter vs user invoking the function directly), we have added a simple mechanism to support such feature via `adapter` property member expression on the callable. This opens the door to transition existing adapters to the new form. E.g.: APEX adapters are all callable objects.
+
+Additionally, those callable objects can implement a forking logic based on the type of argument, if there is a desire to avoid the `adapter` property member expression.
+
+# How we teach this
+
+* For adapter consumers, nothing changes.
```
pmdartus: Side note: there may be some changes that we can do to the [@salesforce/wire-service-jest-util](https://github.com/salesforce/wire-service-jest-util) project to make wire adapter mocking more pleasant.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+* For adapter author, the wire protocol is now well defined and does not need registration, which means it is easier to reasoning about compared to the existing mechanism.
+* As for existing adapters based on `@lwc/wire-service`, they can remain the same until after they get refactored and simplified when possible.
+
+# Unresolved questions
+
+* In the current implementation, a wired field is `writable`, which means the component author can alter the value of the field at will. What should we do? a) throw on setter, b) do nothing on setter, c) preserve the current semantics. This is a breaking change if we do a) or b), while the current behavior is weird.
```
pmdartus: While the semantic can be improved with a read-only field, I don't see any major issue with the current behavior.
At worst developers can override a property that is set by the `@wire` decorator, which is not harmful IMO.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+* How context providers can be provisioned? In theory, a context provider is bound to a particular framework/system, while the context consumer is abstracted out in the wire adapter, and specific implementations per framework can provide the piping into the wire adapter protocol via the second argument in the constructor and the `adapter.context(uid, value)` mechanism. Is this sufficient?
```
pmdartus: I am not sure to understand, what do you mean here?

text/0000-wire-reform.md
caridy: exactly! but in the interim, you can just continue doing what you're doing, by calling register on the wire-service package, and we can figure a way to do it for now for legacy adapters.

text/0000-wire-reform.md
caridy: YES! I believe that package will become just a referral implementation, an optional base implementation, and a bunch of types.

text/0000-wire-reform.md
caridy: let's chat about this today.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+- RFC PR: TBD
+- Lightning Web Component Issue: (leave this empty)
+
+# Summary
+
+This RFC describes the way to decouple the wire service from LWC entirely, and implement reactive tracking for wired configuration and wired methods.
+
+# Motivations
+
+There is a dual-dependency between `@lwc/engine` and `wire-service`, even though neither of those two packages are importing each other, it is the responsibility of the adapter author to connect them via `registerWireService(register)` where `registerWireService()` is provided by `wire-service` and `register()` is provided by `@lwc/engine`. This is, on itself, complex and confusing. Additionally, there is another `register()` method from `wire-service` that is used by authors to link their wire adapters with their adapter ID (identity). This process also posses a limitation, and unnecessary dependency, making adapters to tied to LWC.
```
ravijayaramappa:
```suggestion
There is a dual-dependency between `@lwc/engine` and `@lwc/wire-service`, even though neither of those two packages are importing each other, it is the responsibility of the adapter author to connect them via `registerWireService(register)` where `registerWireService()` is provided by `@lwc/wire-service` and `register()` is provided by `@lwc/engine`. This is, by itself, complex and confusing. Additionally, there is another `register()` method from `@lwc/wire-service` that is used by authors to link their wire adapters with their adapter ID (identity). This process also posses a limitation, and unnecessary dependency, making adapters to tied to LWC.
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+
+Additionally, there are various situations where the wired field or method is not behaving correctly, the most notable example is when a configuration value that uses a member expression might not trigger the update on the config. (e.g. `wire(foo, { x: '$foo.bar' }) data`, if `foo.bar` changes, config is not updated). This is because the wire decorator is not relying on the reactivity system used by LWC, and instead it relies on getter and setters that are slower, intrusive, complex and do not cover the whole spectrum of mutations that could occur on a component.
+
+Finally, keeping the wire service tied to LWC means that when needed, wire adapters will not be very useful beyond LWC, when in reality their are just not tied to the component system.
```
ravijayaramappa:
```suggestion
Finally, keeping the wire service tied to LWC means that when needed, wire adapters will not be very useful beyond LWC, when in reality they are not tied to the component system.
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Goals
+
+The primary goal of this RFC is to decouple the wire service from the LWC and from the LWC Wire Decorator implementation.
```
ravijayaramappa:
```suggestion
The primary goal of this RFC is to decouple the wire service from LWC and the LWC Wire Decorator implementation.
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+
+As a secondary goal is to embrace reactivity for the configuration payload and the wired method in LWC.
```
ravijayaramappa:
```suggestion
As a secondary goal, to embrace reactivity for the configuration payload and the wired method in LWC.
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+# Proposal
+
+This reform is focused on the refactor of the wire decorator code, and the wire-service code. As part of the separation process, there are certain responsibilities that must be well defined:
+
+## Responsibilities of the wire decorator
+
+* To install a prototype descriptor to handle the wired field value from the `vm` of the component.
+* To create an instance of an adapter and link it to the host.
+* To signal to the adapter instance when the component is connected or disconnected.
+* To signal to the adapter instance when the config have changed by providing the new config object.
```
ravijayaramappa:
```suggestion
* To signal to the adapter instance when the config has changed by providing the new config object.
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+* To extract the config value from the host object by relying on the compiler's wire metadata.
```
ravijayaramappa: host object or host element?

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+* To signal to the adapter instance when a context is available by providing the new context value.
```
ravijayaramappa: How is this point different from responsibility #3?

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+* `@wire` decorator will create a mutation tracking phase to track any access executed during the computation of the config before calling the `update` routine to be able to detect mutations on those values and issue another update on the adapter instance.
+* `@wire` decorator will implement the logic to provide contextual information when requested by the wire adapter.
+* `@lwc/compiler` will provide a config function per `@wire()` declaration to produce a new config object when invoked with the `component` as the first argument. The `@wire` adapter can rely on that config function to produce a new config object at will.
+* `@lwc/wire-service` becomes lightning platform specific for most part (`register()` method), since anyone can implement the wire adapter protocol.
+
+### Wire Adapter Protocol
+
+The formalization of the wire adapter protocol is important because that enables the interoperability aspect of this feature. The adapter's code should not be aware for the component system, or even the application framework, it only cares about very specific hints to produce a stream of data. The following describes the proposed protocol:
```
ravijayaramappa:
```suggestion
The formalization of the wire adapter protocol is important because that enables the interoperability aspect of this feature. The adapter's code should not be aware of the component system, or even the application framework. It only cares about very specific hints to produce a stream of data. The following describes the proposed protocol:
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+# Adoption strategy
+
+Since there is a need to support callable adapters that behave differently depending on who uses that (wire adapter vs user invoking the function directly), we have added a simple mechanism to support such feature via `adapter` property member expression on the callable. This opens the door to transition existing adapters to the new form. E.g.: APEX adapters are all callable objects.
+
+Additionally, those callable objects can implement a forking logic based on the type of argument, if there is a desire to avoid the `adapter` property member expression.
+
+# How we teach this
+
+* For adapter consumers, nothing changes.
+* For adapter author, the wire protocol is now well defined and does not need registration, which means it is easier to reasoning about compared to the existing mechanism.
```
ravijayaramappa:
```suggestion
* For adapter author, the wire protocol is now well defined and does not need registration, which means it is easier to reason about compared to the existing mechanism.
```

text/0000-wire-reform.md
```diff
@@ -0,0 +1,146 @@
...
+```ts
+interface WireAdapter {
+    update(config: Record<string, any>);
+    context(uid: string, value: any);
+    connect();
+    disconnect();
+}
+
+interface WireAdapterConstructor {
+    new (callback: EmitDataCallback, contextualizer?: RequestContextCallback): WireAdapter;
```
pmdartus: In order to accept more parameters in the future, I would suggest passing an object instead of 2 callbacks.

```ts
interface WireAdapterConstructorConfig {
    callback: EmitDataCallback, 
    contextualizer?: RequestContextCallback
}

interface WireAdapterConstructor {
    new (config: WireAdapterConstructorConfig): WireAdapter;
}
```

text/0000-wire-reform.md
caridy: well, the wire decorator is LWC specific... in which case this is the component reference. I will update the sentence to reflect that. I will also add the LWC wire decorator to be explicit about what decorator are we talking about.

text/0000-wire-reform.md
caridy: 3 is about connect and disconnect, this is about contextual configuration.

text/0000-wire-reform.md
caridy: I will take care of that.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,197 @@
...
+type DataCallback = (value: any) => void;
+type ConfigValue = Record<string, any>;
+type ContextValue = Record<string, any>;
+type WireAdapterSchemaValue = 'optional' | 'required';
+```
+
+_Notes:_
+
+* Not all environments will support or need context (e.g.: preloading LDS data), but does supporting it can rely on the static field called `contextSchema` to provide the context value when available.
+* Some environments might choose to implement validation rules for `configSchema` and `contextSchema` to guarantee compliance.
```
pmdartus: What is the advantage `configSchema` and `contextSchema` vs doing the validation by hand in the `update` callback?
It feels that we are adding higher-level abstractions to low-level APIs.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,244 @@
...
+}
+type DataCallback = (value: any) => void;
+type ConfigValue = Record<string, any>;
+type ContextValue = Record<string, any>;
+type WireAdapterSchemaValue = 'optional' | 'required';
+```
+
+_Notes:_
+
+* Not all environments will support or need context (e.g.: preloading LDS data), but does supporting it can rely on the static field called `contextSchema` to provide the context value when available.
```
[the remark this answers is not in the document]
caridy: trying to say that the runtime is responsible for providing or not context support, some will, some will not. Does that support context can do so by inspecting the wire adapter's context schema on the individual basis.

text/0000-wire-reform.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Semantic changes for `@wire` decorator IDL
+
+As today, there are few restrictions and ambiguities with the IDL for the config object in `@wire` decorator declarations. This sections will describe the new semantics, which are almost identical for most cases:
+
+* `$token` can only appear as the value of a top level member property, e.g.: `@wire(foo, { x: '$prop1' })` will continue to be valid, while `@wire(foo, { x: { y: '$prop1' } })` will throw a compiler error, while today it doesn't but the value is never transformed, and remains as a string value. This is a non-backwards compatible change, but we believe this is a very low risk for something that was not working as expected.
+* there will be no identity for inline JSON objects when assigned to a property in the config object, e.g.: `@wire(foo, { x: { y: 1 } })` where the value of `x` will be computed every time, instead of cached per instance or per class as today.
+* there will be identity preserved when assigning a reference values in the config object, e.g.: `@wire(foo, { x: someValue })` where the value of `x` will be a reference to `someValue` during the class declaration.
```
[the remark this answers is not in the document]
caridy: yes, someValue retains identity, and x points to it.

additionally:
a reference values => reference values

text/0000-wire-reform.md
```diff
@@ -0,0 +1,244 @@
...
+* every time that `adapter.update()` is invoked, a new config object will be provided as a first argument, no identity is preserved in this case.
```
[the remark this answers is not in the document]
caridy: yes, all your assumptions are correct. maybe we should clarify that... with your example or something.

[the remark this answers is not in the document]
caridy: additionally, `first !== second`

text/0000-wire-reform.md
```diff
@@ -0,0 +1,244 @@
...
+
+* As today, the descriptor installed on the prototype of the component by the `@wire` decorator was identical to `track` when decorating a field. This means that the component author could change the value of the field, and such change will be tracked. This is no longer the case, and even though the author can still change the value, it will not be reactive, causing no side effects on the UI of the element. This is a non-backwards compatible change, but we believe this is a very low risk for something that was not working as expected.
+
+### Wire Adapter Protocol
+
+The formalization of the wire adapter protocol is important because that enables the interoperability aspect of this feature. The adapter's code should not be aware of the component system, or even the application framework. It only cares about very specific hints to produce a stream of data. The following describes the proposed protocol:
+
+```ts
+interface WireAdapter {
+    update(config: ConfigValue, context?: ContextValue);
```
mburr-salesforce: Do you have an example of the difference between `config` and `context`? I'm trying to understand what would motivate an adapter to request/accept information via one vs the other.

mburr-salesforce: The current implementation requires wire adapters to emit `null` when presented with incomplete `config`s. Does this new protocol guarantee that `update` is only ever invoked with fully-resolved `config`s (which implies that an incomplete `config` is an error on the component author's part)?

text/0000-wire-reform.md
```diff
@@ -0,0 +1,244 @@
...
+    connect();
+    disconnect();
+}
+interface WireAdapterConstructor {
+    new (callback: DataCallback): WireAdapter;
+    configSchema?: Record<string, WireAdapterSchemaValue>;
+    contextSchema?: Record<string, WireAdapterSchemaValue>;
+}
+type DataCallback = (value: any) => void;
```
mburr-salesforce: Is `value` here the same `{ data: ..., error: ... }` construct we have today?

caridy: The protocol is not enforcing such structure. We have debated that in other forums multiple times... we ended up letting individual adapters to implement their own protocol when it comes to data and error. We do offer that as a guideline... but there are cases in which that makes no sense... (e.g.: navigation context). This method in this example is precisely to highlight that, you can make it `{ data: value }` it you wish.

text/0000-wire-reform.md
caridy: This is a very good question. We have a more flexible API now. The implementation in `@lwc/wire-service` abstraction does take care of the `null` value before dispatching the config event IIRC (I will double check), but the wire protocol itself doesn't care about that anymore, that's for the adapters to decide when to go and process the new config or the new context.

caridy: another good question, I myself struggled with this, I wanted to make them one thing, but @diervo and @jbenallen convinced me that these should remain separate, one of them optional, mostly because of the types... and validations that we can do in the future, and how easy it is for the adapter to understand the channel in which those two set of data are getting collected.

A good example is probably an adapter that load records, where the record indicator is contextual (the tab that you are on) and the number of elements is configurable.

kevinv11n: I believe the `null` handling was internal to LDS. I don't see such handling in https://github.com/salesforce/lwc/blob/master/packages/@lwc/wire-service/src/index.ts#L113-L117

mburr-salesforce: IIRC it might have been a side effect of the fact that the wire adapters were hooked up before the $ references could be expanded?

caridy: thanks @kevinv11n, that's what I thought! we are good!

caridy:
thanks everyone for the review, we are merging this today and start focusing on the implementation on PR https://github.com/salesforce/lwc/pull/1459/

text/0000-wire-reform.md
```diff
@@ -0,0 +1,247 @@
...
+* Some environments might choose to implement validation rules for `configSchema` and `contextSchema` to guarantee compliance.
+* we favored the `DataCallback` over a promised based on `update()` calls because the callback can be invoked sync and async, but most important, because update might never be called by the environment.
+
+### Semantic changes for `@wire` decorator IDL
+
+There exist a few restrictions and ambiguities with the IDL for the config object in `@wire` decorator declarations. This section will describe the changed semantics. Most use cases of `@wire` are unaffected. 
+
+* `$token` can only appear as the value of a top level member property, e.g.: `@wire(foo, { x: '$prop1' })` will continue to be valid, while `@wire(foo, { x: { y: '$prop1' } })` will throw a compiler error, while today it doesn't but the value is never transformed, and remains as a string value. This is a non-backwards compatible change, but we believe this is a very low risk for something that was not working as expected.
```
priandsf: @caridy I faced the need of having a $token injected into a non top level property. Is there a technical limitation why this can't be done, or is it because of a potential performance concern with object comparison?
On the same subject, I also faced the need of having the $token being replaced as part of a string, like in:
```
    @track offset = 0
    @track limit = 5
    @wire(useFetch, {
        url: '/users?offset=$offset&limit=$limit',
    ...
```
Or, even better, I would love some of these parameters to be functions like the peudo code bellow:
```
    @track offset = 0
    @track limit = 5
    @wire(useFetch, {
        url: url('/users', {offset: '$offset', limit: '$limit'} ),
    ...
```
It probably means that the whole json object should be recreated on any change, and compared to the previous instance. As there are generally few properties, this should not be a performance problem.

text/0000-wire-reform.md
priandsf: @caridy @diervo  It looks like the RFC does not match the implementation:
update(config: Record<string, any>) 
Thus, would be great to update the RFC as it somehow serves as the documentation.

PR #15, opened 2019-10-04
Initial inner html binding RFC

priandsf:
RFC for inner HTML binding

### DIFF
--- a/text/0000-innerhtml-binding.md
+++ b/text/0000-innerhtml-binding.md
@@ -1,0 +1,108 @@
+- Start Date: 2019-10-03
+- RFC PR: (leave this empty)
+- Lightning Web Component Issue: (leave this empty)
+
+# Summary
+
+Applications dealing with rich content, like commerce apps, need to render 
+this content as raw HTML, typically coming from a database or a CMS. Today, 
+the data binding capability always escapes the HTML for security reasons,
+which makes the rich content data not rendered as expected.
+
+# Basic example
+
+The following template render some HTML content as innerHTML  
+
+```
+function createMarkup() { return {__html: 'First &middot; Second'}; };
+
+<template>
+    <div lwc:innerHTML={createMarkup()}/>
+</template>
+```
+
+# Motivation
+
+It is common in customer facing apps to render rich content coming from a
+CMS or equivalent.  
+
+As of today, and according to the LWC documentation, this capability is 
+supported by acting directly with the DOM within `renderedCallback()`.
+The templating mechanism does not provide any binding capability that renders
+some content as raw HTNL.  
+
+This breaks with SSR, as `renderedCallback()` and DOM manipulation won't be
+allowed when a component is rendered on the server. Moreover, even if it was
+possible, it would make the client side hydration more complex as the content
+is not generated as part of the VDOM.  
+
+Having to act on the DOM is also not user friendly compared to a binding
+facility.  
+
+
+# Detailed design
+
+It seems that there is a consensus among the most popular libraries to expose the
+innerHTML property of an element, although the exposed name for this property varies:
+
+React: [https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html]().  
+Angular: [https://angular.io/guide/template-syntax#property-binding-or-interpolation]().  
+VueJS: [https://vuejs.org/v2/guide/syntax.html#Raw-HTML]().  
+
+The Angular choice of `innerHTML` feels the most natural, while the React `dangerouslyInnerHTML`
+has probably be named this way to not conflict with any potential component property.
+To avoid any name collision, LWC can prefix it with the technical "lwc:" prefix.
+
+React makes it even more secure by forcing the HTML to be added to a temporary 
+object with a `__html` property. This avoids undesired assignment from a method
+that does not sanitize the content:
+
+```
+<template>
+    <div lwc:innerHTML={getUserName()}/>
+</template>
+```
+
+It is up to the component developer to make sure that the injected HTML is properly
+sanitized ([https://en.wikipedia.org/wiki/HTML_sanitization]()). Several client side
+libraries, easily consumable by an LWC application, are available to do the job.
+
+The compiler generates an error if an element has both an `innerHTML` property and
+some content defined, like bellow:  
+
+```
+<template>
+    <div lwc:innerHTML={myHTML()}>
+      Adding content when innerMTML is defined leads to a compilation error
+    </div>
+</template>
+```
+
+
+# Drawbacks
+
+- It might open security breaches if not used appropriately. The proposed API forces the developer to know what he/she is doing and avoids mistaken assignment.  
+- Locker might prevent this assignment. To be verified.
+
+
+# Alternatives
+
+Different syntaxes for the `innerHTML` attribute are possible, but they finally lead to
+the same results. It is then more a matter of preference and consistency.  
+
+Another solution would define a new binding syntax, like for example `${{myhtml()}}` or
+`${htnl:myhtml()}}`. But such a syntax could be used everywhere a binding is possible,
+including attribute values. This could lead to undefined behaviors why not providing any
+value. The proposed syntax makes sure that the `innerHTML` binding always applies to the
+content of an element.
+
+
+# Adoption strategy
+
+Because the attribute uses the reserved `lwc:` prefix, there is not backward compatibility
+issue, and there is no potential name conflicts.
+
+
+# How we teach this
+
+inner HTML or raw HTML binding are the industry adopted terms for such a feature.


text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,108 @@
...
+facility.  
+
+
+# Detailed design
+
+It seems that there is a consensus among the most popular libraries to expose the
+innerHTML property of an element, although the exposed name for this property varies:
+
+React: [https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html]().  
+Angular: [https://angular.io/guide/template-syntax#property-binding-or-interpolation]().
```
pmdartus:
```suggestion
Angular: [[innerHTML]](https://angular.io/guide/template-syntax#property-binding-vs-interpolation).  
```

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+# Detailed design
+
+It seems that there is a consensus among the most popular libraries to expose the
+innerHTML property of an element, although the exposed name for this property varies:
+
+React: [https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html]().  
+Angular: [https://angular.io/guide/template-syntax#property-binding-or-interpolation]().  
+VueJS: [https://vuejs.org/v2/guide/syntax.html#Raw-HTML]().
```
pmdartus:
```suggestion
VueJS: [v-html attribute](https://vuejs.org/v2/guide/syntax.html#Raw-HTML).  
```

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,108 @@
...
+React: [https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html]().  
+Angular: [https://angular.io/guide/template-syntax#property-binding-or-interpolation]().  
+VueJS: [https://vuejs.org/v2/guide/syntax.html#Raw-HTML]().  
+
+The Angular choice of `innerHTML` feels the most natural, while the React `dangerouslyInnerHTML`
+has probably be named this way to not conflict with any potential component property.
+To avoid any name collision, LWC can prefix it with the technical "lwc:" prefix.
+
+React makes it even more secure by forcing the HTML to be added to a temporary 
+object with a `__html` property. This avoids undesired assignment from a method
```
pmdartus: I am not convinced about the benefit of adding this `__html` property on an object. The `lwc` prefix is fairly uncommon in the template, adding this extra layer of "security" is feels unnecessary.

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Basic example
+
+The following template render some HTML content as innerHTML  
+
+```
+function createMarkup() { return {__html: 'First &middot; Second'}; };
+
+<template>
+    <div lwc:innerHTML={createMarkup()}/>
```
pmdartus: HTML attribute names are case-insensitive and the parser will not be able to differentiate between `lwc:innerHTML` and `lwc:innerhtml`. I would rather the following directive names:
* `lwc:inner-html`
* `lwc:html`
* `lwc:raw`

seckardt: I would vote for the simple and meaningful `lwc:html` in that case. Also this syntax leaves us the opportunity to bolt on security/sanitization afterwards, similar to for instance Angular's DomSanitizer approach.

pmdartus: I would like to keep LWC lightweight and not have to bring DOMPurify into LWC. I would rather delegate the sanitization to Locker if present, otherwise do a plain `Element.innerHTML`. @caridy what do you think?

priandsf: I agree with that

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,108 @@
...
+It seems that there is a consensus among the most popular libraries to expose the
+innerHTML property of an element, although the exposed name for this property varies:
+
+React: [dangerouslySetInnerHTML attribute](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml).  
+Angular: [[innerHTML]](https://angular.io/guide/template-syntax#property-binding-vs-interpolation).  
+VueJS: [v-html attribute](https://vuejs.org/v2/guide/syntax.html#Raw-HTML).  
+
+The Angular choice of `innerHTML` feels the most natural, while the React `dangerouslyInnerHTML`
+has probably be named this way to not conflict with any potential component property.
+To avoid any name collision, LWC can prefix it with the technical "lwc:" prefix.
```
caridy: we call this prefix a directive.

caridy:
General direction seems fine to me, but many of the semantics are TBD for me, e.g.:

* can you have custom elements inside the inserted html?
* can this be used in a slot element?
* can this be used as light dom of a custom element?
* what is the process of updating the value?
* how can locker and others hook into the sanitization process? it mentioned that lwc is not responsible for what is inserted there, but until now... a HTML only component was always safe, this is not longer the case. How to mitigate this?

Other notes:

* directives (as attributes) are all lowercased, - separated, so `lwc:innerHTML` is not valid.
* ssr as a secondary motivation is fine, but the primary motivation should be very clear, it is about inserting html content produced by some other layer.
* how to make this work with synthetic shadow? how to add keys on the spot for any new inserted node so they can be associated to the respective shadow root? cc @ravijayaramappa 
* have you looked at trusted types proposal from google? it seems that the innerHTML injection is something covered by that proposal, maybe that's the way to introduce the hooking mechanism for such assignment where only trusted html content can be inserted into the dom via this directive.

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Motivation
+
+It is common in customer facing apps to render rich content coming from a
+CMS or equivalent.  
+
+As of today, and according to the LWC documentation, this capability is 
+supported by acting directly with the DOM within `renderedCallback()`.
+The templating mechanism does not provide any binding capability that renders
+some content as raw HTNL.
```
tedconn:
```suggestion
some content as raw HTML. 
```

caridy:
@priandsf @seckardt did you get a chance to look at my general notes on this RFC?

priandsf:
@caridy Thank you for your feedback. I made some changes to the RFC to match your comments. Please feel free to edit it as well.

text/0000-innerhtml-binding.md
```diff
@@ -78,17 +85,41 @@ some content defined, like bellow:
...
+it updates the DOM element content by assigning it to the element.innerHTML property.
+
 
 ## Drawbacks
 
-- It might open security breaches if not used appropriately. The proposed API forces the developer to know what he/she is doing and avoids mistaken assignment.  
-- Locker might prevent this assignment. To be verified.
+  - It might open security breaches if not used appropriately. The proposed API forces 
+  the developer to know what he/she is doing and avoids mistaken assignment.  
+  - Locker might prevent this assignment. To be verified.
```
tedconn: I would add this to an Open Questions section.

text/0000-innerhtml-binding.md
```diff
@@ -78,17 +85,41 @@ some content defined, like bellow:
...
 
 ## Drawbacks
 
-- It might open security breaches if not used appropriately. The proposed API forces the developer to know what he/she is doing and avoids mistaken assignment.  
-- Locker might prevent this assignment. To be verified.
+  - It might open security breaches if not used appropriately. The proposed API forces 
+  the developer to know what he/she is doing and avoids mistaken assignment.  
+  - Locker might prevent this assignment. To be verified.  
+
+Note: the security issue is *not* different from setting the HTML with a DOM operation
```
caridy: That's correct, but the difference is that the user is setting it vs the framework is setting it. Locker only controls the user code, framework on the other hand is system mode, and not controlled by locker.

This touches on another RFC for LWC that is pending to be written, which is set to provide hooks that libraries like locker can take control over to sanitize injections/sinks. That's TBD.

text/0000-innerhtml-binding.md
```diff
@@ -78,17 +85,41 @@ some content defined, like bellow:
...
-    <div lwc:innerHTML={myHTML()}>
+    <div lwc:inner-html={myHTML()}>
       Adding content when innerMTML is defined leads to a compilation error
     </div>
 </template>
 ```
 
+### Validity
+
+The `lwc:inner-html` attribute can be set on any tag in an LWC template with the
```
caridy: bikeshedding on the name of the attribute, I think something like: `lwc:dangerous-inner-html` is a better name, to follow suit with our react friends :)

priandsf: That was actually my initial proposal but I got some pushback on such a long name. I'm fine either way, feel free to modify the RFC.

text/0000-innerhtml-binding.md
priandsf: Ok we need more info then. Is there a way to detect that locker is available and then call its sanitization methods? Sounds like hard-coding it but might be a workaround before a more generic solution.

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,145 @@
...
+</template>
+```
+
+The directive is named `inner-html` as it mimics what the DOM `innerHTML`
+property does. As such, it can be used on any tag where the `innerHTML`
+property is available. The HTML fragment inserted is not processed by the LWC
+compiler and thus LWC components won't be created unless they are exposed as
+custom elements to the browser.
+
+## Motivation
```
tedconn: This is a must-have for any personalization work @showerbee

priandsf:
@caridy @pmdartus What is the status on this? It is labelled as "need consensus". Should I setup a quick meeting so we move forward?

caridy:
We can have a quick meeting yes! I will do another pass next week.

caridy:
## Notes from the meeting:

### General discussion

* the directive is a little confusing (plus it comes with extra restrictions by the parser on where can be used.
* a new syntax for interpolation seems better, but it will require learning a new thing, syntax is always tricky to get right.
* @JanMiksovsky suggested a new component (e.g.: `<lightning-dangerous-html value={someHTML}>`) might be a far better option, and easy to implement as well.

### There were few questions that should be answered by this RFC:

1. What is the exact use case? CKEditor-like producing HTML content that should be shown to the user? Do you need to style that content? Do you need to access (query) those elements? It seems that the answer to this question is that you just want to show it to the user. In which case the component approach is more suitable because you can show the content in the component's shadow dom in a controlled manner.
2. Who is responsible for the sanitization? should the consumer sanitize (maybe via a platform library) and pass the sanitized value down? or should the string be sanitized always by the final tier?
3. What is the best approach? new directive? new interpolation syntax? or a new component?

Once we answer these 3 questions, we can move on @priandsf.

### Some other feedback:

I had a chance to get @diervo up to speed on this discussion, as well as some follow up discussion with @JanMiksovsky and @pmdartus, and I believe we all agree on the fact that a component seems like a safe bet. We are still not clear on the implementation details, but we can figure that out easily.

@diervo thinks (and I agree), that a new constructor from LWC on top of `LightningElement` is good enough, it is going to be probably another optional package, but its job is to provide the foundational parts, and the branding logic so we can apply the logic for server and client there based on the branding of the component instance. This way, anyone can extend that class and call the method that updates the content of its shadow, and that will also work on the server side OOTB.

blittle:
I'm curious if we want to consider other wider community implications of this feature? Right now we are mainly (appropriately) focusing on the use cases in core, commerce cloud, and other SF projects. If we want to foster wider community adoption of LWC, I think we should also consider community and wider web use cases. In general, I'm personally a fan of keeping as much functionality as possible outside the core framework, but this particular feature is something that all major front-end frameworks provide. To me, it seems like a table stake.

With that in mind and the three questions above:

1. I've worked on multiple (non-sf) projects that needed to style dynamic cms html content. Sure, maybe the immediate SF use cases don't need to style or access the inner elements, but I doubt that is the same for the wider web community. 
2. I've also worked on custom business driven sanitization (again, non-sf), where we white-listed only specific elements and attributes. If the framework itself does do sanitization, I would hope that 1) it can be overridden by custom sanitization and 2) there won't be overhead in the framework for those that roll their own.

priandsf:
I concur with @blittle.

My comments:
- Attribute syntax.
Thinking about it, I believe the Frameworks used that because it helps the VDOM comparison with the actual DOM. If an element has this attribute, then you don't compare the body but re-inject the content if the attribute value changed.
- Binding syntax.
I'm fine with that as well, which might be consistent with the localization effort and a built-in function. This is what lit-html is doing.
- Specific component.
Well, we are already doing that, there is no need from Framework support. SSR can be handled with a hook where we provide a specific renderer for this component.
Now, injecting the content in the Shadow DOM won't always work as we sometimes need to style this content (actually Commerce does that). Using the synthetic shadow in not a long term option, as it forces us to stick with that forever, for the whole application.

I also agree that the sanitization depends on the use cases, there is no one size fit all. So it is better to leave it to the developer, even though we can put guards to make sure that whatever the developer does is not a mistake.

tedconn:
@priandsf it seems like there is a general preference to provide a component that does this instead. Are you ok if we mark this as rejected? I can look at implementing a `lightning-html` component for global use to fill this gap for on-platform use cases.

priandsf:
@tedconn A dedicated component works even though a native feature makes LWC on part with the other frameworks. We implemented such a component in the past. Typically, the DOM can be manipulated during in renderCallback(), with a cache to avoid injecting the markup systematically. Thus, there is no change needed to the core LWC code .
You also need a solution for SSR, as there is no DOM available on the server. The SSR engine should have an exception behavior for that lightning-html component, or any other that eventually needs HTML content to be rendered.
That being said, as long as it efficiently renders HTML and it is SSR enabled, I'm fine with a component.

pmdartus:
@tedconn The LWC team will be working on this proposal next release. Let's keep this open until then, so we can get back to it.

tedconn:
> @tedconn The LWC team will be working on this proposal next release. Let's keep this open until then, so we can get back to it.

@pmdartus

seckardt:
@pmdartus @tedconn As this RFC was one of the SSR precursors, can you briefly explain how LWC intends to deal with newer developments such as [declarative shadow DOM](https://web.dev/declarative-shadow-dom/) in general?

pmdartus:
> > @tedconn The LWC team will be working on this proposal next release. Let's keep this open until then, so we can get back to it.
>
> @pmdartus 

This is unfortunately not happening this release based on the planning either.

> @pmdartus @tedconn As this RFC was one of the SSR precursors, can you briefly explain how LWC intends to deal with newer developments such as declarative shadow DOM in general?

I don't think this proposal will do anything special regarding declarative shadow DOM. That being said, its something worth calling out in the proposal.

pmdartus:
I took some time to update the proposal today. You can find a tentative implementation of this proposal at https://github.com/salesforce/lwc/pull/2256

priandsf:
@pmdartus Great work!
One of the risk is with sanitizeHtmlContent not being set. Shouldn't we force it to be explicitly defined? I mean if I don't want sanitization, then I can assign an empty function to sanitizeHtmlContent. But this is a conscious decision.
lwc:dom="manual" is still useful for us as we are using it to instantiate non LWC web components within a template, so it still has applicability. Or to add a close shadow dom to an element to secure the payment component.

text/0000-innerhtml-binding.md
```diff
@@ -0,0 +1,148 @@
...
+
+The reliance on the `renderedCallback()` life-cycle hook to render raw HTML content is a real issue for server-side rendering. In SSR, the `renderedCallback()` is never invoked. There is currently no straightforward way to inject raw HTML content for server-side rendered components.
+
+This proposal introduces a new `lwc:inner-html` directive to inject raw HTML content that would work both in the browser and on the server.
+
+## Detailed design
+
+### Runtime behavior
+
+At runtime the `lwc:inner-html` directive binds the directive value to the `Element.innerHTML` property. During each rendering cycle, the LWC engine diff the current value with its previous value. If the value changes between two rendering cycles, the LWC engine sets the element `innerHTML` property to the new value.
```
jodarove, line 59:
```suggestion
At runtime the `lwc:inner-html` directive binds the directive value to the `Element.innerHTML` property. During each rendering cycle, the LWC engine diff the current string value with its previous value. If the value changes between two rendering cycles, the LWC engine sets the element `innerHTML` property to the new value.

```

This might be confusing as you can diff the generated DOM tree.

caridy:
The more I think about this... the more I believe this is not going to happen. The idea that LWC templates are immune to XSS as it is today is pretty important, and adding this basically throw that into the trash can. In practical terms, how will that work with Locker xNext?

pmdartus:
Originally port at https://github.com/salesforce/lwc/pull/2256#issuecomment-900097551

----

> The idea that LWC templates are immune to XSS as it is today is pretty important, and adding this new mechanism basically throw that into the trash can.

I don't think it is correct to assume that LWC templates are XSS immune. For example:

```html
<template>
	<a href="javascript:alert(1)">Click</a>
</template>
```

One of the main drivers for this proposal is to allow server-side rendering of raw HTML content; where direct DOM manipulation is impossible. In practical terms, we are using a similar approach than we use to sanitize the `xlink:href` attribute on the `use` tag via the `sanitizeAttribute` hook. For the `lwc:inner-html` we would expose a `sanitizeHtmlContent` hook.

caridy:
@pmdartus `sanitizeAttribute` was just a hack for current locker to work with LWC, and I keep pushing to remove such bizarre behavior now that we have locker vnext. The problem really becomes that this injection mechanisms of hooks are difficult to control.

As for `<a href="javascript:alert(1)">Click</a>`, the `sanitizeAttribute` is not really worry about that case because CSP rules can prevent that from working, it is really about other kind of filtering.

pmdartus:
> I keep pushing to remove such bizarre behavior now that we have locker vnext.

I think this proposal should align with what will replace `sanitizeAttribute`. @caridy Without speaking of `sanitizedHtmlContent` how are you planning to handle `sanitizeAttribute` in Locker VNext? I would be happy to participate in the discussion if needed.

> it is really about other kind of filtering

What kind of filtering are you speaking of? Are you speaking about DOM clobbering?

blittle:
In Experience Cloud, we have a rich text editor component, that customers can drag and drop into any experience. That RTE component takes a string of HTML (sanitized) as a property. If there is no native way to innerHTML that content, then any RTE component on a website will not be SSR'd, which means it won't be indexable by bots, and will probably have a flicker of rendering.

PR #16, opened 2019-10-06
Generate PDF option for LWC

yasarshaikh:
We should give user to option to generate pdf for current LWC implementation (considering its children as well).

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -1,68 +1,52 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
-- RFC PR: (leave this empty)
-- Lightning Web Component Issue: (leave this empty)
+- Start Date: 2019-10-06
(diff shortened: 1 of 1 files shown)


pmdartus:
Hi @yasarshaikh, thank you for raising your concerned regarding, PDF generation and LWC. 

The RFC process is primarily used to discuss changes to the framework along with implementation details. For feature requests, I would recommend creating an issue on the [main](https://github.com/salesforce/lwc) repository first.

I would recommend using [pdfjs](http://raw.githack.com/MrRio/jsPDF/master/docs/module-html.html#~html) if you want to convert HTML content to PDF format.

mpsenn:
Please also upvote this Idea to show support for Lightning component rendering to PDF.
https://success.salesforce.com/ideaView?id=0873A000000CTHKQA4

PR #17, opened 2019-10-29
RFC: `lwc:dom` directive reform

caridy:


text/0000-lwc-dom-directive-retorm.md
```diff
@@ -0,0 +1,69 @@
...
+    <h1>Existing feature (for leaf element only):</h1>
+    <div lwc:dom="manual"></div>
+
+    <h1>Lifting restrictions on existing feature:</h1>
+    <div lwc:dom="manual">
+        <p>can have child elements in template to be combined with manually inserted nodes</p>
+    </div>
+
+    <h1>Introducing new feature (for leaf element only):</h1>
+    <div lwc:dom="synthetic-portal"></div>
```
pmdartus: The proposal doesn't provide details about `syntehtic-portal`? What are the intended use cases, and the expected behviors?

text/0000-lwc-dom-directive-retorm.md
```diff
@@ -0,0 +1,69 @@
...
+</template>
+```
+
+# Motivation
+
+There are 3 main motivations for this reform:
+
+1. The existing restriction that only leaf elements can be marked with `lwc:dom="manual"` is limiting. There are valid use-cases that requires a mix of elements generated from a template plus manually inserted elements, one of them is the use of a graph library that can connect elements (visually) where the elements are controlled via the template declarative mechanism, while the edges are manually generated and inserted by a 3rd party library.
+
+2. Legacy systems, specifically flexipages in Aura applications, are in needs to support rendering legacy DOM structures that are not suitable, or were not designed, to work inside a `ShadowRoot` instance. In such cases, it is imperative that from the perspective of those elements, they should behave as inserted in the document.
```
pmdartus: > it is imperative that from the perspective of those elements, they should behave as inserted in the document.

What are concrete cases where this is needed?

text/0000-lwc-dom-directive-retorm.md
caridy: :) it is WIP, you are not supposed to be reviewing it yet... jajaja... let me finish this today.

sfdciuie:
@caridy Perhaps it's an appropriate time to discuss the ability to insert a component in a different place in the dom - specifically, moving it to be a child up the tree (eg. under document.body for say a global modal, and under a tab for a scoped modal). 

This wouldn't be needed if we had say browser's level `blockingElement` (ie. everything but the element is inert) and top layer, but that is unlikely to come any time soon

If we are ok with this in principle (the component created/rendered someplace else but still fully managed), there are a few ways that this could be achieved, purely for illustration purposes here's an example of one:
```html
<body>
  <lightning-portal lwc:portal="modal"></lightning-portal> <!-- Defines the portal -->
 <x-foo> 
   <y-foo>
       <lighthing-dialog ...>
           <div lwc:dom="portal" lwc:portal-name="modal"> <!-- Insertion/content mode - portal, portal name - "modal" -->
             ...
             <slot name="content"></slot>
             <slot name="footer"></slot> 
             ...
          </div>
```

This should definitely be gated.

jbenallen:
What is the state of this proposal? @caridy

gregwhitworth:
@pmdartus @caridy hey - we have a WI open on this. Are we going to complete this PR? If not can you please close this out.

pmdartus:
Correct if I get the intention of the proposal wrong @caridy.

I don't think we want to pursue this proposal (at least in the current state). One of the major drivers of this proposal is to integrate Aura component in LWC as we are migrating larger chunks of the application. I see in the Light DOM proposal a way to integrate Aura components in LWC without running into shadow DOM related issues.

PR #20, opened 2019-11-09
RFC: Module resolution

diervo:
https://rfcs.lwc.dev/rfcs/lwc/0000-module-resolution

### DIFF
--- a/text/0000-module-resolution.md
+++ b/text/0000-module-resolution.md
@@ -18,7 +18,7 @@ As the LWC ecosystem organically grows both internally at Salesforce and externa
 
 Let's look at a very simple scenario: We have a repository of LWC components we want to open source. We want to use other 3party packages as the foundation for my components, in this case `@ui/components@2.0.0` and `fancy-components@1.0.0`. Its important to note that the `fancy-components` package is using internally `@ui/components@1.0.0`, which is a different version of the package I'm using.
 
-Here is how the structure of the repo looks like  you can take a look at [ a real implementation implementation and test here](https://github.com/salesforce/lwc/pull/1602/files#diff-5cfc908f1b15979cf59306304092ddb5):
+Here is how the structure of the repo looks like  you can take a look at [a real implementation implementation and test here](https://github.com/salesforce/lwc/pull/1602/files#diff-5cfc908f1b15979cf59306304092ddb5):
 
 ```markup
 rootDir
@@ -74,11 +74,11 @@ An important topic to understand first is how module resolution works today in d
 
 JavaScript did not have built-in support for modules, but the community has created impressive work-arounds over the years. The two most important (and unfortunately incompatible) standards are: CommonJS (used fundamentally by Node - with some extra semantics) and AMD (Asynchronous Module Definition), used by RequireJs and a lot of other bundlers.
 
-Note that there are other non-standard formats, loaders and tools like UMD, SystemJS, IIFE, etc, which are interesting for you know and understand but are out of the scope for this proposal.
+Note that there are other non-standard formats, loaders and tools like UMD, SystemJS, IIFE, etc, which are interesting for you to know and understand but are out of the scope for this proposal.
 
 ### Node.js resolution
 
-If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is know nas [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
+If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is known as [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3rd party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
 
 All of this being said after years of very hard work, Node is finnaly addopting ES6 modules (which we will see in the next section) since LTS v12 and they are hoping to [fully transition to it soon](https://twitter.com/MylesBorins/status/1189618753065144322)
 
@@ -262,8 +262,8 @@ Here is the algoritm for module resolution (the instructions are not in a riguro
   
   1. Load the LWC configuration file in the rootDir
     1.1 Search for `lwc.config.js`, if it exist read it.
-    1.2 If lwc.config.js is not found search in package.json
-    (If both are found the configuration file has precedence).
+    1.2 If `lwc.config.js` is not found search in `package.json`
+    (If both are found the `lwc.config.js` file takes precedence).
 
   2. Merge all the ModuleRecords found in the config with 
   the list of initModules. If an alias with the same name 
@@ -287,16 +287,20 @@ Here is the algoritm for module resolution (the instructions are not in a riguro
           - `scope` with the closest configuration path.
           - `entry` with the `path` value.
 
-        3.2.2 If is a `DirModuleRecord` validate the path and find of 
+        3.2.2 Else if is a `DirModuleRecord` validate the path and find of 
         the modules that match the structure:
-        [namespace]/[componentName]/[componentName.{html|css|js|ts}]
+        [namespace]/[componentName]/[componentName.{html|css|js|ts}] and create a 
+        ModuleRecordEntry with:
+          - `specifier` be `[namespace]/[componentName]`
+          - `scope` be the closest configuration path
+          - `entry` be `[dir]/[namespace]/[componentName]/[componentName.{html|css|js|ts}]`
 
-        3.2.3 If is a `NpmModuleRecord`, validate that the npm package
+        3.2.3 Else if is a `NpmModuleRecord`, validate that the npm package
         is installed and can be resolved (by using Node `node_modules` 
         algorithm). Then `GoTo 1` recursively merging all the 
         ModuleRecordEntries collecting the scope forEach as npm 
         packages are traversed.
-
+3.2.4 Otheriwse throw an invalid ModuleRecord error
     3.3 Return all the ModuleRecordEntries collected.
 ```
 
@@ -347,7 +351,7 @@ Its important to understand the previous state of the art on how different platf
 
 ### Ecosystem
 
-Developers must learn about `lwc.config.json` for their package or component to be distributed discoverable and used by 3parties (via npm or other forms of registry).
+Developers must learn about `lwc.config.json` for their package or component to be distributed discoverable and used by third-parties (via npm or other forms of registry).
 
 Note that in the Salesforce platform developers already familiar with our metadata file which hold information about the component, altough this will be irrelevant for them.
 
--- a/text/0000-module-resolution.md
+++ b/text/0000-module-resolution.md
@@ -210,8 +210,7 @@ This type of resolution maps a given module specifier to a particular path withi
 `AliasModuleRecord` must contain `name` and a `path` keys.
 
 ::: note 
-    It is encouraged that the name specifier for the modules is of the form namespace/name.
-    Moreover the namespace doesn't necesarily have to match the path, altough it is recommended to follow that convention.
+  It is encouraged that the name specifier for the modules is of the form namespace/name. Moreover the namespace doesn't necesarily have to match the path, altough it is recommended to follow that convention.
 :::
 
 #### DirectoryModuleRecord
--- a/text/0000-module-resolution.md
+++ b/text/0000-module-resolution.md
@@ -78,7 +78,7 @@ Note that there are other non-standard formats, loaders and tools like UMD, Syst
 
 ### Node.js resolution
 
-If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is known as [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3rd party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
+If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is known as [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3rd party package how does that work? Well, Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
 
 All of this being said after years of very hard work, Node is finnaly addopting ES6 modules (which we will see in the next section) since LTS v12 and they are hoping to [fully transition to it soon](https://twitter.com/MylesBorins/status/1189618753065144322)
 
@@ -127,7 +127,7 @@ As a final note, remember that the resolution of ES6 can be done in multiple way
 
 Web Components are composed of multiple pieces (CSS, HTML and JS) and there is no canonical or standard way to bundle them natively yet. Although there are several proposals to standarize the way by which each of these pieces can interoperate and be integrated in the future (see [CSS modules](https://github.com/w3c/webcomponents/issues/759), [HTML templates](https://github.com/w3c/webcomponents/issues/839)), we must build a future-proof, standards compliant way to do so while we help the standardization pieces to land.
 
-Even when all these pieces become available (and without going into too much detail here), it is likely that we will have to do some preprocessing or bundling anyway, which mean that we will have to take over the resolution one way or another - whether at build time to collapse multiple pieces in on, or at runtime by using import-maps with some extra semantics on top. Moreover remember that a give package might contain many components (1:N) relationship, so we'll always have to have a configuration on where to find those components.
+Even when all these pieces become available (and without going into too much detail here), it is likely that we will have to do some preprocessing or bundling anyway, which means that we will have to take over the resolution one way or another - whether at build time to collapse multiple pieces in on, or at runtime by using import-maps with some extra semantics on top. Moreover remember that a give package might contain many components (1:N) relationship, so we'll always have to have a configuration on where to find those components.
 
 To summarize: We will always need an abstraction layer which dictates where a particular module specifier must be resolved from (adapted to different host environments).
 
@@ -186,7 +186,7 @@ In previous draft versions, `module` entries used to be just *strings* which had
 }
 ```
 
-On the example below, it will be impossible to know if `just-string` is the name of an npm package, or a name of a folder that contains components, hence creationg resolution ambiguity.
+On the example below, it will be impossible to know if `just-string` is the name of an npm package, or a name of a folder that contains components, hence creating resolution ambiguity.
 
 We considered other alternatives such as forcing specific prefix like `npm:package-name` or require a final slash (ex. `folder-name/`) for folders, but we believed that those were more cumbersome to learn and future-proof, so standarizing on an consistent object shape as ModuleRecord was a preferable and superior approach.
 
@@ -260,7 +260,7 @@ This type of resolution tells the resolver to find an npm package with that name
 
 ## Module resolution and exposure via NPM packages
 
-So far we have covered how are we going to resolve the modules, however we would want to be able to expose a subset of those modules 
+So far we have covered how we are going to resolve the modules, however we would want to be able to expose a subset of those modules 
 for other developers in the form of an npm package.
 
 ### Exposing modules
@@ -326,9 +326,10 @@ Here is the algoritm for module resolution (the instructions are not in a riguro
   an empty array.
   
   1. Load the LWC configuration file in the rootDir
-    1.1 Search for `lwc.config.js`, if it exist read it.
+    1.1 Search for `lwc.config.json`, if it exist read it.
     1.2 If `lwc.config.js` is not found search in `package.json`
-    (If both are found the `lwc.config.js` file takes precedence).
+    (If both are found the `lwc.config.json` file takes precedence).
+    1.3 Error if not config can't be resolved in rootDir
 
   2. Merge all the ModuleRecords found in the config with 
   the list of initModules. If an alias with the same name 
@@ -336,10 +337,10 @@ Here is the algoritm for module resolution (the instructions are not in a riguro
 
   3. For each ModuleRecord:
 
-    3.1  Validate that its a valid ModuleRecord type, skip it otherwise.
-    3.1.1 For `AliasModuleRecord` `path` and `name` keys must exist.
-    3.1.2 For `DirModuleRecord` the dir key must exist and the 
-    value must match an existing folder on the file system.
+    3.1  Validate that its a valid ModuleRecord type, throw Error otherwise.
+        3.1.1 For `AliasModuleRecord` `path` and `name` keys must exist.
+        3.1.2 For `DirModuleRecord` the dir key must exist and the 
+            value must match an existing folder on the file system.
     3.1.3 For `NpmModuleRecord` the npm key must exist and its value 
     must not have a leading `/` or `./`. Also the npm package must
     be resolvable from Node's perspective.
@@ -365,7 +366,8 @@ Here is the algoritm for module resolution (the instructions are not in a riguro
         algorithm). Then `GoTo 1` recursively merging all the 
         ModuleRecordEntries collecting the scope forEach as npm 
         packages are traversed.
-3.2.4 Otheriwse throw an invalid ModuleRecord error
+
+        3.2.4 Otheriwse throw an invalid ModuleRecord error
     3.3 Return all the ModuleRecordEntries collected.
 ```
 
@@ -377,6 +379,8 @@ the main difference is that we don't traverse upwards all module paths.
 ```
 Lets `importee` be module specifier to be resolved and `importer` the path of the module on 
 which is being imported from.
+
+0. If importee starts with `/` or `.`, exit early since importee does not require resolution
 
 1. Find the closest lwc.config.js or package.json that contains an "lwc" config by traversing 
 upwards in the file system.
@@ -416,6 +420,10 @@ This concept of scope matches the [current specification of import maps](https:/
 ## Resolving modules at build time.
 
 This algorithm can be used and integrated with tools like Webpack or rollup, in fact we have created both the `@lwc/module-resolver` and the `@lwc/rollup-plugin` packages that we officially support. Other projects in the ecosystem such as `create-lwc-app` also leverage those primitives.
+
+::: note
+    If the LWC module resolution algorithm doesn't match, tools might decide to fallback to standard Node resolution (or equivalent host resolution algorithm).
+::: 
 
 ## Alternatives
 
--- a/text/0000-module-resolution.md
+++ b/text/0000-module-resolution.md
@@ -275,20 +275,21 @@ Here is the algoritm for module resolution (the instructions are not in a riguro
 
     3.2 Resolve ModuleRecord entry:
 
-        3.2.1 If is an `AliasModuleRecord` validate the path and add create 
-        a ModuleRecordEntry with: 
+        3.2.1 If is an `AliasModuleRecord` validate the path and add
+        create a ModuleRecordEntry with:
           - `specifier` as the `name` value.
           - `scope` with the closest configuration path.
           - `entry` with the `path` value.
 
-        3.2.2 If is a `DirModuleRecord` validate the path and find of the 
-        modules that match the structure:
+        3.2.2 If is a `DirModuleRecord` validate the path and find of 
+        the modules that match the structure:
         [namespace]/[componentName]/[componentName.{html|css|js|ts}]
 
-        3.2.3 If is a `NpmModuleRecord`, validate that the npm package is 
-        installed and can be resolved (by using Node `node_modules` 
-        algorithm). `GoTo 1` recursively merging all the 
-        ModuleRecordEntries collecting the scope as packages are traversed.
+        3.2.3 If is a `NpmModuleRecord`, validate that the npm package
+        is installed and can be resolved (by using Node `node_modules` 
+        algorithm). Then `GoTo 1` recursively merging all the 
+        ModuleRecordEntries collecting the scope forEach as npm 
+        packages are traversed.
 
     3.3 Return all the ModuleRecordEntries collected.
 ```
--- a/text/0000-module-resolution.md
+++ b/text/0000-module-resolution.md
@@ -72,7 +72,7 @@ This means that from a developers perspective you just import the module you nee
 
 An important topic to understand first is how module resolution works today in different technologies and host environments.
 
-JavaScript did not had built-in support for modules, but the community has created impressive work-arounds over the years. The two most important (and unfortunately incompatible) standards are: CommonJS (used fundamentally by Node - with some extra semantics) and AMD (Asyncronous module definition), used by RequireJs and a lot of other bundlers.
+JavaScript did not have built-in support for modules, but the community has created impressive work-arounds over the years. The two most important (and unfortunately incompatible) standards are: CommonJS (used fundamentally by Node - with some extra semantics) and AMD (Asynchronous Module Definition), used by RequireJs and a lot of other bundlers.
 
 Note that there are other non-standard formats, loaders and tools like UMD, SystemJS, IIFE, etc, which are interesting for you know and understand but are out of the scope for this proposal.
 
@@ -86,7 +86,7 @@ Just to put all of this in a simple example:
 
 ```js
 // This will be resolved by looking a node_modules/ folder "foo-bar" 
-// recursively upwards, and the check the package.kson to find the entry point. 
+// recursively upwards, and the check the package.json to find the entry point.
 const x = require('foo-bar') // This is using CommonJS syntax
 
 // This indicates that we are exporting a default function 
@@ -99,7 +99,10 @@ module.exports = function () { return x() }; // This is using CommonJS syntax
 
 At the end of July 2014, TC39 (Javascript standarization comitee) had another meeting, during which the last details of the ECMAScript 6 (ES6) module syntax were finalized. [Here is the specification text](https://www.ecma-international.org/ecma-262/6.0/#sec-imports).
 
-The fundamental goal for ECMAScript 6 modules was to create a standard format that all users from the previous formats where comfortable and happy about, and lacked none of the caveats and issues - and since was part of the standar a lot of things could be improved like 1) The structure which can be statically analyzed (for static checking, optimizations, etc.) 2) Support for cyclic dependencies in a much better way.
+The fundamental goal for ECMAScript 6 modules was to create a standard format that all users from the previous formats were comfortable and happy about, and lacked none of the caveats and issues - and since was part of the standard a lot of things could be improved like
+
+1. The structure which can be statically analyzed (for static checking, optimizations, etc.)
+2. Better support for cyclic dependencies
 
 Here is how ES6 modules look like:
 
@@ -122,9 +125,9 @@ As a final note, remember that the resolution of ES6 can be done in multiple way
 
 ## Why we need our own module resolution
 
-Web Components are composed of multiple pieces (CSS, HTML and JS) and there is no canonical or standard way to bundle them natively yet (altough there are several proposals to standarize the way by which each of this pieces can interoperate and be integrated in the future see [CSS modules](https://github.com/w3c/webcomponents/issues/759), [HTML templates](https://github.com/w3c/webcomponents/issues/839)) hence we must build a future-proof, standard compliant way to do so while we help the standarization pieces to land.
+Web Components are composed of multiple pieces (CSS, HTML and JS) and there is no canonical or standard way to bundle them natively yet. Although there are several proposals to standarize the way by which each of these pieces can interoperate and be integrated in the future (see [CSS modules](https://github.com/w3c/webcomponents/issues/759), [HTML templates](https://github.com/w3c/webcomponents/issues/839)), we must build a future-proof, standards compliant way to do so while we help the standardization pieces to land.
 
-Even when we have all this pieces become available (and without going into too much detail here), it is likely that we will have to do some preprocessing or bundling anyway, which mean that we will have to take over the resolution one way or another - wether is build time to collapse mutliple pieces in on, or at runtime by using import-maps with some extra semantics on top. Moreover remember that a give package might contain many components (1:N) relationship, so will always have to have a configuration on where to find those components.
+Even when all these pieces become available (and without going into too much detail here), it is likely that we will have to do some preprocessing or bundling anyway, which mean that we will have to take over the resolution one way or another - whether at build time to collapse multiple pieces in on, or at runtime by using import-maps with some extra semantics on top. Moreover remember that a give package might contain many components (1:N) relationship, so we'll always have to have a configuration on where to find those components.
 
 To summarize: We will always need an abstraction layer which dictates where a particular module specifier must be resolved from (adapted to different host environments).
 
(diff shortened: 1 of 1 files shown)


text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+import button from "ui/button"; // should resolve against ui/button@1.0.0
+```
+
+This means that from a developers perspective you just import the module you need.
+
+## State of the art
+
+An important topic to understand first is how module resolution works today in different technologies and host environments.
+
+JavaScript did not had built-in support for modules, but the community has created impressive work-arounds over the years. The two most important (and unfortunately incompatible) standards are: CommonJS (used fundamentally by Node - with some extra semantics) and AMD (Asyncronous module definition), used by RequireJs and a lot of other bundlers.
```
tedconn:
```suggestion
JavaScript did not have built-in support for modules, but the community has created impressive work-arounds over the years. The two most important (and unfortunately incompatible) standards are: CommonJS (used fundamentally by Node - with some extra semantics) and AMD (Asynchronous Module Definition), used by RequireJs and a lot of other bundlers.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+
+Note that there are other non-standard formats, loaders and tools like UMD, SystemJS, IIFE, etc, which are interesting for you know and understand but are out of the scope for this proposal.
+
+### Node.js resolution
+
+If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is know nas [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
```
tedconn: If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is known as [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3party package how does that work? Well Node decided to implement its own resolution by looking at the content of the `node_modules` folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+
+All of this being said after years of very hard work, Node is finnaly addopting ES6 modules (which we will see in the next section) since LTS v12 and they are hoping to [fully transition to it soon](https://twitter.com/MylesBorins/status/1189618753065144322)
+
+Just to put all of this in a simple example:
+
+```js
+// This will be resolved by looking a node_modules/ folder "foo-bar" 
+// recursively upwards, and the check the package.kson to find the entry point.
```
tedconn: // recursively upwards, and the check the package.json to find the entry point.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+// to whoever is requesting this module
+module.exports = function () { return x() }; // This is using CommonJS syntax
+
+```
+
+### ES6 Modules and browser resolution
+
+At the end of July 2014, TC39 (Javascript standarization comitee) had another meeting, during which the last details of the ECMAScript 6 (ES6) module syntax were finalized. [Here is the specification text](https://www.ecma-international.org/ecma-262/6.0/#sec-imports).
+
+The fundamental goal for ECMAScript 6 modules was to create a standard format that all users from the previous formats where comfortable and happy about, and lacked none of the caveats and issues - and since was part of the standar a lot of things could be improved like 1) The structure which can be statically analyzed (for static checking, optimizations, etc.) 2) Support for cyclic dependencies in a much better way.
```
tedconn:
```suggestion
The fundamental goal for ECMAScript 6 modules was to create a standard format that all users from the previous formats were comfortable and happy about, and lacked none of the caveats and issues - and since was part of the standard a lot of things could be improved like

1. The structure which can be statically analyzed (for static checking, optimizations, etc.)
2. Better support for cyclic dependencies
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+export function namedExport() { return foo(); }
+
+// re-export a module
+export { buzz };
+```
+
+Along with that there is another proposal (Stage 4) for [dynamic imports](https://tc39.es/proposal-dynamic-import/) to allow modules to be loaded lazily.
+
+As a final note, remember that the resolution of ES6 can be done in multiple ways.
+  - Browsers have implemented a URL schema base way of resolving modules with [import maps](https://wicg.github.io/import-maps/) (We will be basing our implementation on this)
```
tedconn: This is still in draft stage right? I can't find any information on implementation other than this link: https://chromestatus.com/feature/5315286962012160 and some tools (the biggest of which is SystemJS). Do you have any more info on that?

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+  - Node.js as stated before has integrated ES6 modules along side the old one (see link above).
+  - Tools like Webpack or RollupJs allow you to configure and customize the way we load bundle and resolve modules both run build-time and run-time.
+
+## Why we need our own module resolution
+
+Web Components are composed of multiple pieces (CSS, HTML and JS) and there is no canonical or standard way to bundle them natively yet (altough there are several proposals to standarize the way by which each of this pieces can interoperate and be integrated in the future see [CSS modules](https://github.com/w3c/webcomponents/issues/759), [HTML templates](https://github.com/w3c/webcomponents/issues/839)) hence we must build a future-proof, standard compliant way to do so while we help the standarization pieces to land.
```
tedconn:
```suggestion
Web Components are composed of multiple pieces (CSS, HTML and JS) and there is no canonical or standard way to bundle them natively yet. Although there are several proposals to standarize the way by which each of these pieces can interoperate and be integrated in the future (see [CSS modules](https://github.com/w3c/webcomponents/issues/759), [HTML templates](https://github.com/w3c/webcomponents/issues/839)), we must build a future-proof, standards compliant way to do so while we help the standardization pieces to land.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+
+Even when we have all this pieces become available (and without going into too much detail here), it is likely that we will have to do some preprocessing or bundling anyway, which mean that we will have to take over the resolution one way or another - wether is build time to collapse mutliple pieces in on, or at runtime by using import-maps with some extra semantics on top. Moreover remember that a give package might contain many components (1:N) relationship, so will always have to have a configuration on where to find those components.
```
tedconn:
```suggestion
Even when all these pieces become available (and without going into too much detail here), it is likely that we will have to do some preprocessing or bundling anyway, which mean that we will have to take over the resolution one way or another - whether at build time to collapse multiple pieces in on, or at runtime by using import-maps with some extra semantics on top. Moreover remember that a give package might contain many components (1:N) relationship, so we'll always have to have a configuration on where to find those components.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+
+To summarize: We will always need an abstraction layer which dictates where a particular module specifier must be resolved from (adapted to different host environments).
+
+## Detailed design
+
+In order to resolve modules in LWC that work on different host environments (Salesforce platform, Node.js, npm, git, et all) we will be relying on our own LWC configuration file (similar approach as import-maps, babel, webpack rollup or package.json).
+
+### LWC configuration
+
+A configuration file will be located at the root of the project, optionally at the component namespace level and in the future (this is yet to be spec out) at the module bundle level, where specificitiy will be resolved from inner to outer (merging upwards - similar to node_modules).
```
tedconn:
```suggestion
A configuration file will be located at the root of the project, optionally at the component namespace level and in the future (this is yet to be spec out) at the module bundle level, where specificity will be resolved from inner to outer (merging upwards - similar to node_modules).
 ```

File: text/0000-module-resolution.md [hunk 9]
```diff
@@ -0,0 +1,340 @@
...
+{
+    "modules:" [
+        /* list with modules records goes here */
+    ]
+}
+```
+
+Its important that **any plugin, tool or project honor this configuration**, otherwise the resolution will not work correctly in most environments.
+
+Moreover, note that this type of configuraiton is not something new, many tools use an equivalent configuration (babel, lerna, rollup, webpack, ...)
```
tedconn:
```suggestion
Moreover, note that this type of configuration is not something new, many tools use an equivalent configuration (babel, lerna, rollup, webpack, ...)
```

tedconn: Great point. It was feeling kind of weird to have to have this explicitly defined but everyone does it.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+    "modules": [
+       {
+            "name": "ui/button",
+            "path": "src/modules/ui/button/button.js"
+        }        
+    ]
+}
+```
+
+`AliasModuleRecord` must contain `name` and a `path` keys.
```
tedconn: Since the other `ModuleRecord` types have a defined key, can't we match the import map spec for an alias and do
```json5
{
  "ui/button": "src/modules/ui/button/button.js"
}
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+
+#### DirectoryModuleRecord
+
+This type of resolution allows to point to to a folder that contain LWC modules.
+Its important to note that within that folder the structure must be very specific, it must contain a namespace folder, then a folder per module bundle (moduleName) adnd them the moduleEntry file that matches the name of the bundle (the entry must be of type .css, .html, .js or .ts)
```
tedconn: > Its important to note that within that folder the structure must be very specific

Also important to note that the compiler does not define the same requirements. I understand why you did not want to impose unnecessary restrictions but maybe we should consider formalizing that.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+        a ModuleRecordEntry with: 
+          - `specifier` as the `name` value.
+          - `scope` with the closest configuration path.
+          - `entry` with the `path` value.
+
+        3.2.2 If is a DirModuleRecord validate the path and find of the 
+        modules that match the structure:
+        [namespace]/[componentName]/[componentName.{html|css|js|ts}]
+
+        3.2.3 If is a NpmModuleRecord, validate that the package is installed
```
tedconn: > validate that the package is installed

By looking in `node_modules`, correct?

Also, we should add that we should validate that it is a bare module specifier and has no leading `/` or `./`.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+#### Future ModuleRecord types
+
+We will not discard adding another types of module resolutions, like for example, find modules directly from git or even from a url.
+This schema allows us to add 
+
+## Scoping
+
+Its an invariant of the module resolution system to allow for multiple versions of multiple components within a particular application configuration. As the modules are resolved, we keep track of the context on which those modules appear, basically who is the importer (and its path). You can check again the [basic example](#basic_example) on the top.
+
+This concept of scope matches the [current specification of import maps](https://github.com/WICG/import-maps#multiple-versions-of-the-same-module) which will allows us to create the equivalent resolution for the browser.
```
tedconn:
```suggestion
This concept of scope matches the [current specification of import maps](https://github.com/WICG/import-maps#multiple-versions-of-the-same-module) which will allow us to create the equivalent resolution for the browser.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,340 @@
...
+This algorithm can be used and integrated with tools like Webpack or rollup, in fact we have created both the `@lwc/module-resolver` and the `@lwc/rollup-plugin` packages that we officially support. Other projects in the ecosystem such as `create-lwc-app` also leverage those primitives.
+
+## Alternatives
+
+We discussed different module resolution alteratives previously, but to summarize:
+
+  - We need a custom module resolution to: 
+    - Support the 1:N relationship on packages
+    - Have different build steps
+    - Align with standards in different host environments
```
tedconn: Does NPM have a concept of package "types"? Like a package.json primitive which allows you to say "this has ui components or this has a service" I can see this resolution mechanism becoming a thing

text/0000-module-resolution.md
tedconn: You mentioned above the configuration specificity from inside out. If that part were spec'ed and supported you could allow `glob` pattern matching instead of relying on a particular directory structure. _For example although Jest has it easy because you can match on a file name like `*.test.js`, it still manages to basically load any test module from any matching directory efficiently_

text/0000-module-resolution.md
diervo: The problem is having a  "free-form" key has the potential to have a conflicts in the future, plus it can create ambiguity. I would rather have an extra key: `alias`: true alongside the other two.

text/0000-module-resolution.md
diervo: I will add some extra clarification here for the first comment:

text/0000-module-resolution.md
diervo: Yes that is done in 3.1. I will add some steps there

text/0000-module-resolution.md
diervo: No, but thats why we have the `lwc` config key in there.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+Along with that there is another proposal (Stage 4) for [dynamic imports](https://tc39.es/proposal-dynamic-import/) to allow modules to be loaded lazily.
+
+As a final note, remember that the resolution of ES6 can be done in multiple ways.
+  - Browsers have implemented a URL schema base way of resolving modules with [import maps](https://wicg.github.io/import-maps/) (We will be basing our implementation on this)
+  - Node.js as stated before has integrated ES6 modules along side the old one (see link above).
+  - Tools like Webpack or RollupJs allow you to configure and customize the way we load bundle and resolve modules both run build-time and run-time.
+
+## Why we need our own module resolution
```
pmdartus:
```suggestion
## Why we need our own module resolution?
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+
+#### AliasModuleRecord
+
+This type of resolution maps a given module specifier to a particular path within the root of that module.
+
+```json
+{
+    "modules": [
+       {
+            "name": "ui/button",
```
pmdartus: I think it should also be stated in this sections that: 
* the `name` property should have the following form `<ns>/<name>` where `ns` is a valid module name and `name` is a valid module name.
* the directory structure doesn't have any baring on the module name and namespace. In the following example, the component namespace will be `ui` and the component name will be `button`.

```json
{
  "modules": [
    {
      "name": "ui/button",
      "path": "src/module/foo/bar/bar.js"
    }
  ]
}
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+Here is the algoritm for module resolution (the instructions are not in a rigurous specification format for brevity and time)
+
+```
+  Let `rootDir` be the initial path for the application, `initModules` 
+  the list of ModuleRecords to be resolved and ModuleRecordEntryList 
+  an empty array.
+  
+  1. Load the LWC configuration file in the rootDir
+    1.1 Search for `lwc.config.js`, if it exist read it.
+    1.2 If lwc.config.js is not found search in package.json
```
pmdartus:
```suggestion
    1.2 If `lwc.config.js` is not found search in `package.json`
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+    (If both are found the configuration file has precedence).
```
pmdartus:
```suggestion
    (If both are found the `lwc.config.js` file takes precedence).
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+        ModuleRecordEntries collecting the scope forEach as npm 
+        packages are traversed.
+
+    3.3 Return all the ModuleRecordEntries collected.
+```
+
+#### Future ModuleRecord types
+
+We will not discard adding another types of module resolutions, like for example, find modules directly from git or even from a url.
+This schema allows us to add
```
pmdartus: This sentence is not done.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+- `ModuleRecord`: An entry module to be resolved.
+- `[Type]ModuleRecord`: And ModuleRecord with specific schema and semantics.
+- `ResolvedModuleRecord` The result of resolving a ModuleRecord.
+- Scope: The closest importer of a particular ModuleRecord.
+- Specifier: Name of the module that can be used to reference a module.
+- Entry: The entry point for a particular module bundle.
+
+### Ecosystem
+
+Developers must learn about `lwc.config.json` for their package or component to be distributed discoverable and used by 3parties (via npm or other forms of registry).
```
pmdartus:
```suggestion
Developers must learn about `lwc.config.json` for their package or component to be distributed discoverable and used by third-parties (via npm or other forms of registry).
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+
+    3.2 Resolve ModuleRecord entry:
+
+        3.2.1 If is an `AliasModuleRecord` validate the path and add
+        create a ModuleRecordEntry with:
+          - `specifier` as the `name` value.
+          - `scope` with the closest configuration path.
+          - `entry` with the `path` value.
+
+        3.2.2 If is a `DirModuleRecord` validate the path and find of
```
pmdartus:
```suggestion
        3.2.2 Else if is a `DirModuleRecord` validate the path and find of 
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+        create a ModuleRecordEntry with:
+          - `specifier` as the `name` value.
+          - `scope` with the closest configuration path.
+          - `entry` with the `path` value.
+
+        3.2.2 If is a `DirModuleRecord` validate the path and find of 
+        the modules that match the structure:
+        [namespace]/[componentName]/[componentName.{html|css|js|ts}]
+
+        3.2.3 If is a `NpmModuleRecord`, validate that the npm package
```
pmdartus:
```suggestion
        3.2.3 Else if is a `NpmModuleRecord`, validate that the npm package
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+        is installed and can be resolved (by using Node `node_modules` 
+        algorithm). Then `GoTo 1` recursively merging all the 
+        ModuleRecordEntries collecting the scope forEach as npm 
+        packages are traversed.
+
```
pmdartus:
```suggestion
3.2.4 Otheriwse throw an invalid ModuleRecord error
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+        3.2.1 If is an `AliasModuleRecord` validate the path and add
+        create a ModuleRecordEntry with:
+          - `specifier` as the `name` value.
+          - `scope` with the closest configuration path.
+          - `entry` with the `path` value.
+
+        3.2.2 If is a `DirModuleRecord` validate the path and find of 
+        the modules that match the structure:
+        [namespace]/[componentName]/[componentName.{html|css|js|ts}]
```
pmdartus:
```suggestion
        [namespace]/[componentName]/[componentName.{html|css|js|ts}] and create a 
        ModuleRecordEntry with:
          - `specifier` be `[namespace]/[componentName]`
          - `scope` be the closest configuration path
          - `entry` be `[dir]/[namespace]/[componentName]/[componentName.{html|css|js|ts}]`
```

This section of the spec is missing in which order the extension should be resolved. For example if a module contains `test.html`, `test.css`, `test.js` and `test.ts` which file is considered the entry point. I am proposing the following order:
- `.js`
- `.ts`
- `.html`
- `.css`

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+
+This means that from a developers perspective you just import the module you need.
+
+## State of the art
+
+An important topic to understand first is how module resolution works today in different technologies and host environments.
+
+JavaScript did not have built-in support for modules, but the community has created impressive work-arounds over the years. The two most important (and unfortunately incompatible) standards are: CommonJS (used fundamentally by Node - with some extra semantics) and AMD (Asynchronous Module Definition), used by RequireJs and a lot of other bundlers.
+
+Note that there are other non-standard formats, loaders and tools like UMD, SystemJS, IIFE, etc, which are interesting for you know and understand but are out of the scope for this proposal.
```
nrobertdehault:
```suggestion
Note that there are other non-standard formats, loaders and tools like UMD, SystemJS, IIFE, etc, which are interesting for you to know and understand but are out of the scope for this proposal.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Node.js resolution
+
+If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is know nas [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
```
nrobertdehault:
```suggestion
If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is known as [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
```

nrobertdehault:
```suggestion
If you ever used Node, you know that in order to import a module you need to use the `require()` method, and if you want to export something to some other module or file you must use `module.exports`. That is know nas [CommonJS format](https://en.wikipedia.org/wiki/CommonJS). Now if you want to import a 3rd party package how does that works? Well Node decided to implement its own resolution by looking at the content of a the node_modules folder based on some dependencies defined on `package.json`. You can look at the full algorithm [here](https://nodejs.org/docs/latest/api/modules.html#modules_all_together).
```

text/0000-module-resolution.md
tedconn: No, it was correct before.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+`NpmModuleRecord` must contain `npm` key.
+
+## Algorithm
+
+Here is the algoritm for module resolution (the instructions are not in a rigurous specification format for brevity and time)
+
+```
+  Let `rootDir` be the initial path for the application, `initModules`
```
pmdartus: I was thinking more about this algorithm this morning and I don't think the original constraints forcing us to return a list of `ModuleRecordEntry` still holds with this proposal. 

Previously we needed to resolve ALL the module in advance because it was **too expensive** to search through the `node_modules` directory to find out what `package.json` has a `lwc` property. In order to get arround this issue, we exposed this `resolveModules` method that should ideally run once (before all the test, when starting a new compilation, ...). However this approach suffers from the fact that if you add a new LWC module when the application is running, you will have to restart the application for it to be consumed. The proposed algorithm also suffers from the same issue.

Now that each module have an explicit dependency mapping, there is a smaller set of possibility to resolve a given LWC module name. IMO we should deprecate the `resolveModules` method and expose a new `resolveModule` method that give a module name (`foo/bar`) and an importer (`./src/modules/ui/button/button.html`) should return a module entry point (`./nodes_modules/my-pkg/src/src/modules/foo/bar/bar.js`). This approach also aligns better with the [Node.js module resolution](https://nodejs.org/api/modules.html#modules_all_together).

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+Here is the algoritm for module resolution (the instructions are not in a rigurous specification format for brevity and time)
+
+```
+  Let `rootDir` be the initial path for the application, `initModules` 
+  the list of ModuleRecords to be resolved and ModuleRecordEntryList 
+  an empty array.
+  
+  1. Load the LWC configuration file in the rootDir
+    1.1 Search for `lwc.config.js`, if it exist read it.
```
dbleakley: should this be lwc.config.json?

diervo: yep

text/0000-module-resolution.md
```diff
@@ -244,14 +250,74 @@ This type of resolution tells the resolver to find an npm package with that name
...
+In order to expose a module via npm package, it must be explicity declared in the `expose` array property.
+
+```json
+{
+    "modules": [
+       { 
+           "dir" : "src/modules"
+       }     
+    ],
+    "expose": [
```
rsalvador: maybe use "export" instead

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+This RFC formalizes the module resolution for LWC. It defines the way that we can resolve modules in all host environments (node, browser, rollup, webpack) and in a multi-author multi-version way that is compliant with the latest mental models and specificatons in the different platforms.
+
+## Motivation
+
+As the LWC ecosystem organically grows both internally at Salesforce and externally as OSS, we want to provide and encourage a way to share web components, npm packages and other artifacts, regardless of the host environment, technologies or tools used.
+
+## Basic example
+
+Let's look at a very simple scenario: We have a repository of LWC components we want to open source. We want to use other 3party packages as the foundation for my components, in this case `@ui/components@2.0.0` and `fancy-components@1.0.0`. Its important to note that the `fancy-components` package is using internally `@ui/components@1.0.0`, which is a different version of the package I'm using.
```
nrobertdehault:
```suggestion
Let's look at a very simple scenario: We have a repository of LWC components we want to open source. We want to use other 3party packages as the foundation for our components, in this case `@ui/components@2.0.0` and `fancy-components@1.0.0`. Its important to note that the `fancy-components` package is using internally `@ui/components@1.0.0`, which is a different version of the package I'm using.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+Here is how the structure of the repo looks like  you can take a look at [a real implementation implementation and test here](https://github.com/salesforce/lwc/pull/1602/files#diff-5cfc908f1b15979cf59306304092ddb5):
```
nrobertdehault:
```suggestion
Here is how the structure of the repo looks like  you can take a look at [a real implementation and test here](https://github.com/salesforce/lwc/pull/1602/files#diff-5cfc908f1b15979cf59306304092ddb5):
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+Its important that **any plugin, tool or project honor this configuration**, otherwise the resolution will not work correctly in most environments.
+
+Moreover, note that this type of configuration is not something new, many tools use an equivalent configuration (babel, lerna, rollup, webpack, ...)
+
+### Module resolution configuration
+
+The `modules` key on the configuration will hold an Array of `ModuleRecords`. A module record is an object with a specific shape depending on the type of module record it holds. The type will depend on the mode of resolution, we will be defining three types of `ModuleRecords` for now:
+  - AliasModuleRecord: Define an specifier that points to a specific module.
```
nrobertdehault:
```suggestion
  - AliasModuleRecord: Define a specifier that points to a specific module.
```

text/0000-module-resolution.md
pmdartus: I support the `expose` proposal. It maps with the `isExposed` property used in the `metadata.xml` on the platform.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+Here is the algoritm for module resolution (the instructions are not in a rigurous specification format for brevity and time)
+
+```
+  Let `rootDir` be the initial path for the application, `initModules` 
+  the list of ModuleRecords to be resolved and ModuleRecordEntryList 
+  an empty array.
+  
+  1. Load the LWC configuration file in the rootDir
+    1.1 Search for `lwc.config.js`, if it exist read it.
+    1.2 If `lwc.config.js` is not found search in `package.json`
```
pmdartus: Add 1.3 step to error out if no config can be resolved at the rootDir.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+    (If both are found the `lwc.config.js` file takes precedence).
+
+  2. Merge all the ModuleRecords found in the config with 
+  the list of initModules. If an alias with the same name 
+  exist pick initModules over the ones resolved in the config.
+
+  3. For each ModuleRecord:
+
+    3.1  Validate that its a valid ModuleRecord type, skip it otherwise.
```
pmdartus: If the `ModuleRecord` is invalid we should throw here instead of silently skipping it.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+    3.1.1 For `AliasModuleRecord` `path` and `name` keys must exist.
```
pmdartus: Add indentation for steps `3.1.1` to `3.1.3`

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+          - `specifier` be `[namespace]/[componentName]`
+          - `scope` be the closest configuration path
+          - `entry` be `[dir]/[namespace]/[componentName]/[componentName.{html|css|js|ts}]`
+
+        3.2.3 Else if is a `NpmModuleRecord`, validate that the npm package
+        is installed and can be resolved (by using Node `node_modules` 
+        algorithm). Then `GoTo 1` recursively merging all the 
+        ModuleRecordEntries collecting the scope forEach as npm 
+        packages are traversed.
+3.2.4 Otheriwse throw an invalid ModuleRecord error
```
pmdartus: Missing indentation.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+::: warning
+The module resolver will throw if it detects multiple modules with the same name
+:::
+
+
+## Algorithms
+
+### Preload all modules
```
pmdartus: When is this algorithm used? I don't see any other reference to the algo in this proposal.

text/0000-module-resolution.md
pmdartus: This algorithm still suffers from the same issue. For performance reasons, I don't think we should eagerly resolve all the modules in advance.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+### Algorithm to resolve a module in iterable way
+
+This algorithm is very similar to the [NodeJS require](https://nodejs.org/api/modules.html#modules_all_together),
+the main difference is that we don't traverse upwards all module paths. 
+
+```
+Lets `importee` be module specifier to be resolved and `importer` the path of the module on 
+which is being imported from.
+
+1. Find the closest lwc.config.js or package.json that contains an "lwc" config by traversing
```
pmdartus: Add a step to handle the case where no valid configuration is present. I think in this case this algorithm should just return `null`.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+```
+Lets `importee` be module specifier to be resolved and `importer` the path of the module on 
+which is being imported from.
+
+1. Find the closest lwc.config.js or package.json that contains an "lwc" config by traversing 
+upwards in the file system.
+
+2. Match the current importee againts a ModuleRecord defined in the modules array.
+    2.1 If the ModuleRecord is not found throw since resolution
```
pmdartus: The sentence is not complete.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+the main difference is that we don't traverse upwards all module paths. 
+
+```
+Lets `importee` be module specifier to be resolved and `importer` the path of the module on 
+which is being imported from.
+
+1. Find the closest lwc.config.js or package.json that contains an "lwc" config by traversing 
+upwards in the file system.
+
+2. Match the current importee againts a ModuleRecord defined in the modules array.
```
pmdartus: I don't understand this sentence. What do you mean exactly by match in this context?

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+    2.1 If the ModuleRecord is not found throw since resolution
+
+3. For each ModuleRecord
+    3.1 Match the `importee` to a module specififer:
+        3.1.1 If is an `AliasModuleRecord` validate the path and add
+        match the ModuleRecordEntry with:
+          - `specifier` as the `importee` value.
+          - `entry` with the `importer` value.
```
pmdartus:
```suggestion
          - `entry` with the `specifier` value.
```

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+        3.1.2 Else if is a `DirModuleRecord` validate the path and find of 
+        the modules that match the structure:
+        [namespace]/[componentName]/[componentName.{html|css|js|ts}] and match a 
+        ModuleRecordEntry with:
+          - `specifier` be `[namespace]/[componentName]`
+          - `entry` be `[dir]/[namespace]/[componentName]/[componentName.{html|css|js|ts}]`
```
pmdartus: I would be great to add a note about the entry resolution order: `js` > `ts` > `html` > `css`.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+        3.1.3 Else if is a `NpmModuleRecord`, find lwc.config.js or package.json lwc config
+            3.1.3.1 Check that the component is exposed.
```
pmdartus: We should break this algo into 2 sub-algos: resolving the config and matching the `importee` against the config. The step 3.1.3 should invoke the matching algo recursively until the module gets resolved.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+## Scoping
+
+Its an invariant of the module resolution system to allow for multiple versions of multiple components within a particular application configuration. As the modules are resolved, we keep track of the context on which those modules appear, basically who is the importer (and its path). You can check again the [basic example](#basic_example) on the top.
+
+This concept of scope matches the [current specification of import maps](https://github.com/WICG/import-maps#multiple-versions-of-the-same-module) which will allow us to create the equivalent resolution for the browser.
+
+## Resolving modules at build time.
+
+This algorithm can be used and integrated with tools like Webpack or rollup, in fact we have created both the `@lwc/module-resolver` and the `@lwc/rollup-plugin` packages that we officially support. Other projects in the ecosystem such as `create-lwc-app` also leverage those primitives.
```
pmdartus: I think it might be interesting to note that if the LWC module resolution algorithm doesn't match, tools might decide to fallback to the standard Node resolution.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+    3.3 Return all the ModuleRecordEntries collected.
+```
+
+### Algorithm to resolve a module in iterable way
+
+This algorithm is very similar to the [NodeJS require](https://nodejs.org/api/modules.html#modules_all_together),
+the main difference is that we don't traverse upwards all module paths. 
+
+```
+Lets `importee` be module specifier to be resolved and `importer` the path of the module on
```
pmdartus: We should add a step here to check that the `importee` is a bare import (identifier with no `/` or `./` prefix), otherwise, the algorithm should early exit.

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+#### Future ModuleRecord types
+
+We will not discard adding another types of module resolutions, like for example, find modules directly from git or even from a url.
+This schema allows us to add new types as we see fit.
+
+## Scoping
+
+Its an invariant of the module resolution system to allow for multiple versions of multiple components within a particular application configuration. As the modules are resolved, we keep track of the context on which those modules appear, basically who is the importer (and its path). You can check again the [basic example](#basic_example) on the top.
+
+This concept of scope matches the [current specification of import maps](https://github.com/WICG/import-maps#multiple-versions-of-the-same-module) which will allow us to create the equivalent resolution for the browser.
```
blittle: Has there been additional thought into what dynamic runtime resolution would look like?

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+#### Future ModuleRecord types
+
+We will not discard adding another types of module resolutions, like for example, find modules directly from git or even from a url.
+This schema allows us to add new types as we see fit.
+
+## Scoping
+
+Its an invariant of the module resolution system to allow for multiple versions of multiple components within a particular application configuration. As the modules are resolved, we keep track of the context on which those modules appear, basically who is the importer (and its path). You can check again the [basic example](#basic_example) on the top.
```
apapko: Its -> It's

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+## Detailed design
+
+In order to resolve modules in LWC that work on different host environments (Salesforce platform, Node.js, npm, git, et all) we will be relying on our own LWC configuration file (similar approach as import-maps, babel, webpack rollup or package.json).
+
+### LWC configuration
+
+A configuration file will be located at the root of the project, optionally at the component namespace level and in the future (this is yet to be spec out) at the module bundle level, where specificity will be resolved from inner to outer (merging upwards - similar to node_modules).
+
+At the package root level there are two ways on which you can add a configuration file: Creating a `lwc.config.json` or adding an `lwc` key to package.json (this is mostly for simplicity and easy of use - specially when publishing a package). Here are some examples:
```
apapko: an 'lwc' -> a 'lwc'

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+```json
+{
+    "modules": [
+        "just-string"
+    ]
+}
+```
+
+On the example below, it will be impossible to know if `just-string` is the name of an npm package, or a name of a folder that contains components, hence creationg resolution ambiguity.
```
apapko: an npm -> a npm

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+
+We considered other alternatives such as forcing specific prefix like `npm:package-name` or require a final slash (ex. `folder-name/`) for folders, but we believed that those were more cumbersome to learn and future-proof, so standarizing on an consistent object shape as ModuleRecord was a preferable and superior approach.
```
apapko: an consistent -> a consistent

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+    └── ui/ (namespace)
+        └── button/ (moduleName)
+            └── button.js (moduleEntry)
+```
+
+`DirectoryModuleRecord` must contain `dir` key.
+
+#### NpmModuleRecord
+
+This type of resolution tells the resolver to find an npm package with that name and resolve the modules it might have inside. This process can be recursive (see example at the top).
```
apapko: an npm -> a npm

text/0000-module-resolution.md
```diff
@@ -0,0 +1,458 @@
...
+    ]
+}
+```
+
+`NpmModuleRecord` must contain `npm` key.
+
+## Module resolution and exposure via NPM packages
+
+So far we have covered how are we going to resolve the modules, however we would want to be able to expose a subset of those modules 
+for other developers in the form of an npm package.
```
apapko: a npm

text/0000-module-resolution.md
apapko: an "lwc" -> a "lwc"

text/0000-module-resolution.md
diervo: Dynamic resolution is built on top or importmaps which its generation should use the same algorithm.

diervo: LWR team already implements this @dbleakley can show you

text/0000-module-resolution.md
tariqrafique: i think `an npm` is a more common phrase. see npm docs: https://docs.npmjs.com/configuring-your-registry-settings-as-an-npm-enterprise-user

text/0000-module-resolution.md
manuel-jasso: So can I use either `lwc.config.json` or `lwc.config.js`? If so, what is their precedence? (for inspiration, compare to [eslint configs](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)).

PR #21, opened 2019-11-15
RFC: spellcheck and alike global attributes

jodarove:


jodarove:
fyi/ @sfdciuie @caridy @pmdartus

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Support for spellcheck and alike global attributes.
+status: DRAFTED
+created_at: 2019
```
caridy:
```suggestion
created_at: 2019/11/15
updated_at: 2019/11/20
pr: https://github.com/salesforce/lwc-rfcs/pull/21
```

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Support for spellcheck and alike global attributes.
```
caridy:
```suggestion
title: Spellcheck Attribute
```

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,145 @@
...
+status: DRAFTED
+created_at: 2019
+---
+
+# Support for spellcheck and alike global attributes.
+
+### related Issue: https://github.com/salesforce/lwc/issues/1595
```
caridy:
```suggestion
## Backpointers

 * related issue: https://github.com/salesforce/lwc/issues/1595
```

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Summary
+
+When setting the attribute `spellcheck` of a **custom element** in a component template, the rendered element does not reflect the attribute properly. This proposal introduces support `spellcheck` and alike global attributes when used in LWC custom elements.
```
caridy: reflect the attribute properly => reflect the attribute into the dom.

caridy: support `spellcheck` => support for `spellcheck`

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,164 @@
...
+This case is more complicated because of all the invariants that it involves, but we can group them in 2 general cases that can be handled differently for custom elements:
+
+```html
+<template>
+  <c-foo spellcheck={expr}></c-foo> <!-- should reflect the attribute depending on the value of {expr} -->
+</template>
+```
+
+1. `expr == null` (`expr` result is `null` or `undefined`) This is the more complicated code-path, and the one that deserves more attention: 
+    1. Given that we always set the property, the attribute is always going to be set to `true` or `false`
```
pmdartus: >  Given that we always set the property

What if we were to change this invariant, to handle this case? Only set it the spellcheck property when it is not `null` or `undefined` ?

jodarove: then what will happen in such case that the expression changes from `true`/`false` to `null` or `undefined`?

pmdartus: In this case we would invoke `removeAttribute`.

caridy: no @pmdartus, remember that the main point here is that it MUST be a prop because we don't know if you have redefined that in your component via `@api spellcheck`, in which case you, as the author of the component, are in control.

pmdartus: Yes, you are right.

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,164 @@
...
+created_at: 2019/11/15
+updated_at: 2019/11/20
+pr: https://github.com/salesforce/lwc-rfcs/pull/21
+---
+
+# Support for spellcheck and alike global attributes.
+
+## Summary
+
+When setting the attribute `spellcheck` of a **custom element** in a component template, the rendered element does not reflect correctly the attribute into the DOM. This proposal introduces support for `spellcheck` and alike global attributes when used in LWC custom elements.
```
ekashida:
```suggestion
When setting the `spellcheck` attribute of a **custom element** in a component template to `"false"`, the rendered attribute value is `"true"`, which is incorrect.
```

pmdartus:
We spent some time discussing with @jodarove and @ekashida about this proposal. The [static value](https://github.com/salesforce/lwc-rfcs/blob/jodarove/support-spellcheck-and-alike-ga/text/0000-spellcheck-global-attribute.md#static-values) part of it is not controversial. However, we are not entirely convinced by the [dynamic value](https://github.com/salesforce/lwc-rfcs/blob/jodarove/support-spellcheck-and-alike-ga/text/0000-spellcheck-global-attribute.md#dynamic-values-via-template-expressions) part of this proposal. 

Since it is highly unlikely that the `spellcheck` will be set via a template expression we decided to only implement the static part of it for now, and reconsider the implementation of the dynamic part if we feel the need for it in the future.

Note: As part of this discussion we experimented with different frameworks and we noticed that Vue 3 also suffers from the same issue: https://codepen.io/sh7dm/pen/JjjbMVe

text/0000-spellcheck-global-attribute.md
```diff
@@ -0,0 +1,164 @@
...
+
+```html
+<template>
+  <c-foo spellcheck={expr}></c-foo> <!-- should reflect the attribute depending on the value of {expr} -->
+</template>
+```
+
+1. `expr == null` (`expr` result is `null` or `undefined`) This is the more complicated code-path, and the one that deserves more attention: 
+    1. Given that we always set the property, the attribute is always going to be set to `true` or `false` 
+    2. Since the spellcheck attribute is **true-by-default**, for `undefined` or `null`, we will set it to boolean `true`.
```
ekashida: The default value seems to depend on a combination of user agent and element.

PR #22, opened 2019-12-24
wip: portals

sfdciuie:
https://rfcs.lwc.dev/rfcs/lwc/0000-portals

### DIFF
--- a/text/0000-portals.md
+++ b/text/0000-portals.md
@@ -56,7 +56,7 @@ One non-trivial example of such usage:
 </body>
 ```
 where `x-dialog`'s body is
-```
+```html
 <template>
   <template lwc:portal={portalName}>
     <section role="dialog" ...>
--- a/text/0000-portals.md
+++ b/text/0000-portals.md
@@ -1,0 +1,104 @@
+---
+title: portals proposal
+status: DRAFTED
+created_at: 2019-12-23
+updated_at: 2019-12-23
+pr:
+---
+
+# Short Title
+Out of tree DOM insertion of custom elements via portals.
+
+## Summary
+Introduce a mechanism in LWC for inserting and managing an element outside of its natural dom tree.
+
+## Basic example
+
+The proposal suggests two new APIs:
+1) A way to define where in the DOM the element would be inserted (eg. `lwc:portal-definition` attribute on an element)
+2) A way to define the destination portal of an element (eg. `lwc:portal` attribute on an element)
+
+One non-trivial example of such usage:
+
+```xhtml
+<body>
+  <div lwc:portal-definition="global-modals"></div>
+
+  <div lwc:portal-definition="popups"></div>
+  ...
+  <div id="tabs">
+    <div id="tab-1">
+      <div lwc:portal-definition="tab1-modals"></div>
+
+      <div>
+         <x-dialog lwc:portal="global-modals">Global Modal</x-foo>
+         <x-dialog lwc:portal="tab1-modals">Scoped Modal</x-foo>
+      </div>  
+    </div>  
+    <div id="tab-2">
+      <div lwc:portal-definition="tab2-modals"></div>
+
+      <div>
+         <x-dialog lwc:portal="global-modals">Global Modal</x-foo>
+         <x-dialog lwc:portal="tab2-modals">Scoped Modal</x-foo>
+      </div>  
+    </div>  
+  </div>
+</body>
+```
+since we wouldn't want for a user to specify the portal each and every time one would use a dialog, more realistically -- `<x-dialog>` would have a default behaviour based on the @Context, which can provide whether scoping is enabled or not, the name of the scoped portal and the name of the global portal.
+
+## Motivation
+
+Position of a node in the DOM determines its stacking context with the component having no 
+real ability to change it besides creating a new stacking context. This becomes problematic
+when one wants to present a popup or a modal dialog that appears on top of other existing
+elements. 
+
+Having a custom element be inserted out of its natural dom tree position would enable
+one to have greater control over the stacking context, allowing a predictalbe popup
+and modal implementation (including faux-modals that are scoped to tabs and are not per-se global).
+
+## Detailed design
+
+(not an actual detail designed just adding some possible requirements)
+- Instead of being inserted at its natural position in the template, the element would be inserted in the destination portal,if one exists.
+- When the host element is disconnected, the portalled element should also be disconnected. 
+- The portalled component should still be @Context aware 
+
+## Drawbacks
+
+- new API
+- additional complexity in LWC
+
+## Security
+
+- The portalled element content should not be accessible by anyone but the originally owning host element.
+
+## Alternatives
+
+In case of modals, some other strategies have been considered:
+1. Event based creation of a modal with a modal manager defined at some parent node. 
+Given that a modal could have up to 3 slots (header, default, footer), the ergonomics of usage become very cumbersome.
+Moreover event-based modal creation is inferior to `<dialog>` like experience.
+2. Exposing top layer by the browsers. This would allow one to effectively declare a new top layer and override any 
+existing stacking context. It's unlikely that proposal would support scoping.
+
+## Adoption strategy
+
+It's not a breaking change, but it does introduce new API, adoption would initially be from component libraries with
+support for modals and popups.
+
+## How we teach this
+
+This would allow implementation of components that can ignore the current stacking context. Teaching should be around what components can benefit from this.
+
+## Unresolved questions
+
+1. How would `querySelector` work ?
+2. Is the definiton of the portal immutable (can you change its name, what if it gets removed?)
+3. Can the destional portal be changed ?
+4. What if the portal doesn't exist
+5. What if there are more than 1 portal definition with the same name
+6. Should there be implicit/reserved portals ? (eg. `body` which always implies the `<body>` element of the document) 
+7. Should the events bubble out of the portalled component to the host element ?
--- a/text/0000-portals.md
+++ b/text/0000-portals.md
@@ -21,6 +21,8 @@ This becomes problematic when one wants to present a popup/tooltip/dialog/dropdo
 
 Having a custom element be inserted out of its natural dom tree hierarchy would enable one to have greater control over the stacking context, allowing a predictalbe popup and modal implementation (including faux-modals that are scoped to tabs and are not per-se global).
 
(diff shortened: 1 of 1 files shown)


text/0000-portals.md
```diff
@@ -0,0 +1,108 @@
...
+- An element can't both define a portal, and define a destination portal (ie. no `lwc:portal-define` and `lwc:portal` attributes simultaneously). 
+- Instead of being inserted at its natural position in the template, the element would be inserted in the destination portal, if one exists.
+- When an element's destination portal doesn't exist, but later is inserted in the DOM, any elements that have already rendered stay in their original location, as if the portal doesn't exist.
+- When the host element is disconnected, the portalled element should also be disconnected. 
+- The portalled component should still be `@Context` aware 
+
+## Drawbacks
+
+- new API
+- additional complexity in LWC
```
caridy: we should add the problems with the events propagation, it messes up with the mental model of the composition for authors.

text/0000-portals.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Security
+
+- The portalled element content should not be accessible by anyone but the originally owning host element.
```
caridy: this will require some gymnastics that we should specify in details. Most likely, every dialog will require a placeholder inside the definition element, and that placeholder can have a shadow root to insert the dialog inside it.

text/0000-portals.md
caridy: another one is the CSS issues... if the dialog is moved out, the scoped styles will not work as expected, unless we do some magic in LWC.

text/0000-portals.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Alternatives
+
+In case of dialogs, some other strategies have been considered:
+1. Event based creation of a dialog with a dialog manager defined at some parent node. 
+Given that a modal could have up to 3 slots (header, default, footer), the ergonomics of usage become very cumbersome.
+Moreover event-based modal creation is inferior to `<dialog>` like experience.
+2. Exposing top layer by the browsers. This would allow one to effectively declare a new top layer and override any
```
caridy: @JanMiksovsky has some questions here that we should clarify, starting with "why is scoping a thing?"

text/0000-portals.md
```diff
@@ -0,0 +1,108 @@
...
+It's not a breaking change, but it does introduce new API, adoption would initially be from component libraries with
+support for modals and popups.
+
+## How we teach this
+
+This would allow implementation of components that can ignore the current stacking context. Teaching should be around what components can benefit from this.
+
+## Unresolved questions
+
+1. How would `querySelector` work ?
```
caridy: it will not work if you ask me :)

text/0000-portals.md
```diff
@@ -0,0 +1,108 @@
...
+
+The proposal suggests two new APIs:
+1) A way to define where in the DOM the element would be inserted (eg. `lwc:portal-definition` attribute on an element)
+2) A way to define the destination portal of an element (eg. `lwc:portal` attribute on an element)
+
+One non-trivial example of such usage:
+
+```xhtml
+<body>
+  <div lwc:portal-definition="global-modals"></div>
```
caridy: this is problematic because until now, `lwc:` directives are only useful inside a template files for LWC components. How is this different from using an `id`? it is already guaranteed to be unique (or should be), it must be direct child of body (or should be), and it is easy to find from inside a component (just make a global search for it).

text/0000-portals.md
```diff
@@ -0,0 +1,108 @@
...
+
+  <div lwc:portal-definition="popups"></div>
+  ...
+  <div id="tabs">
+    <div id="tab-1">
+      <div lwc:portal-definition="tab1-modals"></div>
```
caridy: lets challenge this assumption first...

if it turns out that we need the scoping, then we can find ways for the scope to be resolved somehow.

text/0000-portals.md
sfdciuie: i think it would be nice if events would propagate out as expected, (ie. "forwarded" to their original location), regarding CSS as discussed this potentially can be resolved if we restrict the portalled elements to be top-rooted in the template.

text/0000-portals.md
sfdciuie: I added some clarifying notes

text/0000-portals.md
sfdciuie: good point, I agree, it's better to just go with global ids.

text/0000-portals.md
```diff
@@ -0,0 +1,110 @@
...
+This would allow implementation of components that can ignore the current stacking context. Teaching should be around what components can benefit from this.
+
+## Unresolved questions
+
+1. How would `querySelector` work ?
+2. Is the definiton of the portal immutable (can you change its name, what if it gets removed?)
+3. Can the destional portal be changed ?
+4. What if the portal doesn't exist
+5. What if there are more than 1 portal definition with the same name
+6. Should there be implicit/reserved portals ? (eg. `body` which always implies the `<body>` element of the document)
```
Templarian: Seeing as body is the most likely default it makes sense that `lwc:portal` with no value would be set to `<body>`.

text/0000-portals.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Unresolved questions
+
+1. How would `querySelector` work ?
+2. Is the definiton of the portal immutable (can you change its name, what if it gets removed?)
+3. Can the destional portal be changed ?
+4. What if the portal doesn't exist
+5. What if there are more than 1 portal definition with the same name
+6. Should there be implicit/reserved portals ? (eg. `body` which always implies the `<body>` element of the document) 
+7. Should the events bubble out of the portalled component to the host element ?
```
Templarian: React Portals behave like normal child elements. Can't imagine LWC wouldn't want the same relationship. We kind of use the idea of dispatchEvent to communicate with the notifications popovers in the corner I think.

caridy: No, this is not accurate IIRC, in react you have two layers, the component layer, and the dom layer, the dom layer, when using portals, will not propagate out of the portal definition.

ekashida:
I'm a little confused about the mental model for this. For the owner of the portalled element, is the fact that the component is rendered somewhere else supposed to be an implementation detail? Should it be transparent to the owner? It seems like a lot of the questions around querySelector and event-forwarding hinges on that distinction.

text/0000-portals.md
```diff
@@ -0,0 +1,110 @@
...
+
+Having a custom element be inserted out of its natural dom tree hierarchy would enable one to have greater control over the stacking context, allowing a predictalbe popup and modal implementation (including faux-modals that are scoped to tabs and are not per-se global).
+
+Side-note: Modal scoping means that a modal would be tied to a tab (the backdrop only appears inside the tab, and the modal is centered against the tab content) and only blocks the content of the tab, and nothing outside of it (focus trapping still works as with a true modal, but there is a key that allows one to escape the modal). When using Lightning Console in Salesforce, all modals by default are scoped, so this is a very important use-case to capture.
+
+## Basic example
+
+The proposal suggests one new API -- a way to define the destination of an element (position in the DOM where the component would be rendered in). The api in question is a new attribute -- `lwc:portal` that accepts a string that should match the global document id of the element that would contain the rendered element.
```
ekashida: Instead of `id` which will get mangled, we should use another directive like `lwc:portal:id` or something along those lines.

sfdciuie: @ekashida that's what I had initially (can see it in the commit history), this indeed is a problem, so I may need to go back to the original version with the portal definition

ekashida: @sfdciuie thanks for pointing out the comment history. Regarding `lightning-portal`, I guess we should avoid component dependencies for framework features.

After reading that discussion and thinking more about it, I agree there are no issues if the portal root lives outside of any component as a direct descendant of `document.body`. Also, if the need to bypass id-mangling ever arises, we can just add a new template directive to do so.

text/0000-portals.md
```diff
@@ -0,0 +1,110 @@
...
+since we wouldn't want for a user to specify the portal each and every time one would use a dialog, more realistically -- `<x-dialog>` would have a default behaviour based on the `@Context`, which can provide whether scoping is enabled or not, the name of the scoped portal and the name of the global portal. This dynamic choice would indicate that support for dynamically specifying the portal may be needed.
+
+## Detailed design
+
+(not an actual detailed design, just adding some possible requirements)
+
+- If a custom element has `lwc:portal` attribute on it, and its value is an existing global id, instead of being inserted at its natural position in the template, the element would be inserted in the element with the defined global id.
+- When an element's destination portal doesn't exist, but later is inserted in the DOM, any elements that have already rendered stay in their original location, as if the portal doesn't exist.
+- When the host element is disconnected, the portalled element should also be disconnected. 
+- The portalled component should still be `@Context` aware
```
jbenallen: The current implementation for serving context to wire adapters uses custom DOM events. Therefore, without work there, context wire adapters will not receive context through their host source but through their portal container's context. It's possible to change this to support serving context through the source host instead of from the portal's context, but this needs to be debated whether or not that is correct.

I tend to think that context should be provided through the DOM composition, not the host source. It would still be possible to do this differently through manual context passing from host->portalled component, but doing it by default may be problematic.

text/0000-portals.md
jbenallen: I think it would be simpler and cleaner if the mental model for a portalled element is exactly the same as creating an element and appending it as a child of the portal target. The host source of the portalled element can do the following:

1) Determine the composition/tree structure of the portalled content
2) Provide the property input values for that composition (reactivity still works as expected; property changes on the host cause rehydration on the portalled content, wherever it actually lives in the document)
3) Add declarative event handlers, which are bound to methods on the host element's class, to any of the portalled elements

In this mental model, I wonder if `lwc:portal` should only apply to `template` elements. The semantics may fit better there. The contents of the `<template lwc:portal=...>...</template>` are inert (don't exist) until used. When the template is connected, the portal is located. If it exists, the contents of the template are hydrated and inserted into the portal.

Component developers assume the **contents** of a `template` with an `lwc:portal` directive are never actually in its shadow DOM, so `querySelector` from its root won't work, and event propagation won't work. This would be far more natural and unsurprising if portalled contents must be inside a `template`. That template provides the hinge for the portal itself.

Incidentally, this pattern could allow for the host component to still obtain a reference to the `template` with the `lwc:portal`. If it were important to allow the host component to query the portalled composition, it could do so from the template's reference, and LWC could provide an abstraction that fulfills that by querying the actual content where it has been inserted elsewhere in the document.

PR #23, opened 2020-01-07
RFC: LWC server runtime proposal

pmdartus:
The LWC engine has been designed from the beginning to run in an environment with access to the DOM APIs. To accommodate server-side rendering (SSR) requirements, we need a way to decouple the engine from the DOM APIs. The existing `@lwc/engine` will be replaced by 3 new packages: 
- `@lwc/engine-core` exposes platform-agnostic APIs and will be used by the different runtime packages to share the common logic.
- `@lwc/engine-dom` exposes LWC APIs available on the browser.
- `@lwc/engine-server` exposes LWC APIs used for server-side rendering.

----

[Rendered](https://github.com/salesforce/lwc-rfcs/blob/pmdartus/server-runtime/text/0112-server-runtime.md)

### DIFF
--- a/text/0112-server-runtime.md
+++ b/text/0112-server-runtime.md
@@ -1,0 +1,331 @@
+---
+title: LWC server runtime
+status: DRAFTED
+created_at: 2020-01-06
+updated_at: 2020-01-06
+pr: (leave this empty until the PR is created)
+---
+
+# LWC server runtime
+
+## Summary
+
+The LWC engine has been designed from the beginning to run in an environment with access to the DOM APIs. To accommodate server-side rendering (SSR) requirements, we need a way to decouple the engine from the DOM APIs. The existing `@lwc/engine` will be replaced by 3 new packages: 
+- `@lwc/engine-core` exposes platform-agnostic APIs and will be used by the different runtime packages to share the common logic.
+- `@lwc/engine-dom` exposes LWC APIs available on the browser.
+- `@lwc/engine-server` exposes LWC APIs used for server-side rendering.
+
+## Scope
+
+LWC SSR support is a broad subject, this proposal only focusses on a subset of it. This proposal covers the following topics:
+
+- How to evaluate the existing LWC engine module in a javascript context without DOM access.
+- What LWC APIs should be exposed in both environments.
+- How the LWC engine share the same logic when evaluated in both environments.
+
+The following topics are considered out of the scope of this proposal:
+
+- How the LWC component tree rendered on the server gets serialized and sent to the browser.
+- How the serialized component tree gets rehydrated when processed by the browser.
+
+## Motivation
+
+The first step in enabling SSR support on LWC is to be able to run the engine in a non-DOM-enabled environment. It is currently impossible to evaluate the engine in such environment due to its hard dependencies on DOM APIs.
+
+While we want to share as much logic as possible between the different environments and make SSR as transparent as possible for component authors, the runtime behavior and APIs exposed by the engine greatly differs between environments. In this regard, it makes sense to break up the current `@lwc/engine` package into multiple packages specifically tailored for each environment.
+
+Distributing multiple versions of the LWC engine offers the capability to expose different APIs depending on the execution context. On one hand, the current implementation of `createElement` with reaction hooks doesn't make sense in Node.js since there is no DOM tree to attach the created element. In the same way, the `renderToString` method doesn't make sense to ship in a browser.
+
+## Detailed design
+
+The existing `@lwc/engine` will be replaced by 3 new packages:
+
+- `@lwc/engine-core`: Core logic shared by the different runtime including the rendering engine and the reactivity mechanism. This package should never be consumed directly in an application. This package is agnostic on the underlying rendering medium. The only provide APIs for building custom renderers.
+- `@lwc/engine-dom`: Runtime that can be used to render LWC component trees in a DOM environment. This package is built on top of `@lwc/engine-core`.
+- `@lwc/engine-server`: Runtime used that can be used to render LWC component trees as strings. This package is built on top of `@lwc/engine-core`.
+
+The existing `lwc` package stays untouched and will be used to distribute the different versions of the engine. From the developer perspective, the experience should be identical.
+
+**`c/app/app.js`:**
+
+```js
+import { LightningElement } from "lwc";
+
+export default class App extends LightningElement {}
+```
+
+**`client.js`:**
+
+```js
+import { createElement } from "lwc"; // Aliased to @lwc/engine-dom
+import App from "c/app";
+
+const app = createElement("c-app", { is: App });
+document.body.appendChild(app);
+```
+
+**`server.js`:**
+
+```js
+import { createElement, renderToString } from "lwc"; // Aliased to @lwc/engine-server
+import App from "c/app";
+
+const app = createElement("c-app", { is: App });
+const str = renderToString(app);
+
+console.log(str);
+```
+
+### `@lwc/engine-core`
+
+This packages exposes the following platform-agnostic APIs:
+
+- `LightningElement`
+- `api`
+- `track`
+- `readonly`
+- `wire`
+- `setFeatureFlag`
+- `getComponentDef`
+- `isComponentConstructor`
+- `getComponentConstructor`
+- `unwrap`
+- `registerTemplate`
+- `registerComponent`
+- `registerDecorators`
+- `sanitizeAttribute`
+
+The DOM APIs used by the rendered engine are injected by the runtime depending on the environment. A list of all the DOM APIs the engine depends upon can be found in the [DOM APIs usage](#dom-apis-usage) section in the Appendix.
+
+### `@lwc/engine-dom`
+
+This package exposes the following APIs:
+
+- `createElement` + reaction hooks
+- `buildCustomElementConstructor`
+- `isNodeFromTemplate`
+
+This package injects the native DOM APIs into the `@lwc/engine-core` rendering engine.
+
+### `@lwc/engine-server`
+
+This package exposes the following APIs:
+
+- `createElement(name: string, options: { is: typeof LightningElement }): ServerHTMLElement`: This method creates a new LWC component tree. It follows the same signature as the `createElement` API from `@lwc/engine-dom`. Instead of returning a native `HTMLElement`, this method returns a `ServerHTMLElement` with the public properties, aria reflected properties and HTML global attributed.
+- `renderToString(element: ServerHTMLElement): string`: This method creates an LWC component tree synchronously and serialize it to string. It accepts a single parameter, a `ServerHTMLElement` returned by `createElement` and returns the serialized string.
+
+This package injects mock DOM APIs in the `@lwc/engine-core` rendering engine. Those DOM APIs produces a lightweight DOM structure that can be serialized into a string by the `renderToString` method. As described in the Appendix, this package is also in charge of attaching on the global object a mock `CustomEvent`.
+
+## Drawbacks
+
+- Requires a complete rewrite of the current `@lwc/engine`
+- Might require substantial changes to the existing tools (jest preset, rollup plugins, ...) to load the right engine depending on the environment.
+
+## Alternatives
+
+### Using a DOM implementation in JavaScript before evaluating the engine
+
+Prior the evaluation of the LWC engine, we would evaluate a DOM implementation write in JavaScript (like [jsdom](https://github.com/jsdom/jsdom) or [domino](https://github.com/fgnass/domino)). The compelling aspect of this approach is that it requires almost no change to the engine to work.
+
+There are multiple drawbacks with this approach:
+
+- Performance: The engine only relies on a limited set of well-known APIs, leveraging a full DOM implementation for SSR would greatly reduce the SSR performance.
+- Predictability: By attaching the DOM interfaces on the global object, those APIs are not only exposed to the engine but also to the component author. Exposing such APIs to the component author might bring unexpected behavior when component code runs on the server.
+
+## Adoption strategy
+
+TBD
+
+## How we teach this
+
+TBD
+
+## Unresolved questions
+
+**How to load the right version of the engine when resolving the `lwc` identifier?**
+
+For example, when running in Node.js `lwc` might resolve to `@lwc/engine-server` for SSR purposes, but also to `@lwc/engine-dom` when running test via jest and jsdom.
+
+One way to solve this would be to import the platform-specific APIs from the appropriate module: `createElement` from `@lwc/engine-dom`, `createElement` and `renderToString` from `@lwc/engine-server` and use. Platform-agnostic APIs should be imported from `@lwc` (which becomes an alias to `@lwc/engine-core`).
+
+---
+
+## Appendix
+
+### DOM APIs usage
+
+We can break-down the current LWC DOM usages into 3 different categories:
+
+- [DOM constructors used by the engine](#dom-constructors-used-by-the-engine)
+- [DOM methods and accessors used by the engine](#dom-methods-and-accessors-used-by-the-engine)
+- [DOM constructors used by component authors](#dom-constructors-used-by-component-authors)
+
+#### DOM constructors used by the engine
+
+The engine currently relies on the following DOM constructors during evaluation and at runtime:
+
+##### `HTMLElement`
+
+- **DOM usage:** Used by the engine to extract the descriptors and reassign them to the `LightningElement` prototype.
+- **SSR usage:** 🔵 NOT REQUIRED
+- **Observations:** We can create a hard-coded list of all the needed accessors: HTML global attributes, aria reflection properties and HTML the good part.
+
+##### `ShadowRoot`
+
+- **DOM usage:** Used by the engine to traverse the DOM tree to find the error boundary and create the component call stack.
+- **SSR usage:** 🔶 MIGHT BE REQUIRED
+- **Observations:** Framework implementing the error boundary concept (like React or Vue), use equivalent of the LWC `VM.owner` back-pointer to traverse the component tree up instead of relying on the DOM tree. This discussion might require another RFC.
+
+#### DOM methods and accessors used by the engine
+
+On top of this the engine also rely on the following DOM methods and accessors:
+
+##### `EventTarget.prototype.dispatchEvent()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.dispatchEvent()`
+- **SSR usage:** 🔴 REQUIRED
+- **Observations:** Components may dispatch event once connected.
+
+##### `EventTarget.prototype.addEventListener()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.addEventListener()`. Used by the rendering engine to handle `on*` directive from the template.
+- **SSR usage:** 🔴 REQUIRED
+
+##### `EventTarget.prototype.removeEventListener()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.removeElementListener()`.
+- **SSR usage:** 🔴 REQUIRED
+
+##### `Node.prototype.appendChild()`
+
+- **DOM usage:** Used by the upgrade mechanism, synthetic shadow styling and to enforce restrictions.
+- **SSR usage:** 🔵 NOT REQUIRED
+- **Observations:** This API can be replaced by `Node.prototype.insertBefore(elm, null)`.
+
+##### `Node.prototype.insertBefore()`
+
+- **DOM usage:** Used by the upgrade mechanism, by the rendering engine, and to enforce restrictions.
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `Node.prototype.removeChild()`
+
+- **DOM usage:** Used by the upgrade mechanism, by the rendering engine, and to enforce restrictions.
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `Node.prototype.replaceChild()`
+
+- **DOM usage:** Used by the upgrade mechanism, and to enforce restrictions
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `Node.prototype.parentNode` (getter)
+
+- **DOM usage:** Used to traverse the DOM tree
+- **SSR Usage:** 🔶 MIGHT BE REQUIRED
+- **Observations:** Depending on how the error boundary is implemented, we might be able to get rid of this API.
+
+##### `Element.prototype.hasAttribute()`
+
+- **DOM usage:** Used by the aria reflection polyfill
+- **SSR Usage:** 🔵 NOT REQUIRED
+- **Observations:** For SSR, we will not need this polyfill
+
+##### `Element.prototype.getAttribute()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.getAttribute()`. Used by the aria reflection polyfill
+- **SSR Usage:** 🔶 MIGHT BE REQUIRED
+- **Observations:** We might use `Element.prototype.getAttributeNS(null, name)` to reduce the API surface needed by the engine.
+
+##### `Element.prototype.getAttributeNS()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.getAttributeNS()`
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `Element.prototype.setAttribute()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.setAttribute()`. Used by the rendering engine, for synthetic shadow styling and by the aria reflection polyfill.
+- **SSR Usage:** 🔶 MIGHT BE REQUIRED
+- **Observations:** We might use `Element.prototype.setAttributeNS(null, name)` to reduce the API surface needed by the engine.
+
+##### `Element.prototype.setAttributeNS()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.setAttributeNS()`. Used by the rendering engine.
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `Element.prototype.removeAttribute()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.removeAttribute()`. Used by the rendering engine, for the synthetic shadow styling and by the aria reflection polyfill
+- **SSR Usage:** 🔶 MIGHT BE REQUIRED
+- **Observations:** We might use `Element.prototype.removeAttributeNS(null, name)` to reduce the API surface needed by the engine.
+
+##### `Element.prototype.removeAttributeNS()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.removeAttributeNS()`.
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `Element.prototype.getElementsByTagName()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.getElementsByTagName()`.
+
+- **SSR Usage:** 🔵 NOT REQUIRED
+- **Observations:** This method is only exposed to select elements from the component light DOM, which is only available from the `renderedCallback`. Since SSR doesn't run `renderedCallback`, we can always returns an empty `HTMLCollection` when running on the server.
+
+##### `Element.prototype.getElementsByClassName()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.getElementsByClassName()`.
+- **SSR Usage:** 🔵 NOT REQUIRED
+- **Observations:** This method is only exposed to select elements from the component light DOM, which is only available from the `renderedCallback`. Since SSR doesn't run `renderedCallback`, we can always returns an empty `HTMLCollection` when running on the server.
+
+##### `Element.prototype.querySelector()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.querySelector()`. Used to enforce restrictions.
+- **SSR Usage:** 🔵 NOT REQUIRED
+- **Observations:** This method is only exposed to select elements from the component light DOM, which is only available from the `renderedCallback`. Since SSR doesn't run `renderedCallback`, we can always returns an empty `NodeList` when running on the server.
+
+##### `Element.prototype.querySelectorAll()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.querySelectorAll()`. Used to enforce restrictions.
+- **SSR Usage:** 🔵 NOT REQUIRED
+- **Observations:** This method is only exposed to select elements from the component light DOM, which is only available from the `renderedCallback`. Since SSR doesn't run `renderedCallback`, we can always returns an empty `NodeList` when running on the server.
+
+##### `Element.prototype.getBoundingClientRect()`
+
+- **DOM usage:** Exposed via `LightningElement.prototype.getBoundingClientRect()`.
+- **SSR Usage:** 🔵 NOT REQUIRED
+- **Observations:** Running a layout engine during SSR, is a complex task that will not bring much values to component authors. Returning an empty `DOMRect` (where all the properties are set to `0`), might be the best approach here.
+
+##### `Element.prototype.classList` (getter)
+
+- **DOM usage:** Exposed via `LightningElement.prototype.classList`. Used by the rendering engine.
+- **SSR Usage:** 🔴 REQUIRED
+
+##### `ShadowRoot.prototype.host` (getter)
+
+- **DOM usage:** Used to traverse the DOM.
+- **SSR Usage:** 🔶 MIGHT BE REQUIRED
+- **Observations:** Depending on how the error boundary is implemented, we might be able to get rid of this API.
+
+##### `ShadowRoot.prototype.innerHTML` (setter)
+
+- **DOM usage:** Used to reset the shadow root.
+- **SSR Usage:** 🔵 NOT REQUIRED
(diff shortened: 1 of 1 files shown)


text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Appendix
+
+### DOM APIs usage
+
+We can break-down the current LWC DOM usages into 3 different categories:
+
+- [DOM constructors used by the engine](#dom-constructors-used-by-the-engine)
+- [DOM methods and accessors used by the engine](#dom-methods-and-accessors-used-by-the-engine)
+- [DOM constructors used by component authors](#dom-constructors-used-by-component-authors)
```
caridy: component authors might use regular APIs as well, not just constructors

pmdartus: What APIs are you referring to with "regular APIs"? I don't see any other APIs except the one listed in this section that might be involved in SSR.

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+## Summary
+
+The LWC engine has been designed from the beginning to run in an environment with access to the DOM APIs. To accommodate server-side rendering (SSR) requirements, we need a way to decouple the engine from the DOM APIs. The existing `@lwc/engine` will be replaced by 3 new packages: 
+- `@lwc/engine-core` exposes platform-agnostic APIs and will be used by the different runtime packages to share the common logic.
+- `@lwc/engine-dom` exposes LWC APIs available on the browser.
+- `@lwc/engine-server` exposes LWC APIs used for server-side rendering.
+
+## Scope
+
+LWC SSR support is a broad subject, this proposal only focusses on a subset of it. This proposal covers the following topics:
```
tedconn:
```suggestion
LWC SSR support is a broad subject, this proposal only focuses on a subset of it. This proposal covers the following topics:
```

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+- How to evaluate the existing LWC engine module in a javascript context without DOM access.
+- What LWC APIs should be exposed in both environments.
+- How the LWC engine share the same logic when evaluated in both environments.
```
tedconn:
```suggestion
- How the LWC engine share the same logic when evaluated in both environments.
```

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+The following topics are considered out of the scope of this proposal:
+
+- How the LWC component tree rendered on the server gets serialized and sent to the browser.
```
tedconn: By serialized we mean "as HTML", right?

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+- How the serialized component tree gets rehydrated when processed by the browser.
+
+## Motivation
+
+The first step in enabling SSR support on LWC is to be able to run the engine in a non-DOM-enabled environment. It is currently impossible to evaluate the engine in such environment due to its hard dependencies on DOM APIs.
+
+While we want to share as much logic as possible between the different environments and make SSR as transparent as possible for component authors, the runtime behavior and APIs exposed by the engine greatly differs between environments. In this regard, it makes sense to break up the current `@lwc/engine` package into multiple packages specifically tailored for each environment.
+
+Distributing multiple versions of the LWC engine offers the capability to expose different APIs depending on the execution context. On one hand, the current implementation of `createElement` with reaction hooks doesn't make sense in Node.js since there is no DOM tree to attach the created element. In the same way, the `renderToString` method doesn't make sense to ship in a browser.
```
tedconn: > with reaction hooks 

that's the key part, right? Since it does make sense to create an element on the server?

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Detailed design
+
+The existing `@lwc/engine` will be replaced by 3 new packages:
+
+- `@lwc/engine-core`: Core logic shared by the different runtime including the rendering engine and the reactivity mechanism. This package should never be consumed directly in an application. This package is agnostic on the underlying rendering medium. The only provide APIs for building custom renderers.
```
tedconn:
```suggestion
- `@lwc/engine-core`: Core logic shared by the different runtime including the rendering engine and the reactivity mechanism. This package should never be consumed directly in an application. This package is agnostic on the underlying rendering medium. The only provide APIs for building custom renderers.
```

text/0112-server-runtime.md
pmdartus: Yes

text/0112-server-runtime.md
pmdartus: The reaction hook is the key part here you are right, reaction hooks is the way we use to get notified when an element is inserted or removed from the DOM. For environments without DOM, hooks are unnecessary.

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+- `buildCustomElementConstructor`
+- `isNodeFromTemplate`
+
+This package injects the native DOM APIs into the `@lwc/engine-core` rendering engine.
+
+### `@lwc/engine-server`
+
+This package exposes the following APIs:
+
+- `createElement(name: string, options: { is: typeof LightningElement }): ServerHTMLElement`: This method creates a new LWC component tree. It follows the same signature as the `createElement` API from `@lwc/engine-dom`. Instead of returning a native `HTMLElement`, this method returns a `ServerHTMLElement` with the public properties, aria reflected properties and HTML global attributed.
```
tedconn:
```suggestion
- `createElement(name: string, options: { is: typeof LightningElement }): ServerHTMLElement`: This method creates a new LWC component tree. It follows the same signature as the `createElement` API from `@lwc/engine-dom`. Instead of returning a native `HTMLElement`, this method returns a `ServerHTMLElement` with the public properties, aria reflected properties and HTML global attributes.
```

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+- `renderToString(element: ServerHTMLElement): string`: This method creates an LWC component tree synchronously and serialize it to string. It accepts a single parameter, a `ServerHTMLElement` returned by `createElement` and returns the serialized string.
```
tedconn:
```suggestion
- `renderToString(element: ServerHTMLElement): string`: This method creates an LWC component tree synchronously and serializes it to HTML. It accepts a single parameter, a `ServerHTMLElement` returned by `createElement` and returns the serialized string.
```

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+This package injects mock DOM APIs in the `@lwc/engine-core` rendering engine. Those DOM APIs produces a lightweight DOM structure that can be serialized into a string by the `renderToString` method. As described in the Appendix, this package is also in charge of attaching on the global object a mock `CustomEvent`.
```
tedconn:
```suggestion
This package injects mock DOM APIs in the `@lwc/engine-core` rendering engine. Those DOM APIs produces a lightweight DOM structure that can be serialized into a string by the `renderToString` method. As described in the Appendix, this package is also in charge of attaching on the global object a mock `CustomEvent`.
```

tedconn: > Those DOM APIs produce a lightweight DOM structure

Which DOM implementation do you plan to use for that?

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Drawbacks
+
+- Requires a complete rewrite of the current `@lwc/engine`
+- Might require substantial changes to the existing tools (jest preset, rollup plugins, ...) to load the right engine depending on the environment.
```
tedconn:
```suggestion
- Might require substantial changes to the existing tools (jest preset, rollup plugins, etc.) to load the right engine depending on the environment.
```

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+## Alternatives
+
+### Using a DOM implementation in JavaScript before evaluating the engine
+
+Prior the evaluation of the LWC engine, we would evaluate a DOM implementation write in JavaScript (like [jsdom](https://github.com/jsdom/jsdom) or [domino](https://github.com/fgnass/domino)). The compelling aspect of this approach is that it requires almost no change to the engine to work.
+
+There are multiple drawbacks with this approach:
+
+- Performance: The engine only relies on a limited set of well-known APIs, leveraging a full DOM implementation for SSR would greatly reduce the SSR performance.
+- Predictability: By attaching the DOM interfaces on the global object, those APIs are not only exposed to the engine but also to the component author. Exposing such APIs to the component author might bring unexpected behavior when component code runs on the server.
```
tedconn: I would also add scalability. The mocked APIs would have to be kept in sync with what's actually being used in the engine. Also the engine still wouldn't be reusable in other environments without a corresponding "mock" layer.

pmdartus: I don't think it is the case. There is only a well-defined set of DOM APIs the engine requires to operate. Each runtime (dom and server) is in charge of implementing this contract by passing an implementation of those APIs to the engine.

text/0112-server-runtime.md
davidturissini: How are elements in ShadowDom going to be represented as a string? Is this only compatible with synthetic shadow?

davidturissini: One thing worth exploring: Can we pass `@api` data for the top level component here? If values are known at "tree creation time", we can probably hydrate a good portion of the LWC tree sync as the initial tree is being created

text/0112-server-runtime.md
davidturissini: I'm assuming that "Serialized String" is a string HTML element/tree. Is that correct?

text/0112-server-runtime.md
tedconn: Maybe it has stabilized by now but when I took that approach I had to add/change multiple mocked APIs during the couple months I touched that code.

text/0112-server-runtime.md
caridy: Yes, it only works with some sort of synthetic shadow... but that's another RFC... as mentioned in this RFC at the beginning. We have been discussing this with some other folks from google and other companies, we might be able to define this in a way that can be rehydrated on the client side.

caridy: this is probably another area to explore in a different RFC since it touches on how much can you render on the server synchronously.

text/0112-server-runtime.md
caridy: I think it will be our own bake implementation... We do not plan to allow user-code to talk to the dom, and if they do, we will throw or ignore. The mocks needed here (btw, I will not call them mocks, but implementation) are those APIs needed for the engine to function, so it goes hand to hand.

caridy: Yes, probably we should formalize this as: `A DOMString containing the HTML serialization of the element's descendants. `

text/0112-server-runtime.md
caridy: I don't think that will be that challenging, we will own both packages, and we can keep them in sync.

text/0112-server-runtime.md
caridy: this item is about users code, in their constructor, and in their hook callbacks they can do anything. I believe this bullet point was about that... maybe I'm wrong...what is this item about?

pmdartus: IMO, the only predictable way to make component render on a non-DOM environment is to restrict all the DOM API access when running in such environment. This would force component authors to write more defensive code if they want their components to be eligible for SSR. The same constraints are also enforced by the rest of the popular framework.

For example, if a component needs to attach an event handler on the `window` object in the `connectedCallback` the code needs to check first if the component is running in a browser.

```js

export default class App extends LightningElement {
  connectedCallback() {
    if (/* check if running in a browser */) {
      window.addEventListener('resize', () => { /* do something */ })
    }
  }
}
```

It's important to note that only a subset of the components written today will be capable to render on the server.

I also commented out the code checking if the component is running in the browser because I don't LWC should provide such mechanism. I would rather delegate this responsibility to the application framework.

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+---
+
+# LWC server runtime
+
+## Summary
+
+The LWC engine has been designed from the beginning to run in an environment with access to the DOM APIs. To accommodate server-side rendering (SSR) requirements, we need a way to decouple the engine from the DOM APIs. The existing `@lwc/engine` will be replaced by 3 new packages: 
+- `@lwc/engine-core` exposes platform-agnostic APIs and will be used by the different runtime packages to share the common logic.
+- `@lwc/engine-dom` exposes LWC APIs available on the browser.
+- `@lwc/engine-server` exposes LWC APIs used for server-side rendering.
```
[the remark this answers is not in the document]
pmdartus: Agreed, I wasn't happy with the naming. 

What do you think about `engine-dom-renderer` and `engine-string-renderer`? Even if the package names are more verbose I don't think it's a problem because the only thing developers have to remember if `lwc`.

text/0112-server-runtime.md
pmdartus: If the data is known before the component is rendered, the current proposal should be sufficient.

```js
import { createElement, renderToString } from "lwc";
import App from "c/app";

const app = createElement("c-app", { is: App });
app.foo = 'Some value';
app.bar = ['Some', 'other', 'value.'];
const str = renderToString(app);

console.log(str);
```

Like a standard `HTMLElement` returned by `createElement` when LWC is running in the browser, the `ServerHTMLElement` return by `created `createElement` on the server would expose:
- the component public properties
- the global HTML properties 
- aria reflection properties

text/0112-server-runtime.md
caridy: right, no controversy there... just I don't understand what "DOM constructors used by component authors" means?

pmdartus: Those are DOM interface constructors that developers might need access even in a non-DOM enabled context. So far, I have found 2 constructors that need to be present on the global object: `CustomEvent` and `Event`. 

I am happy to change the title name for something more explicit.

caridy: Ok, Ok, the important word was "interface", I understand now what you're saying. We are good.

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+There are multiple drawbacks with this approach:
+
+- Performance: The engine only relies on a limited set of well-known APIs, leveraging a full DOM implementation for SSR would greatly reduce the SSR performance.
+- Predictability: By attaching the DOM interfaces on the global object, those APIs are not only exposed to the engine but also to the component author. Exposing such APIs to the component author might bring unexpected behavior when component code runs on the server.
+
+## Adoption strategy
+
+TBD
+
+## How we teach this
```
blittle: This would probably be a major version change requiring new documentation.

text/0112-server-runtime.md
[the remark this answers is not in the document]
apapko: I also like the inverted version: `@lwc/client-engine`, `@lwc/server-engine`, `@lwc/engine`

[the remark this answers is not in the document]
pmdartus: I don't think there will be a renaming impact here. The `@lwc/engine` module should never be referenced directly. Userland code should only reference `lwc`, the only place where we might need to change is on the resolver.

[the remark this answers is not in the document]
caridy: I'm with @ekashida on this, this is NOT about client and server I'm afraid, this is about dom vs dom string. Imagine that in the future we decide to use the dom string output in a client for some reason.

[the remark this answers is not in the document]
pmdartus: > [...] this is about dom vs dom string.

I fully agree with this statement. In this regard, I think it would less confusing if the module that is loaded in a DOM environment contains the term `dom` in its name. I would suggest reaming `@lwc/engine` to `@lwc/engine-dom` in @ekashida proposal.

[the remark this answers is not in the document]
ekashida: > I don't think there will be a renaming impact here. The @lwc/engine module should never be referenced directly.

I was talking about projects that have `@lwc/engine` as a dependency in their `package.json`. I'm assuming that DOM rendering will be the main use case for users so keeping that implicit in the package name seems like it would be less confusing.

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+- `isNodeFromTemplate`
+
+This package injects the native DOM APIs into the `@lwc/engine-core` rendering engine.
+
+### `@lwc/engine-server`
+
+This package exposes the following APIs:
+
+- `createElement(name: string, options: { is: typeof LightningElement }): ServerHTMLElement`: This method creates a new LWC component tree. It follows the same signature as the `createElement` API from `@lwc/engine-dom`. Instead of returning a native `HTMLElement`, this method returns a `ServerHTMLElement` with the public properties, aria reflected properties and HTML global attributed.
+- `renderToString(element: ServerHTMLElement): string`: This method synchronously serializes an LWC component tree to HTML. It accepts a single parameter, a `ServerHTMLElement` returned by `createElement` and returns the serialized string.
```
priandsf: How are the components stylesheets passed as part of the string? Are they pre-pended to the string as a <style> tag?

text/0112-server-runtime.md
```diff
@@ -0,0 +1,331 @@
...
+
+This package injects mock DOM APIs in the `@lwc/engine-core` rendering engine. Those DOM APIs produce a lightweight DOM structure. This structure can then be serialized into a string containing the HTML serialization of the element's descendants. As described in the Appendix, this package is also in charge of attaching on the global object a mock `CustomEvent`.
```
priandsf: The POC directly serializes the VDOM, without the need of a DOM structure. Why is a lightweight DOM necessary? Isn't that an extra generation step that would impact the performance?

text/0112-server-runtime.md
pmdartus: This is one area that will require more investigation. The serialization format is outside the scope of this proposal and will be subject to another RFC: https://github.com/salesforce/lwc-rfcs/blob/ccbd9f38a7ff30d2bd604107e6eb13608366aed2/text/0112-server-runtime.md#scope

text/0112-server-runtime.md
pmdartus: Yes, this DOM structure will have a performance impact however there needs to be a DOM structure (even really lightweight) to attach the event listeners and dispatch the DOM events.

text/0112-server-runtime.md
priandsf: @seckardt Maybe you want to share your investigation in this area?

seckardt: On our first baby steps towards SSR I didn't reinvent the wheel in that area. I simply reused existing functionality as provided by the `StylesheetFactory`.

text/0112-server-runtime.md
```diff
@@ -0,0 +1,329 @@
+---
+title: LWC server runtime
+status: DRAFTED
+created_at: 2020-01-06
+updated_at: 2020-01-06
+pr: https://github.com/salesforce/lwc-rfcs/pull/23
+---
+
+# LWC server runtime
```
tedconn, line 9: @pmdartus "LWC Server Runtime" sounds like an application framework for running LWC, like LWR (and runtime is super overloaded). Can we rename this to "LWC Server Engine" or "LWC DOM Decoupling"

text/0112-server-engine.md
```diff
@@ -0,0 +1,558 @@
...
+
+When using a type-safe language like TypeScript, it is possible to ensure that all the rendering APIs are fulfilled in all environments at compile-time. In addition, injecting this indirection at runtime ensures that component code doesn't have access to actual DOM APIs when running on the server.
+
+From a drawback perspective, introducing this layer of indirection between the core logic and underlying APIs might introduce a performance overhead and increase the overall size of the engine.
+
+To inject the rendering APIs depending on the environment, we would need to create a different entry point per environment. Below you can find some pseudo-code for each entry point. The two entry points use the same `createComponent` abstraction to create an LWC component but pass different renderers specific to the target environment.
+
+**`entry-points/dom.js`:**
```
jasonsilberman, line 113:
```suggestion
**`entry-points/dom.ts`:**
```

text/0112-server-engine.md
```diff
@@ -0,0 +1,558 @@
...
+
+```ts
+import { LightningElement, createComponent } from '@lwc/engine-core';
+
+const domRenderer = {
+    createElement(tagName: string): Element {
```
jasonsilberman, line 119: super tiny: you have this as a file called `dom.js` but it is a typescript file :)

text/0112-server-engine.md
```diff
@@ -0,0 +1,558 @@
...
+    tagName: string,
+    opts: { Ctor: typeof LightningElement },
+): Element {
+    const elm = document.createElement(tagName);
+    createComponent(elm, Ctor, domRenderer);
+    return elm;
+}
+```
+
+**`entry-points/server.js`:**
```
jasonsilberman, line 143:
```suggestion
**`entry-points/server.ts`:**
```

text/0112-server-engine.md
```diff
@@ -0,0 +1,558 @@
...
+import App from 'c/app';
+
+const app = createElement('c-app', { is: App });
+document.body.appendChild(app);
+```
+
+**`server.js`:**
+
+```js
+import { renderComponent } from '@lwc/engine-server'; // Resolves to `@lwc/engine-server`
```
jasonsilberman, line 325:
```suggestion
import { renderComponent } from 'lwc'; // Resolves to `@lwc/engine-server`
```
You mentioned in your presentation this was a typo, it's meant to be `lwc` automatically resolves to `@lwc/engine-server`.

text/0112-server-engine.md
```diff
@@ -0,0 +1,558 @@
...
+
+**The `renderedCallback` lifecycle hook will not execute on the server:** When running in a browser, this hook is the first life cycle hook which gives the component author access to the rendered DOM elements. If the component were to attempt to access those APIs on the server it would result in a runtime error since the DOM APIs are not mocked.
+
+**Wire adapters will not be invoked:** The Wire protocol emits a stream of data to a component. The current protocol doesn't define a way today to indicate that the stream is done emitting new data. Because of this, the first version of SSR will not invoke any wire adapter. The protocol will need to be changed and new primitives will need to be added to LWC to make wire adapters compatible with SSR.
```
gamedevsam, line 289: We should consider adding an option to capture a one time snapshot of the stream. Disabling wire adapters during SSR  eliminates the ability to use wire protocol to inject initial state including localized text in the SSR output.

gamedevsam (changes requested):
I'm hoping someone will take a look at the complexity of implementing a one time snapshot of wired data during SSR. If SSR comes out with wire adapters disabled it will significantly decrease its appeal.

At least expose an option? Or a plugin mechanism to enable it to work?

text/0112-server-engine.md
pmdartus: As of today, the invocation timing of the `update` callback is different if the wire adapter [has dynamic parameters or not](https://github.com/salesforce/lwc/blob/e1d7efa3e3d20881f8f11caba623492cd6e6addb/packages/%40lwc/engine-core/src/framework/wiring.ts#L336-L340). This difference makes it impossible to invoke wire adapters in a reliable fashion.

Changing the current invocation timing is an observable change that might result in breaking existing customers on the platform. We decided to set wire adapters aside from the initial SSR implementation but it is something that we will reconsider in the future.

AllanOricil: Did you thought about creating a new hook that fetches data on the server, before serving the component? Nuxt did that, it is called `fetch`. I believe it would be something really great for lwc. With a `fetch` method, components could be served straight away with the data developers need from Salesforce. Pages would be rendered so much faster, not just because the component was rendered on the server, but also because components would no longer need to wait for the wiredMethod to be added to the stack call on the client, which can be slow depending on the specs of the computer, and how many other wiredMethods were added to the stack.

AllanOricil:
@pmdartus 
Did you thought about creating a new hook that fetches data on the server, before serving the component? Nuxt does that pretty well. In nuxt we have a method called `fetch`. I believe it would be something really great for lwc. With a `fetch` method, components could be served straight away with the data developers need from Salesforce. Pages would be rendered so much faster, not just because the component was rendered on the server, but also because components would no longer need to wait for the wired method to be added to the stack call on the client, which can be slow depending on the specs of the computer, and how many other wired methods were added to the stack. 

I also believe that with a fetch method, lots of use cases for the wired methods would no longer exist.

Would we also be able to configure wich LWC component can be rendered on the server? Or would it be all or none?

text/0112-server-engine.md
gamedevsam: Perhaps this can enabled as an optional configuration param when we call `renderLwcToString` or something like that. By default wire adapters can be skipped, but users can opt-in if their use case is deterministic like in my case, I don't connect wire adapters to network requests, they are simply connected to redux style stores to decouple state and UI.

AllanOricil:
@pmdartus 


Did you thought about creating a new hook that fetches data on the server, before serving the component? Nuxt does that pretty well. In nuxt we have a method called `fetch`. I believe it would be something really great for lwc. With a `fetch` method, components could be served straight away with the data developers need from Salesforce. Pages would be rendered so much faster, not just because the component was rendered on the server, but also because components would no longer need to wait for the wired method to be added to the stack call on the client, which can be slow depending on the specs of the computer, and how many other wired methods were added to the stack. I also believe that with a fetch method, lots of use cases for the wired methods would no longer exist.

Middlewares that can run on both client/server could also be useful to control the initial state of a component depending on the User permissions.

gamedevsam:
@AllanOricil the fetch method is an interesting approach, NextJS offers something called [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props) that does something similar.

My issue with this proposal, or any other proposal that provides an alternative to wire adapters is that it basically requires developers who leverage wire adapters for their data fetching to re-engineer their components to be server side rendered. This may not be avoidable due to some foundational architecture decision, but it's a real shame.

The wire adapter API is one of the best things about LWC, forking the ecosystem by saying on clients wire adapters are supported, but on servers they are not, is going to be a serious barrier for adopting server side rendering.

Perhaps a reasonable compromise would be to support Wire adapters without dynamic parameters during SSR, which would cover several use-cases (including my own).

AllanOricil:
> @AllanOricil the fetch method is an interesting approach, NextJS offers something called [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props) that does something similar.
> 
> My issue with this proposal, or any other proposal that provides an alternative to wire adapters is that it basically requires developers who leverage wire adapters for their data fetching to re-engineer their components to be server side rendered. This may not be avoidable due to some foundational architecture decision, but it's a real shame.
> 
> The wire adapter API is one of the best things about LWC, forking the ecosystem by saying on clients wire adapters are supported, but on servers they are not, is going to be a serious barrier for adopting server side rendering.
> 
> Perhaps a reasonable compromise would be to support Wire adapters without dynamic parameters during SSR, which would cover several use-cases (including my own).

@gamedevsam 
Can't ssr be optional, and maybe configurable per component, and not just routes/pages. If so, maybe it would not be necessary to rewrite anything that is already working. Wire adapters eval could happen after the page is sent to the client. Why exactly do you think people would have to re-engineer their components?

More about nuxt fetch

````
The fetch method, if set, is called every time before loading the component (only for page components). It will be called server-side once (on the first request to the Nuxt app) and client-side when navigating to further routes.
````

If you think a `fetch` method is a good idea, I would like to add an addition to its proposal. `fetch` must not be a Salesforce org feature, like wire methods are. It must run everywhere. But when running on Salesforce, it must be injected with a Salesforce context object.
For me, wire methods always felt weird because they can only be used inside a Salesforce org. I think they shouldn't even be considered a lwc feature.

**Comment by gamedevsam** [comment]
We should maybe take this discussion elsewhere (maybe Slack? Hit me up @Samuel Batista) since it's getting a bit unrelated to the current topic at hand, but I'd like to answer a few misconceptions in your previous comment:

> Wire adapters eval could happen after the page is sent to the client. Why exactly do you think people would have to re-engineer their components?

Wire adapters can result in an invocation of a wire method, as well as wire values directly into member variables. If a wire adapter invokes a method on a component, that method often assigns values into member variables of the component. These variables can (and often are) accessed in the template of the component. If we render the template without invoking wire adapters, the HTML will be rendered using default values provided for those member variables. This may be acceptable for some use cases, but in my case it will break the template rendering, since we assume wire adapters will be invoked before the initial render, and will provide the initial values for member variables. Here's an example of code that would break in several of our repos, where we leverage wire adapters to inject localized text, ex:

```ts
export default class I18nComponentExample {
    @wire(WireI18n) i18n;
}
```

```html
<p>{i18n.keyToLocalizedText}</p>
```

> For me, wire methods always felt weird because they can only be used inside a Salesforce org. I think they shouldn't even be considered a lwc feature.

I use wire methods / wire adapter API outside of Salesforce, we use it to wire data from a Redux-like data stores in applications hosted on Heroku. Wire adapters are frequently used to pipe networked data into components, and that is the main use case for components rendered within Salesforce. But that isn't their only use case. AppExchange uses them to provide reactivity / connect them to external stores that manage our state. Wire adapters for me are like React hooks, they are in my opinion one of the best features of LWC OSS.

text/0112-server-engine.md
gamedevsam: Perhaps a reasonable compromise would be to support Wire adapters without dynamic parameters during SSR, which would cover several use-cases (including my own).

AllanOricil:
> We should maybe take this discussion elsewhere (maybe Slack? Hit me up `@Samuel Batista`) since it's getting a bit unrelated to the current topic at hand, but I'd like to answer a few misconceptions in your previous comment:
> 
> > Wire adapters eval could happen after the page is sent to the client. Why exactly do you think people would have to re-engineer their components?
> 
> Wire adapters can result in an invocation of a wire method, as well as wire values directly into member variables. If a wire adapter invokes a method on a component, that method often assigns values into member variables of the component. These variables can (and often are) accessed in the template of the component. If we render the template without invoking wire adapters, the HTML will be rendered using default values provided for those member variables. This may be acceptable for some use cases, but in my case it will break the template rendering, since we assume wire adapters will be invoked before the initial render, and will provide the initial values for member variables. Here's an example of code that would break in several of our repos, where we leverage wire adapters to inject localized text, ex:
> 
> ```ts
> export default class I18nComponentExample {
>     // this:
>     @wire(WireI18n) i18n;
> 
>     // is equivalent to this:
>     i18n;
>     @wire(WireI18n)
>     i18nConnected(i18n) {
>         this.i18n = i18n;
>     }
> }
> ```
> Rendering the template without the wire adapter will break this component:
> ```html
> <p>{i18n.keyToLocalizedText}</p>
> ```
> 
> > For me, wire methods always felt weird because they can only be used inside a Salesforce org. I think they shouldn't even be considered a lwc feature.
> 
> I use wire methods / wire adapter API outside of Salesforce, we use it to wire data from Redux-like data stores in applications hosted on Heroku. Wire adapters are frequently used to pipe networked data into components, and that is the main use case for components rendered within Salesforce. But that isn't their only use case. AppExchange uses them to provide reactivity / connect them to external stores that manage our state. Wire adapters for me are like React hooks, they are in my opinion one of the best features of LWC OSS.

@gamedevsam my bad

https://lwc.dev/guide/wire_adapter#wire-adapters

After reading this doc, now I understand how it actually works.

text/0112-server-engine.md
pmdartus: IMO, we should make wire execution predictable (both on the client and on the server). It would be surprising that for a developer to discover that a wire stops getting executed in SSR after adding a dynamic param. 

Over the last 6 months, the LWC team has been investigating how we could roll out potentially risky changes like this one (changing the invocation timing of wires) without breaking existing customers in production. I hope that we could leverage one of the approaches we are discussing to solve this issue and enable wire adapter execution for SSR. 

/cc @nolanlawson

pmdartus:
There are a lot of interesting points that have been raised on this thread. Let me try to answer some of them.

> How would ssr work on a lightning page? Would salesforce let us control which lwcs are rendered on the server, or would it be all or none?

For the record, there is currently no plan to use SSR in the context of Lightning Experience. Salesforce is considering using SSR for other experiences where initial page load time and SEO are important. 

As a side note, the portability of existing LWC components on the Salesforce platform from client-side rendering only to client + server-side rendering is also something that we discussed at length. While we understand that customers have put a lot of effort into building LWC components on top of the platform, it's important to acknowledge that building client-side-only UI components is vastly different than build isomorphic components. And this, regardless of the UI framework. 

We are in the middle of porting some of our existing internal LWC components to be SSR compatible, and as part of this journey, some components had to be redesigned from the ground up. Documentation and linting rules are in preparation to help with this migration.

> Did you thought about creating a new hook that fetches data on the server, before serving the component? 

Yes, it was top of mind as we designed SSR for LWC. We are currently investigating an approach similar to Next [`getServerSideProps`](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props) and Nuxt [`asyncData`](https://nuxtjs.org/docs/features/data-fetching/#async-data) for data retrieval at the page level. 

We also discussed exposing the capability to retrieve async data in arbitrary components (like [`fetch`](https://nuxtjs.org/docs/components-glossary/fetch/#nuxt--212) in Nuxt), but it's something that we have set aside for now as it would require a large rewrite of the LWC engine to make all the internals async. We believe the page-level data retrieval would be sufficient for the initial version of SSR. And as always it's something we might reconsider in the future based on developers' feedback.

/cc @nolanlawson @divmain

PR #24, opened 2020-01-16
RFC: Partial Template

Templarian:
Rough draft for Partial Template. I'll expand on this as I have time.

----

[Rendered](https://github.com/Templarian/lwc-rfcs/blob/master/text/0000-partial-template.md)


text/0000-partial-template.md
```diff
@@ -0,0 +1,101 @@
...
+```html
+<template>
+  Parent
+  {partialTemplate}
+</template>
+```
+
+## Motivation
+
+In investigating performance issues in components that are used hundreds of times on a
```
caridy: I don't think we use perf as one of the motivations because there are multiple ways to improve the performance of this. cc @pmdartus

text/0000-partial-template.md
```diff
@@ -0,0 +1,101 @@
...
+page it was noted that creating reusable child components; while organizationally nice,
+were not as performant as inlining the HTML into the template;
+
+In other instances it was noted that large conditonal blocks of HTML could be more organized
```
caridy: this is probably the driving motivation

text/0000-partial-template.md
```diff
@@ -0,0 +1,101 @@
...
+This simplifies things, but still may lead to complications in optimizations [unfamiliar with
+what this entails, so may need LWC context here]. Example:
+
+```
+get dynamicTemplate() {
+    return this.conditonal ? template1 : template2;
+}
+```
+
+## Drawbacks
```
caridy: one big drawback here is the static analysis... as today, analyzing the template can give us a lot of information, but with something like this, it makes it harder

text/0000-partial-template.md
```diff
@@ -0,0 +1,101 @@
...
+## Motivation
+
+In investigating performance issues in components that are used hundreds of times on a
+page it was noted that creating reusable child components; while organizationally nice,
+were not as performant as inlining the HTML into the template;
+
+In other instances it was noted that large conditonal blocks of HTML could be more organized
+into seperate template files.
+
+## Detailed design
```
caridy: The biggest question for me here is what is the scope of the template? can it be controlled? or we assume that the template have access to the component instance? To be clear, I'm talking about `{x}` inside the partial, what does it do? is that equivalent to `{x}` in the importer?

Also, I think the fact that you need JS to use this makes me doubt that this is the correct approach.

text/0000-partial-template.md
Templarian: @caridy I'll rewrite this to explain it in another way. This quip/demo outlines how I was researching ideas: https://salesforce.quip.com/9KGnAAWaHRNr

caridy: this is a public repo @Templarian, don't share internal links.

text/0000-partial-template.md
Templarian: Would 100% assume it would work the same as if it the partial's content had been pasted in.

```html
template = 
<template>Hello {name}!</template>

class { template = importedTemplate; name = 'world'; }
<template>{template}</template>

Would result in:
Hello world!
```

Same would be the case anywhere the `{template}` is placed picking up any scoped variables.

Need to look into how templates are processed now to better understand how this could work. The alternative being templates are compile time placed in, but that limits what they can do and would be mainly for splitting up templates code.

caridy: well, this is faulty for two main reasons:

1. it is not a reusable partial because you don't have a way to specify what should be accessible to it.
2. it is impossible for the compiler (at its current state) to invoke another function (compiled partial) while providing the same JS scope of the caller.

While 2 is just a technical limitation that we can attempt to figure, I believe 1 is an ergonomic issue that is a deal breaker here, let me put an example:

```html
// foo.html
<template><p>Value: {value}</p></template>

// component.html
<template>
      <h1>Before</h1>
      {foo}
      <h2>After</h1>
      {foo}
</template>
```

I'm trying to reuse a piece of the UI so I don't have to duplicate it in my template, but also I don't have to worry about the two fragments to go out of sync. This is clearly a use-case for partials, yet it doesn't work because the partial expects `value` to be accessible, but I have two values in the caller that I will like to display in two different places. That's why I think the scope MUST be specified, e.g.:

```html
// foo.html
<template><p>Value: {value}</p></template>

// component.html
<template>
      <h1>Before</h1>
      <template lwc:partial="./foo" value={myBeforeValue}></template>
      <h2>After</h1>
      <template lwc:partial="./foo" value={myAfterValue}></template>
</template>
```

It also provides some declarative only syntax where the value of the `lwc:partial` directive doesn't have to be an imported value in JS if the partial is relative.

text/0000-partial-template.md
Templarian: Agreed with that reasoning. Resolving this as the main motivation is organization/reuse of templates. Perf is a "byproduct".

text/0000-partial-template.md
Templarian: Updated the alternative section with the example above.

That greatly simplifies the usage and allows more flexibility with template reuse. Assuming one could pair that with `if:true`/`if:false`.

Would there be a huge jump in complexity between that and...
```html
<template lwc:partial={template} value={val}></template>
```

text/0000-partial-template.md
```diff
@@ -72,10 +68,24 @@ Why should we *not* do this? Please consider:
 
 ## Alternatives
 
-The other design choice would be to compile time import via an include syntax.
+Partials imported via the `template` tag and `lwc:partial` attribute.
+
+- Introduce `lwc:partial="./file.html"` for compiled time content.
+- Scoped `{variables}` within templates passed via attributes.
+- Question: Would `lwc:partial={templateUrl}` not work? What about `lwc:partial={template}` via
```
caridy: it should work fine I believe, but @diervo will have to look at this.

text/0000-partial-template.md
```diff
@@ -72,10 +68,24 @@ Why should we *not* do this? Please consider:
...
+
+- Introduce `lwc:partial="./file.html"` for compiled time content.
+- Scoped `{variables}` within templates passed via attributes.
+- Question: Would `lwc:partial={templateUrl}` not work? What about `lwc:partial={template}` via
+  an `import`.
 
 ```
-<template lwc:include="./template.html"></template>
+// foo.html
+<template><p>Value: {value}</p></template>
```
caridy: there are more details missing here, what you can pass into a template tag are attributes, what you can use in `{something}` are attributes, basically:

```
<template lwc:partial="./foo" some-value={myBeforeValue}></template>
```

In which case, you will use:

```
<template><p>Value: {someValue}</p></template>
```

text/0000-partial-template.md
```diff
@@ -0,0 +1,113 @@
...
+
+For any developer coming from JSX environment this pattern is already normal practice. Showing a
+simple example is enough to explain how partials work.
+
+Currently a developer would assume this syntax would include a partial and not `.toString()` the
+JS template function.
+
+# Unresolved questions
+
+- As with all feature this will depend heavily on performance.
```
pmdartus: I don't think there will be any performance penalty with this proposal, you can remove this unresolved question. Partial templates will just be another function invocation returning an array of virtual DOM nodes that will be inserted into the parent virtual DOM tree. This doesn't have any implications for the diffing algo.

pmdartus (commented):
I would like to better understand the motivation of this proposal. The only framework offering a templating system that also supports partial is Ember and they are currently deprecating its usage: https://github.com/gavinjoyce/rfcs/blob/gj/deprecate-partials/text/0000-template.md. The general feedback is to use components over partials.

I would be really cautious before introducing such API in LWC.

text/0000-partial-template.md
priandsf: Would that work for events as well? Let's assume that my partial has a button and I'd like to trigger an action onclick. How can I assign an action? Is it something like this:
```html
<template lwc:partial="./foo" onbuttonclick={myHandler}></template>

<template>
  <button onclick={onbuttonclick}>Mybutton</button>
<template>
```

Templarian: @priandsf Assuming when it compiles down it has a way to wire up the scope it probably could do the same for events. Not really familiar enough with how the internals work though (just assuming it's a compile time in this case).

(this proposal was put on hiatus, but still something that would be beneficial)

PR #25, opened 2020-01-16
RFC: Focus behavior

ekashida:
[Rendered](https://rfcs.lwc.dev/rfcs/lwc/0000-focus-behavior)

### DIFF
--- a/text/0000-focus-behavior.md
+++ b/text/0000-focus-behavior.md
@@ -1,0 +1,152 @@
+
+title: Focus Behavior
+status: DRAFTED
+created_at: 2020-01-15
+updated_at: 2020-01-15
+pr: (leave this empty until the PR is created)
+---
+
+# Focus Behavior
+
+## Summary
+
+When discussing focus behavior, we need to consider the three ways you can
+focus on elements: click focus, programmatic focus, and sequential focus. In
+general, anything that is click focusable or sequentially focusable is
+programmatically focusable; however, focus behavior for clicks and sequential
+navigation can depend on the user agent and system settings. For example,
+Safari will not apply focus on buttons when clicked, but it will apply focus
+on buttons when the focus method is invoked[^1].
+
+As such, the set of click focusable and sequentially focusable elements is not
+something that we can know a priori. What we do know when observing our
+supported user agent matrix is that, the set of _potentially_ focusable
+elements is the same. This proposal makes the assumption that it is acceptable
+for the framework to introduce shadow semantics via various focus management
+polyfills as long as the set of all potentially focusable elements are
+considered while managing focus.
+
+## Motivation
+
+Generally speaking, it is outside the scope of LWC to normalize different
+behaviors across its supported user agent matrix. In the case of focus
+delegation however, the framework does need to make informed decisions to
+normalize focus behaviors since the alternative to that would be to do nothing
+at all, which would result in a broken user experience--especially in the case
+of sequential focus navigation.
+
+The primary goal when considering any proposal around focus behavior will be to
+minimize differences between its synthetic Shadow DOM and native Shadow DOM
+implementations to facilitate the future transition of components to native
+Shadow DOM.
+
+## Detailed design
+
+For simplification, subsequent mentions of components assume that said
+component is delegating focus, unless otherwise stated.
+
+For simplification, subsequent usage of the term "focusable" will refer to
+whether or not an element is programmatically focusable (i.e., whether or not
+an element can receive focus when its focus method is invoked or it has the
+`autofocus` attribute). To be clear, this means that we will be ignoring the
+distinction between the three classifications of focusability.
+
+```
+focusable = programmatically-focusable = click-focusable ∪ sequentially-focusable
+```
+
+### Focusable elements
+
+The set of focusable elements is currently defined by this [monster selector].
+This has been sufficient for our needs so far but can probably be improved. We
+should go through the exercise of checking it against the results of focusable
+elements listed on [this test page](https://boom-bath.glitch.me/tabindex.html).
+
+### Programmatic focus
+
+In the case of programmatic focus on a custom element, if the custom element is
+a shadow-including ancestor of the currently focused element, then we don't do
+anything and the focused element should remain focused. Otherwise, the
+specification states that the first programmatically focusable element should
+receive focus. Since there is no DOM API that identifies such an element for
+us, we can instead:
+
+1. Invoke the focus method on the first element in our set of focusable
+   elements, where the set of focusable elements is the set of sequentially
+   focusable elements plus elements with a non-negative tabindex content
+   attribute value.
+1. Verify that the element received focus via `activeElement`.
+1. If the element did not recieve focus then keep invoking the focus method on
+   subsequent elements until we verify that focus was transferred.
+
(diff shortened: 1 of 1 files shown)


text/0000-focus-behavior.md
```diff
@@ -0,0 +1,153 @@
...
+a shadow-including ancestor of the currently focused element, then we don't do
+anything and the focused element should remain focused. Otherwise, the
+specification states that the first programmatically focusable element should
+receive focus. Since there is no DOM API that identifies such an element for
+us, we can instead:
+
+1. Invoke the focus method on the first element in our set of focusable
+   elements, where the set of focusable elements is the set of sequentially
+   focusable elements plus elements with a non-negative tabindex content
+   attribute value.
```
JanMiksovsky: Is there a reliable way to always tell if a *custom* element in the shadow is focusable?

* If a custom my-element sets delegatesFocus, it should be sequentially focusable, programmatically focusable, and click focusable.
* Significantly, my-element doesn’t need to set a tabindex — the browser will still know it should look inside my-element’s shadow for focusable elements.
* Now suppose a wrapper-element includes in its shadow an instance of my-element. When the programmatic focus heuristic is looking for focusable elements, can it tell whether my-element is focusable for not? As I read this, it would skip my-element if it didn’t define a tabindex of 0 or -1.
* For open shadow roots, it’s possible to inspect shadowRoot.delegatesFocus. But that’s not available for closed roots.
* LWC could try force the tabindex of my-element to -1, or define some other side-channel to communicate focusability, but what if my-element isn’t LWC — how do we tell whether *those* are focusable?

To be frank, there’s something about the whole area of focus that makes it hard for me to keep all the facts in my brain at once, so I might be overlooking something here.

text/0000-focus-behavior.md
```diff
@@ -0,0 +1,153 @@
...
+`autofocus` attribute).
+
+```
+focusable = programmatically-focusable = click-focusable ∪ sequentially-focusable
+```
+
+### Focusable elements
+
+The set of sequentially focusable elements is currently defined by this
+[monster selector]. This has been sufficient for our needs so far but can
```
JanMiksovsky: Some minor issues with that selector:
* It's missing the focusable native elements `details` and `iframe`
* Also, the query currently looks for any element with `[href]`. This is meant to pick up `a` and `area` elements, but could accidentally pick up `base` and `link` elements, as well as any custom element that decided to expose an href attribute. Recommend tightening the query to `a[href]` and `area[href]`.

text/0000-focus-behavior.md
caridy: Open and close doesn't matter here because this runs first, and can traverse any shadow independently of the mode. @ekashida can talk about the rest.

JanMiksovsky: I'm not sure I understand. LWC can't inspect the closed shadow of an element it doesn't control, correct? Or is our general assumption that every element on a page is LWC?

caridy: No, the idea is that this polyfills provides `Element.p.attachShadow`, and stores the ref to that instance in some internal slotting (weakmap). Therefore, any created shadow in the app, is accessible to this polyfill whether or not it is closed, because we have access to the weakmap to extra the reference and traverse accordingly. Makes sense?

JanMiksovsky: I was sure that the topic of determining whether a custom element is focusable came up at the W3C, but it took me a while to remember where/when. Alice Boxhall at Google has been advocating for a general solution, and the topic was discussed at the last TPAC. The current thinking is that this should be exposed on `elementInternals`: see the rough idea at https://github.com/w3c/webcomponents/issues/762#issuecomment-542537639. There is ongoing bikeshedding around the names, but there's general consensus on that approach.

It does look like the proposal to extend `elementInternals` wouldn't make a component's focusability visible to an algorithm like the one in this RFC — but that seems like it would be useful. I can try to suggest that as an extension of that proposal.

* If that suggestion is adopted, then this RFC's algorithm could directly ask a custom element whether it's focusable.
* If that suggestion is _not_ adopted, then the polyfill could potentially patch `elementInternals`. That would at least let a component indicate its focusability in the same way for both native and the polyfill.

caridy: this is great, btw, we should start compiling the list of things that we should suggest for the agenda for the next F2F so we don't forget about all these little things. @JanMiksovsky can you start some list somewhere?

caridy: Btw, in the past, we had some push back on asking for features that were only useful for polyfilling other related things, but I think in general, there will be value for asking whether or not an element is focusable, I can see how to use this information for other things. Up until now, this information was only available in CSS IIRC.

ekashida: @JanMiksovsky 

> Is there a reliable way to always tell if a custom element in the shadow is focusable?

Yes, today we only need to look at the tabindex content attribute. If that proposal is implemented, then we would not have a way to tell if any custom element is focusable.

> If a custom my-element sets delegatesFocus, it should be sequentially focusable, programmatically focusable, and click focusable.

I believe this is incorrect. Focus delegation is about _not_ setting focus on the host element. A component would only be focusable if it contained a focusable element.

> Significantly, my-element doesn’t need to set a tabindex — the browser will still know it should look inside my-element’s shadow for focusable elements.

Not sure if I understood this one correctly due to the "browser" reference but our implementation queries the DOM for all focusable elements before deciding how to manage focus, so we already consider its subtree without considering a custom element's tabindex. We do, however, subsequently exclude a custom element's subtree from sequential focus if it has a tabindex value of -1.

> Now suppose a wrapper-element includes in its shadow an instance of my-element. When the programmatic focus heuristic is looking for focusable elements, can it tell whether my-element is focusable for not? As I read this, it would skip my-element if it didn’t define a tabindex of 0 or -1.

If `my-element`'s component is delegating focus, then it will ignore `my-element`. It could also be that this section is confusing because I made a mistake--it should read: `the set of focusable elements is the set of sequentially focusable elements plus elements with a negative tabindex content attribute value`.

> To be frank, there’s something about the whole area of focus that makes it hard for me to keep all the facts in my brain at once, so I might be overlooking something here.

Welcome to the club! :)

ekashida: > It does look like the proposal to extend elementInternals wouldn't make a component's focusability visible to an algorithm like the one in this RFC — but that seems like it would be useful. I can try to suggest that as an extension of that proposal.

Yes, that proposal breaks our algorithm because we currently rely on a priori information about elements that are possibly focusable. Given that the `tabIndex` IDL attribute is generally not useful at all, I'd expect that trend to continue for custom elements. We could probably figure something out for LWC on the framework side of things, but component authors would most likely need some kind of API to figure out whether a custom element is focusable or not.

text/0000-focus-behavior.md
ekashida: Looks like the test page covers all your points except for `base`. I'll create an issue specifically for that element so that it doesn't slip through the cracks.

text/0000-focus-behavior.md
```diff
@@ -0,0 +1,155 @@
+---
+title: Focus Behavior
+status: DRAFTED
```
apapko:
```suggestion
status: ACTIVE

```

PR #26, opened 2020-03-11
RFC: Documentation-related Metadata

mysticflute:
https://rfcs.lwc.dev/rfcs/lwc/0000-documentation-related-metadata

### DIFF
--- a/text/0000-documentation-related-metadata.md
+++ b/text/0000-documentation-related-metadata.md
@@ -92,7 +92,6 @@ The following experience values will be documented as recognized by Salesforce:
 * Standalone
 * Snapins
 
-
 ### Parsing
 
 _Please note the following implementation is already in place, but is included here for completion. That said, now is a good time to address any concerns before opening up usage on the platform._
@@ -113,7 +112,7 @@ export interface DocumentationMetadata {
 
 In Aura this information will be placed on the ModuleDefinition as a `Map<String, Object>`.
 
-#### Value Agnostic
+#### Field Agnostic
 
 The platform-compiler will be agnostic to the fields and their meanings. Correct YAML grammar will be enforced, but downstream code such as the Component Library will be responsible for determining which fields are applicable and which values may need further processing, such as splitting the experience field on commas.
 
--- a/text/0000-documentation-related-metadata.md
+++ b/text/0000-documentation-related-metadata.md
@@ -1,0 +1,139 @@
+---
+title: Documentation-related Metadata
+status: DRAFTED
+created_at: 2020-03-10
+updated_at: 2020-03-10
+pr: (leave this empty until the PR is created)
+---
+
+# Documentation-related Metadata
+
+## Summary
+
+Component authors need a way to provide documentation-related metadata— metadata that is only useful for documentation purposes and not runtime behavior. This is information such as component categorization, and it usually takes the format of key-value pairs.
+
+## Basic example
+
+```markdown
+---
+category: Data Entry
+experience: Lightning
+---
+
+descriptive documentation...
+```
+
+## Motivation
+
+As part of our path towards allowing customers to add documentation for their modules, there's a need for customers to be able to add documentation-related metadata that will be useful for tooling and documentation purposes. For example, component categories and experiences are displayed in the Component Library documentation website and used for filtering.
+
+In addition, for Salesforce-authored modules there's a need to put component authors in control of this metadata. Currently a hardcoded list of metadata is maintained for a hardcoded list of components, which is a roadblock for teams creating new components or wishing to update the values.
+
+There needs to be a standard way to specify certain documentation-related key-value pairs that balances component-author ownership with 
+
+## Detailed design
+
+Key-value pairs will be placed as [frontmatter](https://github.com/jonschlinkert/gray-matter#what-does-this-do) (in YAML format) in the module markdown file that lives under __docs__. For example, this could be the content for `input/__docs__/input.md`:
+
+```markdown
+---
+category: Data Entry
+experience: Lightning
+---
+
+descriptive documentation...
+```
+
+### Recognized Fields
+
+The following information can be specified in the frontmatter:
+
+* __category__  `string`
+    The main category for the component. Only one may be specified. Salesforce will maintain a list of recognized categories in public documentation and the VS Code extension for code completion/linting. For applications like the Component Library, values that match the whitelist will be displayed on the component’s reference page. Also see the filters on the homepage of the [Component Library](https://developer.salesforce.com/docs/component-library/overview/components) for other ways this information can be used.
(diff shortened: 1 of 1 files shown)


mysticflute:
@diervo looks like I don't have access to add reviewers to this PR. Can you please be a reviewer and also add others, probably @apapko and @caridy and whoever else, thanks!

diervo:
@mysticflute gave you access

text/0000-documentation-related-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+### Recognized Fields
+
+LWC itself will not restrict the set of fields a component author could add to the frontmatter, however only specific fields will be documented as recognized by the Salesforce platform. Unrecognized fields may be present but they will be ignored.
+
+The following keys will be recognized in the frontmatter:
+
+* __category__  `string` \
+    The main category for the module. Only one may be specified. Salesforce will maintain a list of recognized categories in public documentation and the VS Code extension for code completion/linting. For applications like the Component Library, values that match the whitelist will be displayed on the component’s reference page. Also see the filters on the homepage of the [Component Library](https://developer.salesforce.com/docs/component-library/overview/components) for other ways this information can be used.
```
ravijayaramappa, line 57: Use a reusable link
```suggestion
    The main category for the module. Only one may be specified. Salesforce will maintain a list of recognized categories in public documentation and the VS Code extension for code completion/linting. For applications like the Component Library, values that match the whitelist will be displayed on the component’s reference page. Also see the filters on the homepage of the [Component Library][component-library] for other ways this information can be used.

[component-library]: https://developer.salesforce.com/docs/component-library/overview/components
```

text/0000-documentation-related-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+    html?: string;
+    metadata?: DocumentationMetadata;
+}
+
+export interface DocumentationMetadata {
+    [key: string]: any;
+}
+```
+
+In Aura this information will be placed on the ModuleDefinition as a `Map<String, String>`.
```
ravijayaramappa, line 115: Don't think it is necessary to state this here.
```suggestion
~~In Aura this information will be placed on the ModuleDefinition as a `Map<String, String>`.~~

```

text/0000-documentation-related-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+* __experience__  `string[]` \
+    One or more Salesforce experiences that the module works in. Salesforce will maintain a list of recognized experiences in public documentation and the VS Code extension for code completion/linting. For applications like the Component Library, values that match the whitelist will be displayed on the component’s reference page. Also see the filters on the homepage of the [Component Library](https://developer.salesforce.com/docs/component-library/overview/components) for other ways this information can be used.
+* __isSubComponent__  `boolean` \
+    This denotes that the component is only intended for use inside of another component, for example `lightning-breadcrumb` is only used inside of `lightning-breadcrumbs`. The Component Library may use this information for filtering options, to automatically cross-link pages, or to provide [component grouping](https://gus.lightning.force.com/lightning/r/a07B0000006HrUNIA0/view). Also this information could be used for linting.
+* __deprecated__  `boolean` \
```
pmdartus, line 62: > Is there an existing way deprecation is already specified? 

I am not aware of a standard way to mark a component as deprecated.

> Does it make more sense to use the `@deprecated` tag from JSDoc? 

It makes sense to keep the `@deprecated` notation in JSDoc. Even if a module/component is not deprecated, some of their methods and properties might be.

> Also, should we have another field for the replacement component?

It is not uncommon that a deprecated module doesn't have a 1-1 replacement. Deprecated requires most of the time some free form text to indicate how to migrate, we can ask for deprecated modules to add a banner in the documentation explaining the migration strategy. We can also add another property called `deprecationMessage`  to have a standard way to surface in the component library and the different tools.

text/0000-documentation-related-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+descriptive documentation...
+```
+
+### Recognized Fields
+
+LWC itself will not restrict the set of fields a component author could add to the frontmatter, however only specific fields will be documented as recognized by the Salesforce platform. Unrecognized fields may be present but they will be ignored.
+
+The following keys will be recognized in the frontmatter:
+
+* __category__  `string` \
```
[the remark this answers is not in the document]
mysticflute: It would be up to the consumer to decide which text to use to describe a lack of categorization. None of these fields would be considered required from the compiler standpoint.

text/0000-documentation-related-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+    The main category for the module. Only one may be specified. Salesforce will maintain a list of recognized categories in public documentation and the VS Code extension for code completion/linting. For applications like the Component Library, values that match the whitelist will be displayed on the component’s reference page. Also see the filters on the homepage of the [Component Library](https://developer.salesforce.com/docs/component-library/overview/components) for other ways this information can be used.
+* __experience__  `string[]` \
```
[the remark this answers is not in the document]
mysticflute: yes that's a good point. An 'All' experience is reasonable. The risk is that people will tend to use that without testing their components in the experience, or new experiences may be added that some existing components don't support.

text/0000-documentation-related-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+    One or more Salesforce experiences that the module works in. Salesforce will maintain a list of recognized experiences in public documentation and the VS Code extension for code completion/linting. For applications like the Component Library, values that match the whitelist will be displayed on the component’s reference page. Also see the filters on the homepage of the [Component Library](https://developer.salesforce.com/docs/component-library/overview/components) for other ways this information can be used.
+* __isSubComponent__  `boolean` \
+    This denotes that the component is only intended for use inside of another component, for example `lightning-breadcrumb` is only used inside of `lightning-breadcrumbs`. The Component Library may use this information for filtering options, to automatically cross-link pages, or to provide [component grouping](https://gus.lightning.force.com/lightning/r/a07B0000006HrUNIA0/view). Also this information could be used for linting.
```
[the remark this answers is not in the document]
mysticflute: Yes that's a good idea.

text/0000-documentation-related-metadata.md
mysticflute: I like the idea of `deprecationMessage`, as that can be used easier in tools like the editor LSPs. For more complicated scenarios I agree this could be supplemented with more information in the documentation itself.

When I talked with Alex and Ravi it came up that a deprecation flag might be added to the `.js-meta.xml` file. I'm not sure if it belongs there, but I agree the `@deprecated` tag works better than in the doc metadata should that can be a separate discussion.

PR #27, opened 2020-03-16
RFC: Dynamic import hints

apapko:
This RFC proposes the syntax and metadata format for using hints with dynamic imports.

Relevant information to know before reading this RFC:
1. https://github.com/salesforce/lwc-rfcs/blob/master/text/0110-dynamic-components.md

### DIFF
--- a/text/0000-dynamic-import-hints.md
+++ b/text/0000-dynamic-import-hints.md
@@ -1,0 +1,71 @@
+---
+title: Dynamic Import Hints and Its Metadata
+status: ACTIVE
+
+created_at: 2020-03-16
+updated_at: 2020-03-16
+---
+
+# Hints Authoring Syntax and Its Metadata
+
+## Summary
+This RFC suggests authoring syntax for specifying dynamic import hints and describes its metadata shape.
+
+## Motivation
+The Hints in dynamic components allows component authors to declare the hint context by parameterizing dynamic imports. The values from the hints can be used to optimize dependency resolution and pre-fetching.
+
+## Detailed design
+
+### Hints Syntax
+Hints are expressed as a comment inside a dynamic import expression. 
+The hint syntax is a block string with a *key *as a resource and a *value* as a resolved resource value
+
+```js
+/* <key>=<value> */
+```
+Hint comment ex:
+```js
+/* @salesforce/client/formFactor=Small */
+```
+
+Dynamic Import with Hint ex:
+```js
+export async function example1() {
+    loadedModule = await import("lightning-foo"/* @salesforce/client/formFactor=Small */);
+}
+```
+
+### Reasons
+The hint syntax in a form of a comment was selected for the following reasons:
+* complies with ‘what you write is what you see’ concept - an originally authored code remains untouched
+* eliminates any additional code transformations at compiler time
+* eliminates any additional processing during module resolution at runtime (to strip out or pre-process module name)
+
+### Hint Invariants
+* hint values must be globally available and resolvable on the server side 
+* hint values must be statically verifiable
+* hint values must be finite and enumerable (ex: no Math.random())
+
+### Dynamic Import Metadata
+The proposed metadata format will include a dynamically imported component name and a hint object, which in-itself contains the original query and a hint value. A hint value will be an array of objects with raw hint query, expected hint condition, and hint resource information.
+
+Proposed Shape
+```js
+{
+  dynamicImports: [{
+    moduleName: "lightning-foo",
+    file: "foo.js"
+    hints: [
+                rawValue: "/* @salesforce/client/formFactor=Small */"
+                resource: {
+                    id: "@salesforce/client/formFactor",
+                    namespacedId: undefined, // scoped modules are namespaced at compile time
+                    type: "client",
+                    file: "foo.js",
+                    location: {...}
+                },
+                value: "Small", 
+    ]
+  }]
+}
+```
--- a/text/0000-dynamic-import-hints.md
+++ b/text/0000-dynamic-import-hints.md
@@ -39,7 +39,7 @@ Using hints allows component authors to specify a condition on which a dynamic i
 ## Detailed Design
 
 ### Grammar
-This section defines the grammar to be followed for hints so that the compiler can parse component modules, gather hints metadata and provide that as part of the compilation result.
+This section defines the grammar to be followed for expressing hints. The grammar establishes a contract between the component and the compiler to facilitate hint recognition by the compiler, gathering the metadata and providing that as part of the compilation result.
 
 #### Hint Syntax
 A hint is expressed as a comment inside a dynamic import expression. The hint syntax is either a line comment or a block comment with a *key* and *value* pair. The key and value will be separated by a colon(`:`) and delimited using double quotes(`"`).
--- a/text/0000-dynamic-import-hints.md
+++ b/text/0000-dynamic-import-hints.md
@@ -12,7 +12,12 @@ updated_at: 2020-03-16
 This RFC suggests authoring syntax for specifying dynamic import hints and describes its metadata shape.
 
 ## Motivation
-The Hints in dynamic components allows component authors to declare the hint context by parameterizing dynamic imports. The values from the hints can be used to optimize dependency resolution and pre-fetching.
+Using hints allow component authors to declare the component preload context by parameterizing dynamic imports. The values from the hints can be used to optimize dependency resolution and pre-fetching.
+
+## Hint Invariants
+* hint values must be globally available and resolvable on the server side 
+* hint values must be statically verifiable
+* hint values must be finite and enumerable (ex: no Math.random())
 
 ## Detailed design
 
@@ -28,7 +33,7 @@ Hint comment ex:
 /* @salesforce/client/formFactor=Small */
 ```
 
-Dynamic Import with Hint ex:
+Dynamic import with a hint ex:
 ```js
 export async function example1() {
     loadedModule = await import("lightning-foo"/* @salesforce/client/formFactor=Small */);
@@ -38,16 +43,12 @@ export async function example1() {
 ### Reasons
 The hint syntax in a form of a comment was selected for the following reasons:
 * complies with ‘what you write is what you see’ concept - an originally authored code remains untouched
-* eliminates any additional code transformations at compiler time
+* eliminates any additional code transformations at compile time
 * eliminates any additional processing during module resolution at runtime (to strip out or pre-process module name)
 
-### Hint Invariants
-* hint values must be globally available and resolvable on the server side 
-* hint values must be statically verifiable
-* hint values must be finite and enumerable (ex: no Math.random())
 
 ### Dynamic Import Metadata
-The proposed metadata format will include a dynamically imported component name and a hint object, which in-itself contains the original query and a hint value. A hint value will be an array of objects with raw hint query, expected hint condition, and hint resource information.
+The proposed metadata format will include dynamically imported component name and a hint object, which in-itself contains the original query and a hint value. A hint value will be an array of objects with raw hint query, expected hint condition, and hint resource information.
 
 Proposed Shape
 ```js
@@ -72,8 +73,8 @@ Proposed Shape
 
 ## Other Options Considered
 
-### Hints as query parameters
-The authoring hints syntax is a query paramter to the dynamic import value:
+### Hints as a query attached to the component name
+The hint syntax is a parameterized query appended to the component name:
 ```js
 export async function example1() {
     loadedModule = await import("lightning-foo*?*@salesforce/client/formFactor=Small");
@@ -87,10 +88,10 @@ export async function example1() {
 *Cons:*
 
 * requires transform to remove the query 
-* Using a query string syntax to discover modules to prefetch doesn't align with the way standard import() works. JavaScript VM caches modules by URL (including the query string), this means that lightning-foo lightning-foo?1 and lightning-foo?2 are 3 different JavaScript module instances. 
+* using a query string syntax to discover modules to prefetch doesn't align with the way standard import() works. JavaScript VM caches modules by URL (including the query string), this means that lightning-foo lightning-foo?1 and lightning-foo?2 are 3 different JavaScript module instances. 
 
-### Hints as proprietery function
-The authoring hints syntax is expressed via a proprietery lwc function, where the first parameter is a component name and the second parameter is a query object - hintImport(cmpName, queryObj). At compile time, the function will be transformed into a regular import without a string. Note, the hint is consumed by metadata parser only, which is traversed outside of the compilation sequence on the original source. Therefore, it is safe to have a ‘throw-away’ wrapper that will be eliminated at compiler time.  
+### Hints via proprietary LWC function
+The hint syntax is expressed via a proprietary lwc function, where the first parameter is a component name and the second parameter is a query object - hintImport(cmpName, queryObj). At compile time, the function will be transformed into a regular import without a string. Note, the hint is consumed by metadata parser only, which is traversed outside of the compilation sequence on the original source. Therefore, it is safe to have a ‘throw-away’ wrapper that will be eliminated at compiler time.  
 
 ```js
 import formFactor from '@salesforce/client/formFactor';
--- a/text/0000-dynamic-import-hints.md
+++ b/text/0000-dynamic-import-hints.md
@@ -34,13 +34,12 @@ Using hints allows component authors to specify a condition on which a dynamic i
 
 ## Hint Invariants
 * Hint key and value are string literals
-* Hint key and value must be statically verifiable i.e. no curly braces to implicitly refer to a variable and such
 * Hints do not influence the outcome of the compilation phase. It only produces additional metadata for the application framework to consume.
 
 ## Detailed Design
 
 ### Grammar
-There will be grammar restrictions in place for hints so that the compiler can parse component modules, gather metadata and provide that as part of the compilation result.
+This section defines the grammar to be followed for hints so that the compiler can parse component modules, gather hints metadata and provide that as part of the compilation result.
 
 #### Hint Syntax
 A hint is expressed as a comment inside a dynamic import expression. The hint syntax is either a line comment or a block comment with a *key* and *value* pair. The key and value will be separated by a colon(`:`) and delimited using double quotes(`"`).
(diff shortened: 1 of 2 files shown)


text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,128 @@
...
+```
+Hint comment ex:
+```js
+/* @salesforce/client/formFactor=Small */
+```
+
+Dynamic import with a hint ex:
+```js
+export async function example1() {
+    loadedModule = await import("lightning-foo"/* @salesforce/client/formFactor=Small */);
```
pmdartus: There cases this RFC doesn't cover:
- What happen if the `import` first parameter is not a string literal but rather an identifier?
- Should the compilation fails if the comment doesn't respect the grammar?
- The proposed syntax only covers values being parsed a string? Do we want to allow in the future boolean, number or object values?
- Do we want to support multiple hints combinations? (AND / OR operators)
- What happen if the `key` or the `value` contains a string?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,128 @@
...
+created_at: 2020-03-16
+updated_at: 2020-03-16
+---
+
+# Hints Authoring Syntax and Its Metadata
+
+## Summary
+This RFC suggests authoring syntax for specifying dynamic import hints and describes its metadata shape.
+
+## Motivation
```
tedconn: I personally would like to see more explanation as to how the dynamic loading lifecycle in a typical application is improved with the hint. How would it typically be used and what gains could one expect?

diervo: Very fair point.

text/0000-dynamic-import-hints.md
ravijayaramappa: From the team discussion:
1. If the first parameter is not a string literal, no metadata will be gathered. Neither for the dynamic import statement nor for any hint specified.
2. Compilation does not fail if the comment does not respect grammar. The hint will be ignored. An ESLint rule will be added as part of future work to flag incorrect hints.
3. The new syntax will be `"<hint key>": "<hint value>"` to disambiguate and accommodate string literals, boolean and number. 
   i. Object values will be considered for a future improvement.
4. Since the value part of the hint will be delimited with double quotes, it is left to the discretion of the component author to use operators.

In addition to these, the hint grammar will be restricted to use the following characters
1. hint key: `^"[a-zA-Z@][a-zA-Z0-9_-/]+"$`
  - Starts with an alphabet or @, followed by any number of alphanumeric characters, underscore, dash and forward slash
  - Begins and ends with a double quote 

2. delimiter: `[:]\s+`

3. hint value`: `^"[a-zA-Z0-9_-/@{}:]+"$`
 - Any alphanumeric character, underscore, dash, forward slash, @, curly braces and colon. We will not validate the hint value for its completeness or if its a valid json string etc.
 - Beings and ends with a double quote

pmdartus: > This should probably be defined and standardized if we think it's useful.

IMO, the definition of the available hint key and values should be part of this RFC. On the implicit invariant of this RFC is that hints don't have any impact on the compilation output. Due to this, the RFC focussed on the grammar with hints so the compiler could parse and returned the hints as part of the metadata. How those hints are validated, post-processed and how they might impact the way modules are loaded is an application container concern.

But yes I agree that those hints should be standardized somewhere. Mobile team might give a first stab at it since they are the first consumer of this new feature. /cc @cseymourSF

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,130 @@
...
+Using hints allow component authors to declare the component preload context by parameterizing dynamic imports. The values from the hints can be used to optimize dependency resolution and pre-fetching.
+
+## Hint Invariants
+* hint values must be globally available and resolvable on the server side 
+* hint values must be statically verifiable
+* hint values must be finite and enumerable (ex: no Math.random())
```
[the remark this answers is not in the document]
pmdartus: As defined by the grammar in the [above comment](https://github.com/salesforce/lwc-rfcs/pull/27#discussion_r394750299), hints support only string, but we might be able to add support in the future for other primitives (eg. integer, boolean) if needed.

ravijayaramappa:
@pmdartus @jbenallen this is ready for another pass.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+```js
+/* "@salesforce/client/formFactor": "SMALL" */
+/* "MY_CONDITION": "VALID" */
+// "MY_OTHER_CONDITION": "NOT_VALID"
+```
+
+Dynamic import with a hint e.g.:
+```js
+export async function example1() {
+    loadedModule = await import("lightning-foo"/* "@salesforce/client/formFactor": "SMALL" */);
```
tedconn: Would this example be better if it showed an `if` statement wrapping the import as well as the import of `@salesforce/client/formFactor`? To make it clear that the component author is variably importing something and the app framework will react to that?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+```js
+{
+    dynamicImports: [{
+        moduleName: 'lightning-foo',
+        file: 'foo.js'
+        hints: [{
+            rawValue: '/* "@salesforce/client/formFactor": "SMALL" */'
+            specifier: {
+                id: '@salesforce/client/formFactor',
+                namespacedId: undefined, // scoped modules are namespaced at compile time
```
tedconn: The hint key is not always a module, right? It could be anything? The compiler metadata probably shouldn't dictate your hint's meaning?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+Using hints allow component authors to declare the component preload context by parameterizing dynamic imports. The values from the hints can be used to optimize dependency resolution and pre-fetching.
+
+## Hint Invariants
+* Hint key and value are string literals
+* Hint keys and values must be statically verifiable i.e. no curly braces to implicitly refer to a variable and such
+* Multiple hints will be allowed but with restrictions
+
+## Platform Specific Hint Invariants
+For the salesforce platform, the following invariants are proposed
+* Hint values must be globally available and resolvable on the server side
```
tedconn: How will this be enforced?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+implementation: https://git.soma.salesforce.com/lwc/lwc-api/pull/42 & https://git.soma.salesforce.com/lwc/lwc-platform/pull/361
+---
+
+# Hints Syntax and Its Metadata
+
+## Summary
+This RFC suggests authoring syntax for specifying dynamic import hints and describes its metadata shape.
+
+## Motivation
+Using hints allow component authors to declare the component preload context by parameterizing dynamic imports. The values from the hints can be used to optimize dependency resolution and pre-fetching.
```
tedconn: > Using hints allow component authors to declare the component preload context by parameterizing dynamic imports.

This sentence feels a little cryptic. What about something along the lines of:

"Using hints allows component authors to specify a condition on which a dynamic import will be necessary. This is a way to describe the conditions required for this import to happen. An application framework can use these hints to optimize the loading of those resources (for example: pre-fetch at the time of resolving dependencies)"

As I said below an example showing an if statement might be a little more clear.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Static Semantics
+* Both key and value have to be string literal
+* Is case sensitive
+* Key and value have to be delimited with a double quote to be explicit about the literal value
+* Spaces before and after the colon(`:`) is optional
+* Hint phrase can have leading and trailing spaces
+
+##### Hint Key:
+* First character is an alphabet, can be uppercase or lower case. Can also start with `@` character
```
pmdartus: Is there any reason for restricting the key charset?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+* Cannot contain any special characters except for underscore(`_`), hyphen(`-`), dot(`.`) and forward slash(`/`)
+* Begins and ends with a double quote(`"`)
+* Cannot contain spaces, cannot be delimited by a single quote(`'`)
+
+##### Hint value:
+* Can contain alphanumberic characters
```
pmdartus: Is there any reason for restricting the key charset?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+
+// Dynamic module name specifier
+const bar = 'module-name';
+import(bar/* "@salesforce/client/formFactor": "SMALL" */)
+```
+
+#### Separation of Concern
+Dynamic imports hint metadata gathering happens during component compilation. An implicit invariant of this process is that the metadata collection will **not** fail due to a bad hint. This could be a hint not following specified grammar or not using a well understood hint value. Validation of the hint keys and values will not be the concern of this RFC.
+
+Additional validation and post processing of the gathered metadata is not a concern of this RFC. That logic will be a responsibility of the business logic that consumes the metadata.
```
pmdartus:
```suggestion
Additional validation and post processing of the gathered metadata is out of the scope of this RFC. That logic will be a responsibility of the business logic that consumes the metadata.
```

text/0000-dynamic-import-hints.md
pmdartus: I echo this as well, the hints should just be a key/value pair. I don't see the value of restricting keys to valid modules. For example, we might add in the future: 
- a `prefetch` hint that will always prefetch the chunk
- or a `chunkName` like in webpack to force the name of the generated chunk.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+* Having to explicitly define a resource reference will allow for ref integrity support if required
+
+*Cons:*
+
+* Requires an additional transform
+* Requires knowledge of syntax existence
+* Maintanance burden
+
+## Future work
+* Supporting hints for static imports(`import Foo from "lightning-foo"; /* "@salesforce/client/formFactor": "SMALL" */`)
```
pmdartus: Why, static imports are runtime independent.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+* Supporting usage of operators to combine hints and hint values
+* Support for other primitive data types boolean, number
+* Support for JSON string as hint value to represent an object
+
+## Annex
```
pmdartus:
```suggestion
## References
```

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,264 @@
...
+/* "@salesforce/client/formFactor": "SMALL" */
+/* "MY_CONDITION": "VALID" */
+// "MY_OTHER_CONDITION": "NOT_VALID"
+```
+
+Dynamic import with a hint e.g.:
+```js
+export async function example1() {
+    loadedModule = await import("lightning-foo"/* "@salesforce/client/formFactor": "SMALL" */);
+}
```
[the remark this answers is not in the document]
tedconn: Thanks! I didn't have the patience to figure out how to do a multiline suggestion.

text/0000-dynamic-import-hints.md
ravijayaramappa: The intention was to reuse the [Reference](https://git.soma.salesforce.com/lwc/lwc-api/blob/master/src/main/java/com/salesforce/lwc/reference/Reference.java#L16) data structure. I am fine to open this up to a more generic definition like
```
 interface DynamicImportHint {
    rawValue: string;
    hintKey: string;
    hintValue: string;
    location: SourceLocation;
}  
```
_SourceLocation: From @babel/types_

@apapko fyi

text/0000-dynamic-import-hints.md
ravijayaramappa: This basic restriction was to keep the hints simple and readable. We can start out with alphanumeric characters and some limited special characters. We can loosen the restriction later if there is need.

text/0000-dynamic-import-hints.md
ravijayaramappa: It could be as a way to override a backend configuration and force to load a specific variation of a module.
For example: Consider there is an application framework which uses AMD format and injects dependencies. Based on a runtime condition, it could be injecting a mobile version of the dependencies. If the client chooses to override that for a specific case, it can be done with a hint to force the desktop version of the dependency. 
Something along those lines. There is no such usecase yet, but was leaving that option open.

I could update the language to say `Support for hints in static imports is not in consideration at this point.`

text/0000-dynamic-import-hints.md
ravijayaramappa: Will remove it. We can find another place to capture the salesforce specific parts of this RFC.

ravijayaramappa:
Open Questions I have at this point:
- When multiple hints are provided, what is the default operator on the hints. Is it an `AND` or `OR`?
- We could restrict to single hints for now until we are ready to support operators.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,246 @@
...
+* Cannot combine multiple values using operators(i.e `"KEY": "LARGE"&"SMALL"` or `"KEY": "LARGE && SMALL"` will not qualify as a valid hint)
+
+##### Multiple Hints:
+* Each hint must be in a separate block comment or line comment
+* Multiple hints cannot be combined in a single comment using operator
+* Each hint will be specified in a separate comment block
+* They cannot be combined using operators like AND or &&.
+
+#### Semantics
+The [_ModuleSpecifier_](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-imports) of the import statement should be a string literal. The compiler cannot analyze data about a module name that will be discovered or learned at runtime. For example, in the following case no hint metadata will be gathered:
```
tedconn: Isn't this already defined here? Isn't that out of scope for the hints proposal?

https://github.com/salesforce/lwc-rfcs/blob/8e33931ddec7f55216489ec636a4ad98b77a5c11/text/0110-dynamic-components.md#L296

text/0000-dynamic-import-hints.md
jbenallen: Something like [import maps](https://github.com/salesforce/lws-rfcs/pull/16) may be a better tool for that, particularly since the hint value doesn't vary across the lifetime of the `document`.

text/0000-dynamic-import-hints.md
ravijayaramappa: The [dynamic imports implementation](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/babel-plugin-component/src/dynamic-imports.js#L55-L68) allows for this rule to be relaxed.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,257 @@
...
+const bar = 'module-name';
+import(bar/* "@salesforce/client/formFactor": "SMALL" */)
+```
+
+#### Separation of Concern
+Dynamic imports hint metadata gathering happens during component compilation. An invariant of this process is that the metadata collection will **not** fail due to a bad hint. This could be due to a hint not following specified grammar or not using a well understood hint value.
+
+Additional validation and post processing of the gathered metadata is out of the scope of this RFC. That logic will be the responsibility of the application framework that consumes the metadata.
+
+Similarly, any design time validations of the hint grammar should be handled by linting code. For example, on salesforce platform, allowed hint keys might be restricted to only `@salesforce` scoped specifiers. The hint values will be validated based on the relevant value for the specifier. An ESLint rule will enforce this restriction via the [eslint-plugin-lwc](https://github.com/salesforce/eslint-plugin-lwc).
```
pmdartus: > Similarly, any design time validations of the hint grammar should be handled by linting code.

I don't think the term design time makes much sense here. Validation is done by the consumer of the compiled output. Linting might also be used as a way to validate import hints without having to compile a module.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,257 @@
...
+export interface DynamicImportHint {
+    rawValue: string; // Trimmed hint phrase
+    hintKey: string;
+    hintValue: string;
+    location: SourceLocation;
+}
+
+export interface DynamicImportMetadata {
+    moduleName: string;
+    file: string;
```
pmdartus: Why do we need a back pointer to the file here? The consumer of the `DynamicImportMetadata` should already be grouped by file.

ravijayaramappa: Metadata is gathered per BundleConfig. A BundleConfig can have more than one file. Currently, the metadata is not grouped by file. The relevant code is in getMetadata() routine of collector.ts
I will keep this in mind for our metadata factor project.

ravijayaramappa:
@diervo had expressed opinion about restricting grammar. 
Deigo, the discussion here is trending towards handling all of the validation(allowed characters, allowed hint values etc) using lint rules. This RFC will specify the syntax format and restrict the key/value pair to string literals. No additional grammar rules. Let us know your opinion.

text/0000-dynamic-import-hints.md
cseymourSF: doesn't forcing use of an alphabetical first character prevent the use of chinese or other non-latin languages in naming?

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,257 @@
...
+* Begins and ends with a double quote(`"`)
+* Cannot contain spaces, cannot be delimited by a single quote(`'`)
+
+##### Hint Value:
+* Can contain alphanumeric characters
+* Can contain space, underscore(`_`), hyphen(`-`), at sign(`@`), forward slash(`/`)
+* Cannot combine multiple values using operators(i.e `"KEY": "LARGE"&"SMALL"` or `"KEY": "LARGE && SMALL"` will not qualify as a valid hint)
+
+##### Mutiple Hints:
+When multiple hint comments are specifed for a single dynamic import statement, only the fist hint is considered.
```
cseymourSF: fist -> first

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+```
+
+#### Static Semantics
+* Both key and value have to be string literal
+* Is case sensitive
+* Key and value have to be delimited with a double quote to be explicit about the literal value
+* Spaces before and after the colon(`:`) is optional
+* Hint phrase can have leading and trailing spaces
```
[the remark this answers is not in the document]
ravijayaramappa: The `rawValue` property that represents the hint phrase will be trimmed.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,257 @@
...
+export interface SourceLocation {
+    startLine: number;
+    startColumn: number;
+    endline: number;
+    endColumn: number;
+    file: string;
+}
+
+export interface DynamicImportHint {
+    rawValue: string; // Trimmed hint phrase
```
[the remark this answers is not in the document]
ravijayaramappa: Leading and trailing spaces will be trimmed. Retaining them will be not of much value to the user. Space in the middle of the hint phrase will be preserved.

text/0000-dynamic-import-hints.md
```diff
@@ -125,23 +122,24 @@ import("c-bar-mobile"/* "KEY": "1"*/);
 Invalid
 ```js
 import("c-bar-mobile"/* "KEY SUBKEY": "SMALL"*/);
-import("c-bar-mobile"/* "😀": "1"*/);
 import("c-bar-mobile"/* @salesforce/client/formFactor = SMALL*/);
 
-import("c-bar-mobile" /* "@salesforce/client/formFactor": "SMALL && LARGE"*/);
+// Operators
+import("c-bar-mobile" /* "@salesforce/client/formFactor": "SMALL" && "LARGE"*/);
```
pmdartus: We can remove this operator section, it is more confusing that anything.

text/0000-dynamic-import-hints.md
```diff
@@ -0,0 +1,255 @@
...
+## Hint Invariants
+* Hint key and value are string literals
+* Hints do not influence the outcome of the compilation phase. It only produces additional metadata for the application framework to consume.
+
+## Detailed Design
+
+### Grammar
+This section defines the grammar to be followed for expressing hints. The grammar establishes a contract between the component and the compiler to facilitate hint recognition by the compiler, gathering the metadata and providing that as part of the compilation result.
+
+#### Hint Syntax
```
jbenallen: Consider defining syntax and grammars within RFCs using [ABNF](https://tools.ietf.org/html/rfc5234) since it's the standard IETF language for defining grammar and syntax rules.

ravijayaramappa: @jbenallen Thanks for the suggestion. I will merge this PR and use a follow up PR to update the grammar pieces in the ABNF format. This will isolate the styling changes.

PR #32, opened 2020-05-06
Allow sharing of LWC across namespaces

shahprincea:
Some teams have more than one namespace and only desire to share their common components internally. However, it is not possible today, which forces such teams to make their component available globally. This is an undesired effect as it introduces a maintenance burden.

The proposal here is to allow the sharing of Internal LWC across targeted namespaces.


caridy:
hey @shahprincea, I haven't read the RFC in details yet, but at first glance, there are a couple of things to clarify:

1. ns-to-ns communication is NOT a limitation on LWC, it is a limitation of current locker, a product security requirement that we can't fulfill with the current locker design.
2. we have been working on this problem for quite some time without requiring any change in user's code. If you're interesting in piloting this, you can talk to @seksenov 

I will take a look at the proposal tomorrow, but in principle we should not need to change anything in user's code.

shahprincea:
> hey @shahprincea, I haven't read the RFC in details yet, but at first glance, there are a couple of things to clarify:
> 
> 1. ns-to-ns communication is NOT a limitation on LWC, it is a limitation of current locker, a product security requirement that we can't fulfill with the current locker design.
> 2. we have been working on this problem for quite some time without requiring any change in user's code. If you're interesting in piloting this, you can talk to @seksenov
> 
> I will take a look at the proposal tomorrow, but in principle we should not need to change anything in user's code.

shahprincea:
> hey @shahprincea, I haven't read the RFC in details yet, but at first glance, there are a couple of things to clarify:
> 
> 1. ns-to-ns communication is NOT a limitation on LWC, it is a limitation of current locker, a product security requirement that we can't fulfill with the current locker design.
> 2. we have been working on this problem for quite some time without requiring any change in user's code. If you're interesting in piloting this, you can talk to @seksenov
> 
> I will take a look at the proposal tomorrow, but in principle we should not need to change anything in user's code.

1. I am not sure about the Locker limitation but from a validation point of view, we do restrict access based on isExposed flag in bundle.xml file. 
2. I will reach out to @seksenov to understand more.

0000-template.md
```diff
@@ -1,74 +0,0 @@
----
```
pmdartus, line 1: The original template should be kept for future RFCs.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+## Summary
+
+There is a tight coupling (1:1) between namespace and Lighting Web Components (LWC). Developer can use isExposed flag to expose the component to the world or keep it private within the namespace.
+
+This RFC proposes functionality to allow sharing of LWC across namespaces.
+
+
+## Motivation
+
+Some teams have more than one namespaces and only desire to share their common components internally. However, it is not possible today, which forces such teams to make their component available globally. This is an undesired effect as it introduces maintenance burden.
```
pmdartus, line 20: It's something that I was thinking about recently, why do teams own multiple namespaces? As far as I know, teams use multiple namespaces to have a structural breakdown between the different modules composing a feature or a set of features. This approach is used due to the flat nature of the LWC module structure.

IMHO, this proposal is attempting to solve the symptoms instead of addressing the core problem which is the lack of flexibility in the way modules can be structured into a single namespace. 

/cc @diervo @kevinv11n @caridy

0000-template.md
shahprincea: Committed template file to master.

text/0000-allow-sharing-of-LWC-across-namespaces.md
shahprincea: The specific use case here is to allow ISV to build the shared library (utility packages or common components) which could be shared internally across their customer.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+- the impact on teaching people Lightning Web Components
+- integration of this feature with other existing and planned features
+- cost of migrating existing Lightning Web Components applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+
+## Adoption strategy
+
+* Since we are only focusing on Internal LWC scope should be limited
```
pmdartus, line 217: What do you mean by internal LWC?

text/0000-allow-sharing-of-LWC-across-namespaces.md
kfmartin: @shahprincea We need to clearly articulate the audience for this feature. Here you are talking about ISVs, who are external customers. However you also use the word "internal" a lot, which implies that the feature only applies to SFDC teams, viz., code running in core. Not on platform.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+| MISSING  | MISSING (FALSE) | Open to local name space only                                 |
+
+**Error Cases**
+
+| namespaceScope    | isExposed | Comments                   | Errors             |
+|----------|-----------|----------------------------|--------------------|
+| Global   | FALSE     | Failed during compile time | Compile time error |
+| Targeted | TRUE      | Failed during compile time | Compile time error |
+| Local    | TRUE      | Failed during compile time | Compile time error |
+
```
kfmartin, line 55: IMHO having such a matrix over complicates its. Requires too much thinking, too much room for misunderstanding and unintentional exposure.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+## Drawbacks
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
+- the impact on teaching people Lightning Web Components
+- integration of this feature with other existing and planned features
+- cost of migrating existing Lightning Web Components applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
```
kfmartin, line 213: Alternative 3... Fine grain control at the component level:

<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
    <apiVersion>49.0</apiVersion>
    <isExposed>false</isExposed>
    <exposedTo>
        <namespace>force</namespace>
        <namespace>forcechatter</namespace>
        <namespace>ui</namespace>
    </exposedTo>
</LightningComponentBundle>

With the option of broader control at the namespace level:

<NamespaceBundle xmlns="http://soap.sforce.com/2006/04/metadata">
    <exposedTo>
        <namespace>force</namespace>
        <namespace>forcechatter</namespace>
        <namespace>ui</namespace>
    </exposedTo>
</NamespaceBundle>

Each component decide whether or not it `isExposed`. Each component has the option to declare who it is `exposedTo`. If it's exposed and doesn't limit who it's exposedTo, then it's exposed to all the namespaces specified in the NamespaceBundle. Essentially your component bundle has precedence.

Without fine grain control, you're going to hit the issue of over exposing. For example, I have a new component and I want it to be exposed to namespace "Foo". My only option is to modify the NamespaceBundle to include "Foo". Darn, I also have to expose all other components in this namespace to "Foo".

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+* Complex implementation as we have to set precedence logic in the code.
+
+## High level design
+
+In order to honor the sharing of LWC across namespaces, we will be doing validation at difference stages viz.
+
+#### Compile Time Validation : 
+
+During compile time all we do is compile the component either using RINO (Aura) or NodeJS (LWC) compiler and load it into registry (BundleAwareDefRegistry). Since compiler has no context of namespace there is nothing much we can do here. During compile time we can validate Bundle xml file and error out as listed in the table above. 
+
```
kfmartin, line 96: I would consider LAS the experts, but in my view LWC on core is tied into the Aura framework for compiling. In Aura, compiling also involves linking. Every time you compile a component it's also gets linked. In Aura look at AccessChecker.computeModuleAccess. That's where this enforcement should happen.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+* Update the documentation and communicate internally about isExposed flag. 
+* Since we are only focusing on Internal LWC scope should be limited
+
+## Unresolved questions
+
+* Does compiler brings in direct and in direct dependencies ? Not sure how does sub definition plays in to this.
+* Can we do anything better than in memory map for storing the access check result ? 
+
+
```
kfmartin, line 230: There's a quirky detail about access checking for on core components. If there's no `minApiVersion` specified in your bundle's meta.xml file, then your component is not exposed to external namespaces. It's only applicable to SFDC code running in core. This maybe our opportunity to improve this, by making it more explicit that you intend to be exposed to external namespaces.

0000-template.md
kevinv11n: See the second bullet of https://git.soma.salesforce.com/aura/aura/blob/master/aura-spring/src/main/java/org/auraframework/system/Annotations.java#L74:
> Copy 0000-template.md to text/0000-my-feature.md (where 'my-feature' is descriptive. Don't assign an RFC number yet).

This needs to be fixed immediately because you'll lose comment history.

pmdartus:
Closing this RFC since it doesn't have any implication on the open-source version of LWC. Please refer to the internal quip version of this proposal.

text/0000-allow-sharing-of-LWC-across-namespaces.md
midzelis: For some random prior art on 'sharing internals', see: https://www.vogella.com/tutorials/OSGi/article.html#defining-provisional-api-via-the-x-internal-and-x-friends

text/0000-allow-sharing-of-LWC-across-namespaces.md
shahprincea: thank you good to know.

text/0000-allow-sharing-of-LWC-across-namespaces.md
shahprincea: Yes we are planning to do the linking time access for imports. However, with the component, if users have mis configured say isExposed to true but namespaceScope is local in that case we wanted to throw QFE.

text/0000-allow-sharing-of-LWC-across-namespaces.md
shahprincea: The use case you mention is it valid are our customer doing that now? As per internal discussion, we decided that components within the namespaces are either all exposed to not. We don't have a case of cherry picking here. Please let us know if that is the case.

text/0000-allow-sharing-of-LWC-across-namespaces.md
shahprincea: The only reason for this table was to call out all the cases. This was difficult with the table representations.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+## Motivation
+
+Some teams have more than one namespaces and only desire to share their common components internally. However, it is not possible today, which forces such teams to make their component available globally. This is an undesired effect as it introduces maintenance burden.
+
+## Basic example
+
+**Proposal** - Don't touch IsExposed flag for now
+
+We will introduce new tag, **namespaceScope** in LightingComponentBundle.xml (**-meta.xml) file. It’s a string which can take following value viz. global, targeted or local. This tag can be missing in that case we treat it as MISSING internally.
```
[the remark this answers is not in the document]
shahprincea: I am open for the suggestion however PO wanted to have two word combo to suggested the scope.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+    <apiVersion>49.0</apiVersion>
+    <isExposed>false</isExposed>
+    <namespaceScope>global/targeted/local</namespaceScope> 
+</LightningComponentBundle>
+```
+```
+<!-- NamespaceBundle.xml -->
+<?xml version="1.0" encoding="UTF-8"?>
+<NamespaceBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <namespaces>
```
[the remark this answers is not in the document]
shahprincea: point taken will work with PO for names

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+We will introduce new tag, **namespaceScope** in LightingComponentBundle.xml (**-meta.xml) file. It’s a string which can take following value viz. global, targeted or local. This tag can be missing in that case we treat it as MISSING internally. 
+
+
+**Global** : Acts similar to isExposed to true flag. It opens the LWC to the world.
+**Targeted** : It implies that we are opening this LWC to limited namespaces. 
+**Local** : This LWC is only open to local namespace
+
+We will also introduced new xml file called NamespaceBundle.xml. This file is at namespace level. It will enlist namespaces where all the LWCs can be shared. This will allow sharing for LWCs to targeted namespaces. 
+
+For now we will honor isExposed flag and display error for in valid scenarios listed below. However, in the future we will completely rely on namespaceScope tag.
```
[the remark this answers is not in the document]
shahprincea: Since isExposed is boolean and we have tri states like Global, targed and local plan was to introduce new tag and retired the old one. So for now namespaceScope = global should behave exactly like isExposed=true.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+## Basic example
+
+**Proposal** - Don't touch IsExposed flag for now
+
+We will introduce new tag, **namespaceScope** in LightingComponentBundle.xml (**-meta.xml) file. It’s a string which can take following value viz. global, targeted or local. This tag can be missing in that case we treat it as MISSING internally. 
+
+
+**Global** : Acts similar to isExposed to true flag. It opens the LWC to the world.
+**Targeted** : It implies that we are opening this LWC to limited namespaces. 
+**Local** : This LWC is only open to local namespace
```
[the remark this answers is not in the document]
shahprincea: Plz is to retire isExposed eventually so we are looking for valid state which exposes it locally.

text/0000-allow-sharing-of-LWC-across-namespaces.md
```diff
@@ -0,0 +1,230 @@
...
+
+Algo: 
+loop for all the dependencies for given def
+```
+for given referencing def compare its namespace against allowed/targeted namespaces.
+
+    if (true)
+        continue;
+    else
+        QFE (fail fast)
```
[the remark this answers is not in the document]
shahprincea: QuickFixExceptions

text/0000-allow-sharing-of-LWC-across-namespaces.md
apapko: Copying my commend from the quip rfc:
Internal teams having multiple namespaces is a reality that we live in today, it is not a symptom, it is a diagnosis :). In my opinion, lwc flat structure isn't something that will be changing any time soon, meanwhile developer ergonomics is suffering. With that, this proposal not only allows teams to logically organize their code base, but it also enables fine-grained code re-use across other namespaces. Just yesterday, we've had a team [asking](https://gus.lightning.force.com/lightning/r/0D5B0000016c85yKAA/view) if they can share their library to a specific namespace that falls outside of theirs. 
In my opinion this proposal doesn't conflict with working out the details of LWC structure. However, I do agree that all the use cases must be specified in the proposal for clarity.

PR #33, opened 2020-05-08
feat: RFC for canonical metadata shape of LWC modules

ravijayaramappa:


text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,223 @@
...
+* [Goals](#goals)
+* [Use Cases](#use-cases)
+* [Current State of Art](#current-state-of-art)
+* [Detailed design](#detailed-design)
+* [References](#reference)
+
+# Summary
+> The [dictionary definition](https://www.merriam-webster.com/dictionary/canonical%20form) of **Canonical form** is `the simplest form of something`. 
+
+This RFC aims to document the metadata collected by statically analyzing the source code of Lightning Web Component(LWC) modules. It documents the standardized shape of metadata.
```
tedconn, line 24: Are you aligned with @moses-salesforce on ensuring this is an exact subset of the broader type used by that workgroup to describe a component across any runtime?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Goals
+
+- Completeness
+- Eliminate redundancy: There should be only one way to get a desired information about a module.
+- Consitency in property names
```
tedconn, line 30:
```suggestion
- Consistency in property names
```

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,223 @@
...
+- Flexibility: Allow for new data to be collected in the future.
+
+# Use Cases
+## Referential Integrity
+One of the key features of the Salesforce platform is referential integrity capabilities. Referential integrity is how the platform allows a given resource to be safely updated and deleted without adverse cascading effects.
```
tedconn, line 35:
```suggestion
Referential integrity is how the platform allows a given resource to be safely updated and deleted without adverse cascading effects.
```

That first sentence is just fluff and doesn't really support this RFC. I would also provide an example of what referential integrity features this RFC aims to help with - for example renaming components, code, apex etc

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Editor support
+Metadata about components can be used augment standard developer experience in code editors with code completion, attribute name & type validation, peek definition.
```
tedconn, line 38:
```suggestion
Metadata about components can be used to augment the standard developer experience in code editors with code completion, attribute name & type validation, peek definition, etc.
```

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Documentation
+LWC components Metadata can be used to automatically generate documentation about components. Standard javascript documentation formats like JSDoc provided in component can also be gathered as metadata.
+
+## Dependency analysis and prefetching
+Components depend on static resources such as images, javascript files, css files and component definitions to name a few. For offline usage, knowing these dependencies at compile time allows the dependencies to be prefetched and avoid a network call.
+
+## Code generation
+Finally, metadata can facilitate low code development environment by allowing developer to declaratively define component. This metadata can be used to generate component stubs and the developer can add the specific business logic into the stubs thus reducing the amount of manually authored boiler plate code.
```
tedconn, line 47: Would be nice to have a proper java version of this to allow us to code-gen components on the server.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                        "$ref": "#/definitions/CustomElementReference"
+                    }
+                },
+                "staticResources": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/StaticResourceReference"
+                    }
+                },
+                "dynamicChildren": {
```
pmdartus, line 168: Can you provide more details on what `dynamicChildren` represents in comparison with `children`?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                        },
+                        "lwcDom": {
+                            "type": "array", 
+                            "items": { "$ref" : "#/definitions/TemplateLwcDomDirective" }
+                        }
+                    }
+                }
+            }
+        },
+        "JSFile": {
```
pmdartus, line 197: Let's plan for a future where we might add support for Typescript. Let's convert this type to `Script`.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+            "properties": {
+                "importType": { 
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "enum": [ "DefaultBinding", "NamedImports", "NamespacedImport"]
+                    }
+                },
+                "importsList": { "type": [ "array", "string"] },
+                "moduleSpecifier": { "$ref": "#/definitions/ModuleReference" }
```
pmdartus, line 487: We should add `Location` to all objects referencing the source code of the module. This includes the `Import` type.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                "eventHandlers": {
+                    "type": "array",
+                    "items": { "$ref": "#/definitions/EventHandlerReference" }
+                }
+            }
+        },
+        "DynamicImport": {
+            "type" : "object",
+            "properties": {
+                "moduleName": { "type": "string" },
```
pmdartus, line 405: This entry should be named `moduleSpecifier` to mirror the `Import` type.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                    "type": "array",
+                    "items": { 
+                        "$ref": "#/definitions/DynamicImport"
+                    }
+                },
+                "exports": {
+                    "type": "array",
+                    "items": [
+                        { "$ref": "#/definitions/Export"},
+                        { "$ref": "#/definitions/AggregatingExport"}
```
pmdartus, line 224: What is `AggregatingExport` referencing to?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                    ]
+                },
+                "programmaticEvents": {
```
pmdartus, line 227: Why calling this `programmaticEvents` instead just `Events` or `DOMEvents` ?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/ProgrammaticEvent"
+                    }
+                },
+                "staticResources": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/StaticResourceReference"
```
pmdartus, line 236: What would a static resource reference be in a JavaScript module?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                    }
+                }
+            }
+        },
+        "CSSFile": {
+            "type": "object",
+            "properties": {
+                "fileType": { "$ref": "#/definitions/FileType" },
+                "fileName": { "type": "string" },
+                "tokens": { "type": "array", "items": { "type": "string" }},
```
pmdartus, line 246: Let's remove this to `customProperties`. 

On top of listing all the custom properties present in the CSS, it might also be interesting to add the fallback value if present. This would be interesting information to surface.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+            "properties": {
+                "type": { "const": "api" },
+                "location": { "$ref": "#/definitions/SourceLocation" }
+            }
+        },
+        "Class": {
+            "type": "object",
+            "properties": {
+                "name": { "type": "string" },
+                "extends": { "$ref": "#/definitions/Class" },
```
pmdartus, line 284: Just to confirm, the `extends` value will be present if the base class is defined in the file file. Right?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                    "default": []
+                },
+                "methods": {
+                    "type": "array",
+                    "items": { "$ref": "#/definitions/ClassMethod" },
+                    "default": []
+                }
+            }
+        },
+        "ClassMemberType": {
```
pmdartus, line 297: This type is not used.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                    "enum": ["public", "private", "static"]
+                },
+                "decorator": { 
+                    "anyOf": [
+                        {"$ref": "#/definitions/ApiDecorator" },
+                        {"$ref": "#/definitions/TrackDecorator" },
+                        {"$ref": "#/definitions/WireDecorator" }
+                    ]
+                },
+                "parameterNames": {
```
pmdartus, line 316: This syntax doesn't account for the rest arguments operator.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+            }
+        },
+        "ClassProperty": {
+            "type" : "object",
+            "properties": {
+                "accessType": {
+                    "type": "string",
+                    "enum": ["public", "private", "static"]
+                },
+                "dataProperty": { "type": "boolean"},
```
pmdartus, line 333: What is this property used for?

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                "name": { "type": "string" },
+                "namespace": { "type": "string" },
+                "id": {
+                    "type": "string",
+                    "enum": [ "apexClass", "apexMethod", "apexContinuation", "client", "community", "component", "contentAssetUrl", "customPermission", "dynamicComponent", "slds", "messageChannel", "i18n", "gate", "label", "metric", "module", "internal", "resourceUrl", "schema", "sobjectClass", "sobjectField", "user", "userPermission"
+                    ]
+                },
+                "type": {
+                    "type": "string",
+                    "enum": [ "lwc", "salesforce", "internal", "external", "local"]
```
pmdartus, line 502: We should provide a definition for each of the types.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                "location": { "$ref": "#/definitions/SourceLocation" },
+                "adapterId": { "type": "string" },
+                "adapterModule": { "$ref": "#/definitions/ModuleReference" },
+                "adapterConfig": { "type": "object" }
+            }
+        }
+    }
+}
+```
+# Open questions
```
pmdartus, line 615:
```suggestion

# Open questions
```

text/0113-canonical-metadata.md
ravijayaramappa: It has a 'lwc:dynamic' directive. Updated the definition of DynamicComponentReference to add on to CustomElementReference. Will add a description to each property.

text/0113-canonical-metadata.md
ravijayaramappa: It is referring to re-exporting. I could rename it to reexports if that is more intuitive.
https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Re-exporting_Aggregating

text/0113-canonical-metadata.md
ravijayaramappa: A few examples:
1. Programmatically injected elements that load a static resource
2. Resources loaded using the platformResourceLoader.

There are more examples in this document from LMR(https://salesforce.quip.com/9WbbA0wTX9lE)

text/0113-canonical-metadata.md
ravijayaramappa: It should accommodate external class references as well. I have added a new ParentClass definition to accommodate imported classes.

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+* Should gathering metadata about configuration(`*.js-meta.xml`) file in a bundle be in the scope of this RFC?
```
moses-salesforce, line 616: I don't think we should include the *.js-meta.xml yet:
1. Large parts of the metadata in *.js-meta.xml will be revamped based on the Canonical UI Metadata RFC here:  https://salesforce.quip.com/ZJ93A0XQZvnU. If you include the *.js-meta.xml now it'll be enshrining the old system.
2. Some of the information in it is pretty Salesforce-specific and my impression here is that you're going for an open source thing.

Maybe in the future, when this information is more based on a standard and less Salesforce specific.

text/0113-canonical-metadata.md
ravijayaramappa: It represents a property defined on an instance. Our documents generalize them as properties. So I derived this term on the lines of data descriptor and accessor descriptor concept. Let me know if another name would describe it better.
```js
class MyComponent extends LightningElement {
         @track
         trackedProperty; // this is a data property

         privateField; // this is a data property

         get anotherProperty() { // For this one, hasGetter will be true
         }
}
```

text/0113-canonical-metadata.md
```diff
@@ -0,0 +1,625 @@
...
+                        "$ref": "#/definitions/StaticResourceReference"
+                    }
+                },
+                "dynamicChildren": {
+                    "type": "array",
+                    "items": { 
+                        "$ref": "#/definitions/DynamicComponentReference"
+                    }
+                },
+                "directives": {
```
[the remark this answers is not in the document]
diervo: As long is not a full AST... I don't think we need any innert HTML element that has no bindings.

text/0113-canonical-metadata.md
diervo: 🙀

text/0113-canonical-metadata.md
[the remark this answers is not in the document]
pmdartus: If we were to trim the AST, it means that we would have to maintain 2 different AST formats out of the compiler. Where do you see the value of the compiler trimming it down, while the consumer can do it on its end?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "A representation of metadata gathered for a LWC bundle",
+    "type": "object",
+    "properties": {
+        "name": { "type": "string" },
+        "namespace": { "type": "string" },
+        "moduleType": { 
+            "type": "string",
+            "enum": [ "Component", "Library", "CssOnly" ]
```
diervo, line 134: I'm not sure this enum will give any value nor if the values make sense

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+                        "$ref": "#/definitions/StaticResourceReference"
+                    }
+                },
+                "dynamicChildren": {
+                    "type": "array",
+                    "items": { 
+                        "$ref": "#/definitions/DynamicComponentReference"
+                    }
+                },
+                "directives": {
```
diervo, line 178: After thinking it more, and having debated this with @pmdartus in the past, I think we should just end up exposing a "pseudo AST": Doesn't contain all the nodes, but the nodes that has a meaning. ex: a node with an expression, directive, or any custom element.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+                        "$ref": "#/definitions/Import"
+                    }
+                },
+                "dynamicImports": {
+                    "type": "array",
+                    "items": { 
+                        "$ref": "#/definitions/DynamicImport"
+                    }
+                },
+                "exports": {
```
diervo, line 224: Add location to all things.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+                        { "$ref": "#/definitions/AggregatingExport"}
+                    ]
+                },
+                "domEvents": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/DOMEvent"
+                    }
+                },
+                "staticResources": {
```
diervo, line 237: What is this? Why is it separated?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+                    }
+                },
+                "exports": {
+                    "type": "array",
+                    "items": [
+                        { "$ref": "#/definitions/Export"},
+                        { "$ref": "#/definitions/AggregatingExport"}
+                    ]
+                },
+                "domEvents": {
```
diervo, line 231: What is this why is it separated?

text/0000-canonical-metadata.md
diervo: Also anything that contains a url will be interesting

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+                },
+                "staticResources": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/StaticResourceReference"
+                    }
+                }
+            }
+        },
+        "AggregatingExport": {
```
diervo, line 264: How is this different than ModuleRecord?

text/0000-canonical-metadata.md
apapko: Agree. The LNS team is already consuming our bindingAST, exposed as part of template metadata. They are also open for consuming an AST for js files as well.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+        * Public methods
+    * Programmatically generated events
+        * type
+        * options: composed, bubbles
+    * Programmatically fetched static resources
+        * `<script>`
+        * `<img>`
+        * `<link>`
+* CSS file
+    * Tokens
```
apapko, line 120:
```suggestion
    * Custom Properties ([--*](https://developer.mozilla.org/en-US/docs/Web/CSS/--*))
    * Parts ([::part](https://developer.mozilla.org/en-US/docs/Web/CSS/::part))
    * Tokens
```

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+    * Custom element references
+        * Tag name
+        * Attributes
+        * Slot content - default and named slots
+    * Static resource
+        * Image urls
+        * Javascript resource
+        * CSS resource
+    * Dynamic components
+    * Template directives
```
apapko, line 82: directives should probably fold under custom element references section. Alternatively, since we already expose bindingAST, which contains such information, perhaps the directives section can be left out all together

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+        * if
+        * iterator
+        * key
+        * lwc:dom="manual"
+    * Event handlers
+    * Slots
+* Javascript file
+    * Component classes
+        * Inheritance
+        * Public & private properties
```
apapko, line 93:
```suggestion
        * Public & private class members
```

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+        * Decorators
+        * getters/setters
```
apapko, line 95:
```suggestion
```

apapko: this is part of public and private properties (getter/setter is an implementation detail of authoring a property)

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+        * Public methods
```
apapko, line 96:
```suggestion
```

apapko: consolidate with line 93

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+    * Imports
+        * Builtin modules
+        * External modules
+        * Salesforce scoped modules
+        * Other modules(components and libraries)
+        * Relative imports
```
apapko, line 102: what is the use case for relative imports? any bundling tools, such as rollup, will inline relative imports during build time.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+        * Dynamic imports
```
apapko, line 103: should css imports be explicitly stated as well since we are listing everything? Alternatively we can rename imports to 'dependencies' which would imply any external module

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,673 @@
...
+    * Exports
+        * Default exports
+        * Named exports
+        * Aggregating modules
+    * JSDoc
+        * Class
+        * Properties
+        * Public methods
+    * Programmatically generated events
```
apapko, line 112: we should also include programmatically attached listeners

text/0000-canonical-metadata.md
apapko: should these fall under dependencies?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+### TypeScript
+
+#### Bundle metadata shape
+```ts
+interface BundleMetadata {
+    version: string;
+    name?: string;
+    namespace?: string;
+    // Define the type of module that is represented by the given metadata object.
+    moduleType?: "HTML" | "CSS" | "Javascript";
```
apapko, line 134: at bundle level we should be able to to deduce if the bundle is a 'Component' or 'LightningElement'

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+
+    /**
+     * For a component bundle, this property is a reference to the main component class in the bundle.
+     * This class must be the default export of the javascript file. The file must be named the same
+     * as the bundle
+     */
+    defaultComponentClass?: Class;
```
apapko, line 141: since this property is optional, i suggest we include 'exports', which would be an array of all javascript exports coming from the entry file, whether it is a class, function or a const

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+
+    // Information about HTML template files in the bundle.
+    templates?: HTMLTemplateFile[];
+
+    // Information about javascript files in the bundle.
+    scripts?: ScriptFile[];
+
+    // Information about css files in the bundle.
+    css?: CSSFile[]
+}
```
apapko, line 151: what about documentation? '.md' extension

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+    version: string;
+    name?: string;
+    namespace?: string;
+    moduleType?: "HTML" | "CSS" | "Javascript";
+
+    defaultComponentClass?: ComponentClass[];
+
+    templates?: { [key: string]: HTMLTemplateFile[]};
+
+    scripts?: { [key: string]: ScriptFile[]};
```
apapko, line 166: what about having all files under the same property? Such as sources: _{ [key: string] : File }_

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+```
+
+#### HTML Template file metadata shape
+```ts
+// Root metadata object for an HTML template file
+interface HTMLTemplateFile {
+    fileType: 'html';
+    fileName: string;
+    // List of unique components referenced in the template
+    componentReferences?: ComponentReference[];
```
apapko, line 179: what about using 'customElements' as property name instead?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+    astRoot?: Node;
+}
+
+// Represents a component referenced in the template.
+// Use the 'tagName' property for kebab case and 'name' property for the reference in canonical form
+interface ComponentReference extends ModuleReference {
+    // Component reference always has a namespace
+    namespace: string;
+    // No moduleIdentifier for a component
+    moduleIdentifier?: never;
```
apapko, line 192: what is this for? And if it is 'never' why have it at all?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+    // Components are always imported from an external bundle
+    type: 'external';
```
apapko, line 194: what is the purpose of this if it is always the same?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+    tagName: string;
+}
+
+// Information about a node in the html template
+interface Node {
+    tagName: string;
+    location: SourceLocation;
+    // Any properties set in the template, will include html attributes, public properties
+    attributes?: ElementAttributeReference[];
```
apapko, line 203: where do 'properties' fall into here?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+    // Event listeners attached declaratively in the template with a 'on' prefix, eg: 'onclick'
+    eventListeners?: DeclarativeEventListener[];
+    children?: Node[];
+}
```
apapko, line 207: depending on the AST implementation, we may want to have a reference to the parent node.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+ * specify a 'slot' attribute in the slotted content in the case of named slot. Slot content that
+ * do not have a 'slot' attribute will be passed as default slot content. The 'name' property in
+ * the SlotReference object will be 'default' for such a case.
+ */
+interface SlotReference {
+    name: string;
+}
+
+// Information about an element referenced in the template with a 'lwc:dynamic' attribute.
+interface DynamicCustomElementNode extends CustomElementNode {
```
apapko, line 226: should we be more explicit with the naming - LwcDynamicNodeDirective

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+
+// Information about for:each directive usage to render a list in a html template.
+interface TemplateForEachDirective extends Node {
+    tagName: 'template';
+    directive: 'for:each';
+    items: ComputedValue;
+    itemName?: string;
+    indexName?: string;
+    key?: string;
+}
```
apapko, line 238: scope? to indicate the nesting

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1232 @@
...
+
+// Information about if:true or if:false directive usage to perform conditional rendering in a html template.
+interface TemplateIfDirective extends Node {
```
apapko, line 241: do we want these explicitly typed? We could introduce TemplateDirective { tagName, directiveKey, directiveValue }

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1466 @@
...
+          // Location of the parent class identifier
+          location: SourceLocation;
+      };
+
+// Information about a module imported into a module using the 'import' statement.
+interface ModuleReference {
+    // Full canonical name, eg. lightning/input, lightning/recordForm
+    name: string;
+    namespace?: string;
+    moduleIdentifier?:
```
apapko, line 457: the notion of scoped modules is salesforce specific, i suggest that we don't introduce special treatment at the root level and treat every reference as a module. However, when we do encounter salesforce specific identifier, we can capture that information under a separate property 
```js
interface ModuleReference {
    // Full canonical name, eg. lightning/input, lightning/recordForm
    name: string;
    namespace?: string;
    type: 'module' | '@salesforce',
    sfdcResource?: {
         scoped: 'apexClass' | 'client' | 'slds' ... (the rest);
         namespacedId: string; // scoped module namespace normalization 
    }

```

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1466 @@
...
+        | 'internal'
+        | 'resourceUrl'
+        | 'schema'
+        | 'sobjectClass'
+        | 'sobjectField'
+        | 'user'
+        | 'userPermission';
+    // whether the module reference is the built in lwc module or the @salesforce scoped module or
+    // a local file in the current bundle or an module imported from an external source(eg. a component)
+    type: 'lwc' | '@salesforce' | 'internal' | 'external';
```
apapko, line 483: what are the last two types for? Today, the notion of 'internal' vs 'platform' is specific to compilation and are tied to the bundle not a reference.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1466 @@
...
+// Use the 'tagName' property for kebab case and 'name' property for the reference in canonical form
+interface ComponentReference extends ModuleReference {
+    // Component reference always has a namespace
+    namespace: string;
+    // No moduleIdentifier for a component
+    moduleIdentifier?: never;
+    // Components are always imported from an external bundle
+    type: 'external';
+    tagName: string;
+}
```
apapko, line 196: how does ComponentReference shape reflect the non-default namespace for which the current file is being compiled? As it stands, the metadata collection is done on a 'as authored' raw source and doesn't take custom namespace into account. For comparison, in the existing lwc-compiler meta collection flow, all the [Reference](https://git.soma.salesforce.com/lwc/lwc-api/blob/master/src/main/java/com/salesforce/lwc/reference/Reference.java#L20) objects surface 'namespacedId' property, which reflects the mapping from default to custom namespace. In addition, there is an assumption that all component dependencies, including template dependencies, will be namespaced by post-processing the compiler output, in 'amd' format, by applying the namespace mapping to the array of 'amd' module dependencies. 
Now back to the current proposal, the new metadata package will be invoked stand alone, outside of any namespace specific logic that resides in the compiler, thus the responsiblity for any namespace manipuations will have to be done by the consumer. With that, should we ease the namespacing pain and allow the meta package to accept an optional configuration to apply namespacing?

I propose we do. For example, if we encounter c-view-source template reference and the meta collection is invoked with a namespace mapping from the default 'c' to 'foo', then the namespace should be applied on all  ComponentReference properties. The question is, how do we surface it? Do we apply the namespace on the existing 'name' and 'tagName' properties or create the new ones? 

_template reference c-view-source_
_Example of an output with namespace applied to the existing props:_
```js
        {
            "name": "foo/viewSource",
            "namespace": "foo",
            "type": "external",
            "tagName": "foo-view-source"
        }
```
_Example of an output with adding namespacing as a new optional 'namespaceMapping' property_:
```js
        {
            "name": "c/viewSource",
            "namespace": "c",
            "type": "external",
            "tagName": "c-view-source",
            "namespaceMapping": {
                  "name": "foo/viewSource", 
                  "tagName": "foo-view-source"
             }
        }
```

text/0000-canonical-metadata.md
```diff
@@ -435,53 +559,76 @@ interface ProgrammaticEventListener {
...
+            | 'contentAssetUrl'
+            | 'customPermission'
+            | 'dynamicComponent'
+            | 'slds'
+            | 'messageChannel'
+            | 'i18n'
+            | 'gate'
+            | 'label'
+            | 'metric'
+            | 'module'
```
apapko, line 617: 'module' should be removed as it is not a scoped module

text/0000-canonical-metadata.md
ravijayaramappa: Can you elaborate how `namespace` and `namespacedId` different? I will incorporate that information into the RFC.

text/0000-canonical-metadata.md
ravijayaramappa: Given a module, the metadata gathering collects information about every import statement in the module. It is agnostic of what build tools are used post metadata collection. That is not the concern of the metadata collector. If the module has a relative import, that information should be surfaced.

text/0000-canonical-metadata.md
ravijayaramappa: It is an inherited property. Marking this a never here means that this metadata object is not expected to use that value.

text/0000-canonical-metadata.md
apapko: Per offline discussion, metadata will be namespace agnostic and will have to be performed by the consumed. Marking this conversation as resolved.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1198 @@
...
+interface CSSFile {
+    fileType: 'css';
+    fileName: string;
+    imports?: ModuleReference[];
+    customProperties?: CSSCustomProperty[];
+    staticResources?: StaticResourceReference[];
+}
+
+// Information about css custom property.
+interface CSSCustomProperty {
```
aputinski, line 779: Do we need a property to distinguish between declaration and usage?

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1198 @@
...
+    // Location starts with the first character of the member's identifier and ends at the semi-colon
+    // in case of a property and closing curly in case of a method
+    // For a accessor property type, the getter and setter property has the specific location
+    location?: SourceLocation;
+    // A string literal representing the leading comments of the class member
+    doc?: string;
+}
+
+// Information about a class property of a component class.
+interface ClassProperty extends ClassMember {
```
aputinski, line 395: Is there a proposed way of indicating `attribute` vs `property` when exposed via `buildCustomElementConstrutor`?

text/0000-canonical-metadata.md
ravijayaramappa: The shape has been updated. Declarations and references are called out separately.

text/0000-canonical-metadata.md
gordosfdc: Why do we want to surface internal information like that. No one outside the module should be able to see or access that information. I would argue that we should not surface that.

cseymourSF: we will need information about salesforce scoped imports in order to pre-cache resources (contentAssetUrl, resourceUrl, Java resources)

diervo: Once I put forward my RFC/proposal for per-file pipeline this will make more sense.

text/0000-canonical-metadata.md
```diff
@@ -0,0 +1,1208 @@
...
+        type:
+            | 'number'
+            | 'string'
+            | 'boolean'
+            | 'null'
+            | 'object'
+            | 'undefined'
+            | 'array'
+            | 'unresolved';
+        value?: any;
```
nsand, line 412: @ravijayaramappa As we discussed today, it would be useful to have the canonical shape for objects and arrays defined for `initialValue.value` and `static` field `value`s.

text/0000-canonical-metadata.md
aputinski: @ravijayaramappa Any reason this distinction between reference and declaration was removed?

pmdartus (approved):
The RFC looks good. Shall we get this merged?

PR #36, opened 2020-05-26
docs: clarify contribution steps

mpsenn:
This repository does not have CI or testing, so remove testing steps from contribution steps.


kevinv11n:
Rather than remove this I rather we add the validation steps we're using in other RFC repos. @diervo

mpsenn:
It looks like a couple other repos use these instructions.

diervo:
The remplate for rfcs is this https://github.com/salesforce/template-rfcs

mpsenn:
I will do this thing, if the language in this PR looks good to you.

PR #38, opened 2020-07-15
feat: secure script container RFC

tedconn:


text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+TBD
+
+## How we teach this
+
+TBD
+
+## Unresolved questions
+
+- Do we want to support all the existing `script` tag attributes like `type`, `async` and `defer`?
```
tedconn, line 183:
```suggestion
- Do we want to maintain parity with the `script` element including supporting attributes like `type`, `async` and `defer`? If so can we just extend [`HTMLScriptElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement)? 
```

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+By providing a secure container for scripts to run, see, and interact with all the shadow DOM trees as if it were the light DOM, application developers can continue to run their existing integrations until these libraries support shadow DOM semantics natively.
+
+## Basic example
+
+```html
+<secure-script src="//cdn.optimizely.com/js/12345678.js"></secure-script>
+```
+
+`secure-script` is a custom element which acts the same as the `script` tag, but any code which is evaluated inside this container is able to traverse the entire shadow tree using ligh DOM semantics. So if the above script runs `document.querySelectorAll('button')`, it will return all button elements regardless if they are in the shadow or not.
```
gregwhitworth, line 23: mis-spelling: `ligh` -> `light`

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+TBD
+
+## How we teach this
+
+TBD
+
+## Unresolved questions
+
+- Do we want to maintain parity with the `script` element including supporting attributes like `type`, `async` and `defer`? If so can we just extend [`HTMLScriptElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement)?
```
gregwhitworth, line 183: are there any downsides to doing this? I'd prefer to not have to go implement any of what you just outlined and given that this is effectively a privileged script tag I can see authors adding in those attr thinking they'll work.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Design
+
+`secure-script` will solve for the synthetic shadow and subsequently the native shadow.
+
+### Synthetic Shadow
+Many LWC applications apply the [synthetic-shadow polyfill](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow/). This polyfill emulates the native shadow behavior while still allowing global styles to cascade into the shadow trees. Synthetic shadow currently [patches](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow/src/polyfills/document-shadow) the `document.querySelector` and `document.querySelectorAll` APIs which is what prevents developers from penetrating the shadow DOM boundaries either in their LWC components or any other scripts added to the document. We only want to solve for the latter scenario.
+
+Ok, so what if we store a reference to the original, unpatched APIs before they get patched by our polyfill? And we let *only* our "secure scripts" access those APIs? That's the premise of this approach for synthetic shadow:
```
gregwhitworth, line 71: I think this paragraph could be re-worded. You ask two questions and then follow it with bullets. Would something like this work?

> To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components, we are currently proposing the following characteristics for the `secure-script` design:

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
+---
+title: Secure Script Container
+status: DRAFTED
+created_at: 2020-07-15
+pr: https://github.com/salesforce/lwc-rfcs/pull/38
+champion: Caridy Patiño, Ted Conn
+---
+
+# Secure Script Container for LWC Apps
```
gregwhitworth, line 9: How are we ensuring that the scripts are indeed secure? If we aren't doing anything to secure the script (eg: an allow list of approved scripts) then we should possibly make this `<lwc-script>` or something similar?

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+- Do not apply the shadow DOM polyfill at all: this would allow component authors to violate shadow DOM encapsulation which is not an option.
+- If we don't do this, then companies / websites building on our platform will not be able to integrate with common third party libraries.
+
+## Adoption strategy
+
+TBD
+
+## How we teach this
+
+TBD
```
gregwhitworth, line 179: I'll have more opinions on this following the answer to my question on adoption strategy

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Alternatives
+
+- **Allow shadow DOM to be toggled at a more granular level.** There is a forthcoming RFC for allowing component authors to decide if they want to opt in to shadow DOM or not. This would mean we can keep top-level elements and layout components in the light DOM, but base components or managed components can choose to safeguard their internals. Until we allow this, `secure-script` serves as a mitigation for application developers looking to add third party libraries to their apps.
+- Do not apply the shadow DOM polyfill at all: this would allow component authors to violate shadow DOM encapsulation which is not an option.
+- If we don't do this, then companies / websites building on our platform will not be able to integrate with common third party libraries.
+
+## Adoption strategy
+
+TBD
```
gregwhitworth, line 175: Is the expectation that the admin will click an application/extension for their site to add Google Analytics (for example) and it will be added by us into the `secure-script`? Or will the admins/devs place `secure-script` themselves and it may happen to be Google Analytics, Evergage,etc.

PR #39, opened 2020-07-15
Virtual Script Container

tedconn:


text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+By providing a secure container for scripts to run, see, and interact with all the shadow DOM trees as if it were the light DOM, application developers can continue to run their existing integrations until these libraries support shadow DOM semantics natively.
+
+## Basic example
+
+```html
+<secure-script src="//cdn.optimizely.com/js/12345678.js"></secure-script>
+```
+
+`secure-script` is a custom element which acts the same as the `script` tag, but any code which is evaluated inside this container is able to traverse the entire shadow tree using ligh DOM semantics. So if the above script runs `document.querySelectorAll('button')`, it will return all button elements regardless if they are in the shadow or not.
```
gregwhitworth, line 23: mis-spelling: `ligh` -> `light`

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+TBD
+
+## How we teach this
+
+TBD
+
+## Unresolved questions
+
+- Do we want to support all the existing `script` tag attributes like `type`, `async` and `defer`?
```
gregwhitworth, line 183: are there any downsides to doing this? I'd prefer to not have to go implement any of what you just outlined and given that this is effectively a privileged script tag I can see authors adding in those attr thinking they'll work.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Design
+
+`secure-script` will solve for the synthetic shadow and subsequently the native shadow.
+
+### Synthetic Shadow
+Many LWC applications apply the [synthetic-shadow polyfill](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow/). This polyfill emulates the native shadow behavior while still allowing global styles to cascade into the shadow trees. Synthetic shadow currently [patches](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow/src/polyfills/document-shadow) the `document.querySelector` and `document.querySelectorAll` APIs which is what prevents developers from penetrating the shadow DOM boundaries either in their LWC components or any other scripts added to the document. We only want to solve for the latter scenario.
+
+Ok, so what if we store a reference to the original, unpatched APIs before they get patched by our polyfill? And we let *only* our "secure scripts" access those APIs? That's the premise of this approach for synthetic shadow:
```
gregwhitworth, line 71: I think this paragraph could be re-worded. You ask two questions and then follow it with bullets. Would something like this work?

> To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components, we are currently proposing the following characteristics for the secure-script design:

text/0113-secure-script-container.md
tedconn: Yeah. I was going to test if we can extend `HTMLScriptElement` and get some of this behavior for free.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
+---
+title: Secure Script Container
+status: DRAFTED
+created_at: 2020-07-15
+pr: https://github.com/salesforce/lwc-rfcs/pull/38
+champion: Caridy Patiño, Ted Conn
+---
+
+# Secure Script Container for LWC Apps
```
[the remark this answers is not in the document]
tedconn: They are secure because they are running inside a secure javascript sandbox. I'm not tied to the name but `<lwc-script>` implies this is an LWC-only concept and it actually would work for any web component application.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Alternatives
+
+- **Allow shadow DOM to be toggled at a more granular level.** There is a forthcoming RFC for allowing component authors to decide if they want to opt in to shadow DOM or not. This would mean we can keep top-level elements and layout components in the light DOM, but base components or managed components can choose to safeguard their internals. Until we allow this, `secure-script` serves as a mitigation for application developers looking to add third party libraries to their apps.
+- Do not apply the shadow DOM polyfill at all: this would allow component authors to violate shadow DOM encapsulation which is not an option.
+- If we don't do this, then companies / websites building on our platform will not be able to integrate with common third party libraries.
+
+## Adoption strategy
+
+TBD
```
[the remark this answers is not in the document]
tedconn: An application developer can choose to expose it or not, so I think it could be both. For Salesforce Communities / Commerce we would probably do the latter since we side on giving the admin the markup.

text/0113-secure-script-container.md
diervo: I think the semantics of the `<secure-script>` will be by definition async, otherwise we will have to generate an equivalent imperative inline version so evaluation is sync.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+## Basic example
+
+```html
+<secure-script src="//cdn.optimizely.com/js/12345678.js"></secure-script>
+```
+
+`secure-script` is a custom element which acts the same as the `script` tag, but any code which is evaluated inside this container is able to traverse the entire shadow tree using light DOM semantics. So if the above script runs `document.querySelectorAll('button')`, it will return all button elements regardless if they are in the shadow or not.
+
+```html
+<secure-script>
```
diervo, line 26: I would love to see the equivalent imperative inline example, something like:

```js
SecureScript.evaluate(() => {});
```

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+integrations into the components themselves. When every component's root element is a shadow
+root, there is no light DOM for them to work with, and their functionality is severely crippled.
+
+Some examples of these libraries are:
+
+- **Analytics tools**: Google Tag Manager / Google Analytics, Pendo
+- **Personalization platforms**: Optimizely, Evergage, Google Optimize
+- **Instrumentation tools**: New Relic, Sentry.io
+
+In addition, developers are finding that this model is not compatible with their **design systems** and **testing frameworks**.
```
diervo, line 60: I would argue that design systems will not fit into this model since they will still be bound to the shadow semantics

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+  * This has to be done before the synthetic shadow is imported
+
+* Run or include the third party library script inside this sandbox
+  * This allows the code to run in a non-patched environment, while the rest of the document is running synthetic shadow
+* Package the above functionality in a web component which semantically matches the `<script>` tag.
+
+
+### Native Shadow
+
+For applications which are using the native shadow, the underlying implementation will be different. There are two approaches we can choose from:
```
diervo, line 84: We should add way more detail here about how we are thinking on implementing this as we move forward

text/0113-secure-script-container.md
tedconn: I agree, this doesn't apply here.

text/0113-secure-script-container.md
tedconn: What is the motivation for matching imperative APIs, just programmatic convenience? Also there's nothing stopping us from having both, right?

```typescript
SecureScript.evaluateSync
SecureScript.evaluate
```

text/0113-secure-script-container.md
tedconn: I agree, I think it's important to understand that now, even if its implementation comes later.

text/0113-secure-script-container.md
pmdartus: I do think like @diervo that `<secure-script>` should be async by nature but for different reasons. Preserving the default script semantic doesn't behavior would require to force HTML parsing to stop, and I don't feel that we should be in this business.

I would be supportive of adding the `defer` attribute.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,181 @@
...
+<script>
+    window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+    ga('create', 'UA-XXXXX-Y', 'auto');
+    ga('send', 'pageview');
+</script>
+<script async src='https://www.google-analytics.com/analytics.js'></script>
+<!-- End Google Analytics -->
+```
+
+If we included the above snippet inside our secure-script container, then `ga` will only be available to the *inner* `window`, i.e. the secure container's window. So `ga` will not be usable by anyone else in the entire application. We can however, allow for globals to be created inside our secure container, and redefined in any additional application sandboxes created by Locker (or the global outer window).
```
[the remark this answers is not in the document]
tedconn: I'm referring to the first inline script but we can assume the second one is referencing globals defined in the first one.

text/0113-secure-script-container.md
diervo: Im not sure how we will be able to honor `defer` though

text/0113-secure-script-container.md
diervo: That's a very fair point, we can do a follow up, however I would love to mention it here somewhere.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
tedconn: @gregwhitworth you said you had more opinions here?

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,181 @@
...
+
+A syntax like the following could be provided:
+
+```html
+<secure-script extra-globals="ga"></secure-script>
```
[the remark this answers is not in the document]
tedconn: > Just to confirm extra-globals="ga" means that the secure script can set the ga property on the global object. Right?

Yes, more specifically it means that any globals defined by a script in the secure container, are re-set in the outer window.

The secure script container does not have access to globals defined by other scripts. i.e. extra-globals does not define globals to keep in sync between the outer window and the secure script container. Rather it should copy any globals defined to the outer window.

So in your example it would log `undefined` even if `extra-globals` was set. I admit `extra-globals` may imply it's a pass-in.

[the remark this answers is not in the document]
pmdartus: > The secure script container does not have access to globals defined by other scripts

If it's the case, I don't see how the Google Analytics example works. The `analytics.js` does expect `ga` to be available on the `window` in order to run.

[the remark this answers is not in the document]
tedconn: The secure script would be the first one, as I said above, which defines the globals, like this:

```html
<!-- Google Analytics -->
<secure-script extra-globals="ga">
    window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
    ga('create', 'UA-XXXXX-Y', 'auto');
    ga('send', 'pageview');
</script>
<script async src='https://www.google-analytics.com/analytics.js'></script>
<!-- End Google Analytics -->
```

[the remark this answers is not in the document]
pmdartus: Shouldn't the analytics scripts be executed as well in a `secure-script` in order to work across shadows transparently?

text/0113-secure-script-container.md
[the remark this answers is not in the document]
caridy: yeah, this is misleading for sure. We will change the name from `secure` to `virtual` since the goal here is not to secure anything, but to virtualize it.

text/0113-secure-script-container.md
caridy: yes, we can mention it. I believe a static method on the VirtualScript web component class is sufficient, so you could import the class, and register yourself, extend it, or just call the static evaluator method. I will add a note for that to be a follow up proposal.

text/0113-secure-script-container.md
caridy: will do in a bit!

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+## Adoption strategy
+
+TBD
+
+## How we teach this
+
+TBD
+
+## Unresolved questions
+
```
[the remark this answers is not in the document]
caridy: the budget is around 5k. To support native shadow, it might increase a little bit more. @tedconn can you compute the minified/gzip size of the current demo for synthetic shadow? just to have a good reference.

text/0113-secure-script-container.md
caridy: Ok, in today's demo, it supports everything supported by the original script tag, in fact, the way it works is that `<virtual-script>` becomes `<script>` with the same set of attributes and content inside the sandbox.

Now, many of the issues about scripts are for script tags parsed and executed as part of the html document, which is NEVER the case for `<virtual-script>` because it is always inserted dynamically inside the sandboxed iframe. In which case, browsers will do a good job no matter what to not block anything else. I don't think we need to complicate this more.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
caridy: ok, changes are in, please, take another look.

text/0113-secure-script-container.md
caridy: added... please review

text/0113-secure-script-container.md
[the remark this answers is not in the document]
caridy: yes, both should be executed under the `<virtual-script>`

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+* Package the above functionality in a web component which semantically matches the `<script>` tag.
+
+
+### Native Shadow
+
+For applications which are using the native shadow, the underlying implementation will be different. There are two approaches we can choose from:
+
+1. Virtualize every shadow DOM tree in the outer window into a single DOM structure which this container would see as the default DOM.
+2. Implement a `document.querySelector` API which traverses the shadow DOM naturally.
+
```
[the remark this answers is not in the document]
caridy: why do you think that native shadow is not possible? it is certainly more difficult to do, in fact it is the inverse implementation of the synthetic shadow polyfill, but we do have the expertise now, we can make it happen.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
caridy: added the performence budget everywhere...

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,183 @@
...
+To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components, we are currently proposing the following characteristics for the secure-script design:
+
+* Create a secure JS sandbox which doesn’t allow its object graph to be extended using [`Object.preventExtensions`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions)
+  * The sandbox is created using a same-domain iframe provided by Locker v.Next
+  * This has to be done before the synthetic shadow is imported
```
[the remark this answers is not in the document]
caridy: because the goal is to make this to work with native shadow as well, we should not assume synthetic is present.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
priandsf: Yeah, not saying that it is impossible, I trust that you can do the magic. But as you said it is going to be more difficult, so we could use the synthetic shadow dom until the tools are adapted, which I hope will happen (see, for example, what codecept.js is doing: https://codecept.io/shadow/). At least we can get Evergage to adopt a similar approach. For GTM, I'm seeing requests being sent at Google although I didn't see any roadmap yet.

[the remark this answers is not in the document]
caridy: makes sense! I'm not that optimistic about tools being able to overcome the encapsulation paradigm dictated by shadow dom.

text/0113-secure-script-container.md
pmdartus: Since the elements are dynamically inserted into the document sandbox, the original script semantic is not preserved regarding timing and evaluation order. In the following example, we can ensure that `script1.js` is evaluated before `script2.js`.

```html
<script src="script1.js"></script>
<script src="script2.js"></script>
```

Now in the next example, we can guarantee the evaluation order since script processing is not done by the HTML parser. In some cases, `script2.js` might evaluate before `script1.js`.

```html
<virtual-script src="script1.js"></virtual-script>
<virtual-script src="script2.js"></virtual-script>
```

This is why I think one of the invariant here is that all `virtual-script`s should be considered `async` by nature.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
tedconn: Yes, the point here was just to illustrate the global being referenced across script containers, both virtual and standard.

text/0113-secure-script-container.md
```diff
@@ -1,88 +1,121 @@
...
 - **Personalization platforms**: Optimizely, Evergage, Google Optimize
 - **Instrumentation tools**: New Relic, Sentry.io
 
-The goal of this proposal is to provide a way for application developers to continue to use their existing integrations until library authors support shadow DOM integration.
+The goal of this proposal is to provide a way for application developers to continue to use their existing integrations until library authors support Shadow DOM integration.
 
 ## Design
 
-`secure-script` will solve for the synthetic shadow and subsequently the native shadow.
+A new element called `<virtual-script>` encapsulates the various mechanism to evaluate the scripts in a virtual environment, which solves the problem when using synthetic shadow and subsequently, when using the native Shadow DOM. We can implement this solution in two phases:
```
tedconn, line 63:
```suggestion
A new element called `<virtual-script>` encapsulates the various mechanisms to evaluate the scripts in a virtual environment, which solves the problem when using synthetic shadow and subsequently, when using the native Shadow DOM. We can implement this solution in two phases:
```

text/0113-secure-script-container.md
```diff
@@ -1,88 +1,121 @@
...
 
-* Create a secure JS sandbox which doesn’t allow its object graph to be extended using [`Object.preventExtensions`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions)
-  * The sandbox is created using a same-domain iframe provided by Locker v.Next
-  * This has to be done before the synthetic shadow is imported
+To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components running on the main window. We are currently proposing the following characteristics for the virtual-script design:
```
tedconn, line 69:
```suggestion
To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components running on the main window, we are currently proposing the following characteristics for the virtual-script design:
```

text/0113-secure-script-container.md
```diff
@@ -1,88 +1,121 @@
...
+   * operation budget: 50% slower than executing the same task is the main window.
+   * 3rd party scripts are rarely part of the critical path for the application to function.
+
+### Phase II: Native Shadow
+
+For applications which are using the native shadow, the underlying implementation will be different, a more complicated one. We are currently proposing the following characteristics for the virtual-script design:
+
+* During the evaluation of the virtual script component class, patch `Element.prototype.attachShadow`, to record any shadow created in the app.
+
+* When evaluating code via `<virtual-script>` element, create a new sandbox with patched DOM APIs to simulate a flatten tree.
```
tedconn, line 100: Why are we proposing to flatten instead of preserving the original DOM's hierarchy?

> This will require a query selector mechanism that can translate regular selectors into cross shadow selections

Eg: how will we reason with matching `element1 > element2` and *not* matching `element1 + element2`

text/0113-secure-script-container.md
caridy: I see... this is a very valid point, thanks for raising this issue @pmdartus , let me think more about it.

text/0113-secure-script-container.md
caridy: We are talking about the native shadow here, so, there is no DOM hierarchy here that can work, we have to simulate it. I'm not sure I understand the question, but the point is, that in order for those 2 queries to function, specially when element1 is in one shadow, and element2 is in another shadow... how can you make that work? well, you have to reconstruct in memory a flatten representation of the intersection of those two shadow fragments, and do a query based on that... obviously that will be too slow, too complicated, but we can do something that represents that without having to construct it in memory. I have some ideas.

tedconn: Ok. When we say a flattened tree to me sounds like we would take something like this:

```html
<x-app>
  ▼ #shadow-root
    <x-child>
      ▼ #shadow-root
        <button />
    </x-child>
</x-app>
```

and flatten it to something like this:

```html
<x-app />
<x-child />
```

caridy: no, flatten tree reference to how browsers compute the UI elements, in this case, the flatten tree is:

```
x-app
   x-child
      button
```

text/0113-secure-script-container.md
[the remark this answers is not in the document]
gregwhitworth: I'm hesitant to allow this to be a free-for-all due to this statement in your RFC

> We are potentially prolonging the life of our synthetic shadow dom polyfill, instead of encouraging community library authors to adopt Shadow DOM semantics.

I would prefer for us to leverage the top libs identified and give them access to this for Phase 1 and when the need for synthetic shadow is removed we can update the <n> libs to leverage Phase 2.

Is this a large request outside of these large products?

text/0113-secure-script-container.md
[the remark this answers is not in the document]
gregwhitworth: > The secure script container does not have access to globals defined by other scripts. i.e. extra-globals does not define globals to keep in sync between the outer window and the secure script container. Rather it should copy any globals defined to the outer window.

I think this should be defined in the spec, at least as a note.

> Yes, the point here was just to illustrate the global being referenced across script containers, both virtual and standard.

Does this negate @pmdartus point that due to not being able to access other globals that GA won't work?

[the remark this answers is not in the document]
tedconn: I think it seems like a good best practice but technically it should negate @pmdartus point since it would work, any globals created will be available in all sandboxes, and yeah I'll add that to the spec.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
tedconn: > we can update the libs

That part is throwing me off a bit. The only work on us will be to support a native shadow implementation of `virtual-script` but this will be transparent for libraries. I'm certainly misunderstanding something.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
tedconn: @caridy if that is the case then the shadow dom implementation will be much more important here. Do we need to flesh out any more details for how that will work?

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,178 @@
...
+
+To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components running on the main window, we are currently proposing the following characteristics for the virtual-script design:
+
+* Create a JS sandbox before the Synthetic Shadow Polyfill is evaluated.
+  * The sandbox is created using a same-domain iframe.
+
+* Use this JS sandbox to virtualize the DOM APIs in order to prevent these APIs to be subject to the Synthetic Shadow Polyfill patches:
+  * This can be done by caching references to the to-be-patched APIs inside the sandbox before the synthetic shadow is evaluated.
+
+* Evaluate one or more third party libraries inside this sandbox
```
tedconn, line 77: I think we may need a few more details here about how exactly this code gets evaluated. In the POC we are creating a `script` element and appending it to the `magicBody`. Does that actually get added to the `virtual-script`'s Shadow DOM? What if it fails?

text/0113-secure-script-container.md
tedconn: The order is still preserved when multiple script tags are set as async, so forcing async doesn't solve the problem. For example:

```html
<script src="script1.js" async></script>
<script src="script2.js" async></script>
```

`script2` is always going to evaluate only after `script1` has been downloaded and evaluated.

@caridy can we insert each script into a single sandbox instead of having a separate sandbox per `virtual-script` tag?

caridy: Initially, that was my assumption, that they will be inserted on the same sandbox, we only built one... or a way to identify them somehow to be anchored to one or a handful of sandboxes.

pmdartus: > The order is still preserved when multiple script tags are set as async, so forcing async doesn't solve the problem.

@tedconn This is not the case, there is no execution order between async scripts.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Phase I: Synthetic Shadow
+
+Many LWC applications use the [synthetic-shadow polyfill](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow/). This polyfill emulates the native Shadow DOM behavior while still allowing global styles to cascade into the shadow trees. Synthetic Shadow patches many of the DOM APIs that allow you to interact and traverse the DOM tree, (e.g.: `document.querySelector` and `document.querySelectorAll`), these APIs will prevent developers from penetrating the Shadow DOM boundaries defined by their LWC components. In this phase, we want to solve the LWC + Synthetic Shadow scenario.
+
+To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components running on the main window, we are currently proposing the following characteristics for the virtual-script design:
+
+* Create a JS sandbox before the Synthetic Shadow Polyfill is evaluated.
```
pmdartus, line 71: Now that I think about it, what is the story around IE11? How does `<virtual-script>` behave when Locker can't be used?

tedconn: We aren't looking to support IE11 with `<virtual-script>`

text/0113-secure-script-container.md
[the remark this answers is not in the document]
gregwhitworth: Thanks, and sorry for the delayed response. We've since discussed this further but it comes down to my concern of compat risk, eg: who can use this and when? If we allow anyone to utilize it and run any code they want to bypass the shadow then we need to never close that hole. That puts a technical debt on the LWC team to maintain that and it may come with a whole host of other issues such as other teams that desire the encapsulation (for the very reason it exists) and them not being able to evolve due to this script. So ultimately I'd like to understand this plan further and given the business scenarios I'm aware of I think there can be a path forward that not only blocks us from a compat/tech debt perspective but also in the prodsec arena as well.

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,207 @@
...
+
+Note: these global values are controlled side-channels between the main window and the sandbox, and should be used with caution. For example, creating a component that depends on `ga` to be defined, is an anti-pattern.
+
+### Supported Attributes
+
+`crossorigin`: indicates that the browser should provide more information to window.onerror when handling CORS errors
+
+`defer`: indicates that the script should be executed after the document has parsed (has no effect if `src` is absent)
+
+`integrity`: enables the browser to verify thata fetched resource is delivered without unexpected manipulation
```
nrobertdehault, line 153:
```suggestion
`integrity`: enables the browser to verify that a fetched resource is delivered without unexpected manipulation
```

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,207 @@
...
+
+`nomodule`: the script should not be executed in browsers that support ES2015 modules
+
+`nonce`: user provided nonce values for CSP will be supported
+
+`referrerpolicy`: indicates which referrer to send when fetching the focument
```
nrobertdehault, line 159:
```suggestion
`referrerpolicy`: indicates which referrer to send when fetching the script
```

text/0113-secure-script-container.md
```diff
@@ -0,0 +1,207 @@
...
+
+`crossorigin`: indicates that the browser should provide more information to window.onerror when handling CORS errors
+
+`defer`: indicates that the script should be executed after the document has parsed (has no effect if `src` is absent)
+
+`integrity`: enables the browser to verify thata fetched resource is delivered without unexpected manipulation
+
+`nomodule`: the script should not be executed in browsers that support ES2015 modules
+
+`nonce`: user provided nonce values for CSP will be supported
```
nrobertdehault, line 157: Will we be able to support CSP's `<hash-algorithm>-<base64-value>` to enable execution of inline virtual scripts?

text/0113-secure-script-container.md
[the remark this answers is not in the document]
tedconn: @gregwhitworth is it fair to say that the usage of `virtual-script` is outside of the scope of this RFC? I can certainly flesh that out in the "Rollout strategy". Something to the tune of "usage of virtual script containers require above average scrutiny because of a, b, c..."?

text/0113-secure-script-container.md
tedconn: As long as virtual-script doesn't mutate the textContent of the tag then it shouldn't have an impact on that.

text/0113-secure-script-container.md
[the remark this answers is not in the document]
tedconn: We discussed this offline and agreed to something like the above.

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,207 @@
...
+## Summary
+
+We need a way for application developers to integrate common third-party libraries into their Web Component based apps (specifically, LWC applications). These libraries, such as Google Analytics, are severely hindered by the Shadow DOM which prevents global interaction with any element on the page.
+
+By providing a virtual container for scripts to run, see, and interact with all the Shadow DOM trees as if it were the light DOM, application developers can continue to run their existing integrations until these libraries support Shadow DOM semantics natively.
+
+## Basic example
+
+```html
+<virtual-script src="//cdn.optimizely.com/js/12345678.js"></virtual-script>
```
pmdartus, line 20:
```suggestion
<virtual-script src="https://cdn.optimizely.com/js/12345678.js"></virtual-script>
```

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,207 @@
...
+ * performance penalties will be added to the virtual scripts as they run inside a membrane:
+   * evaluation budget: 30% slower than evaluating the same script in the main window.
+   * operation budget: 50% slower than executing the same task is the main window.
+   * 3rd party scripts are rarely part of the critical path for the application to function.
+
+### Phase II: Native Shadow
+
+For applications which are using the native shadow, the underlying implementation will be different, a more complicated one. We are currently proposing the following characteristics for the virtual-script design:
+
+* During the evaluation of the virtual script component class, patch `Element.prototype.attachShadow`, to record any shadow created in the app.
```
pmdartus, line 98: Patching `Element.prototype.attachShadow` will not be sufficient to intercept all the shadow trees creations if the application is using the declarative shadow DOM syntax.

We can wait for the `DOMContentLoaded` event and retrieve the list of all the open shadow trees created via the declarative shadow DOM syntax as part of the initial page load. I don't there will be an easy way to find all the other shadow trees created via the declarative shadow after that. But I guess it's fine.

pmdartus (approved):

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+    * The sandbox does NOT include any global name defined in the main window, whether it is defined before or after the creation of the sandbox. But the `<virtual-script>` can opt-in to share a global value between the sandbox and the main window, this must be done via the `extra-globals` argument passed to the `<virtual-script extra-globals="dataLayer">` element.
+
+Other considerations for Phase I:
+
+ * due to the explicit order of evaluation required for this to function, we plan to only support one sandbox that can be shared by multiple virtual scripts.
+ * budget is around 5kb (minify/gzip).
+ * performance penalties will be added to the virtual scripts as they run inside a membrane:
+   * evaluation budget: 30% slower than evaluating the same script in the main window.
+   * operation budget: 50% slower than executing the same task is the main window.
+   * 3rd party scripts are rarely part of the critical path for the application to function.
```
matt-buland-sfdc, line 92: How can you be so sure? Suggesting this is different from claiming it as fact.

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+  * This allows the code to run in a non-patched environment, while the rest of the document is running synthetic shadow.
+
+* Package the above functionality in a web component which semantically matches the `<script>` tag API.
+
+* Provide an additional mechanism to remap globals when needed.
+    * The sandbox does NOT include any global name defined in the main window, whether it is defined before or after the creation of the sandbox. But the `<virtual-script>` can opt-in to share a global value between the sandbox and the main window, this must be done via the `extra-globals` argument passed to the `<virtual-script extra-globals="dataLayer">` element.
+
+Other considerations for Phase I:
+
+ * due to the explicit order of evaluation required for this to function, we plan to only support one sandbox that can be shared by multiple virtual scripts.
```
matt-buland-sfdc, line 87: multiple virtual-scripts at what layer? 1 sandbox for the whole page? 1 sandbox per shadow?

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Phase I: Synthetic Shadow
+
+Many LWC applications use the [synthetic-shadow polyfill](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow/). This polyfill emulates the native Shadow DOM behavior while still allowing global styles to cascade into the shadow trees. Synthetic Shadow patches many of the DOM APIs that allow you to interact and traverse the DOM tree, (e.g.: `document.querySelector` and `document.querySelectorAll`), these APIs will prevent developers from penetrating the Shadow DOM boundaries defined by their LWC components. In this phase, we want to solve the LWC + Synthetic Shadow scenario.
+
+To ensure we can solve the problem while continuing to guarantee the benefits that are provided by the synthetic-shadow polyfill for all other components running on the main window, we are currently proposing the following characteristics for the virtual-script design:
+
+* Create a JS sandbox before the Synthetic Shadow Polyfill is evaluated.
+  * The sandbox is created using a same-domain iframe.
```
matt-buland-sfdc, line 72: Doesn't this present a security concern because of the allowances afforded to same-domain scenarios? While many globals and APIs are restricted, separate domains put up very firm walls that are heavily enforced by the browser.

text/0113-virtual-script-container.md
tedconn: For the use cases we are targeting, this is a fact. i.e. B2C applications with third party integrations for tracking, analytics, or personalization. We aren't trying to load application framework code and that is not the use case for `virtual-script`

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+
+* Currently there is an outstanding issue with Synthetic Shadow in that the Event object patching happens lazily, so we are unable to prevent its patching at the time of creating the sandbox, which means that events are still re-targeted in some cases. [See LWC PR #1569](https://github.com/salesforce/lwc/pull/1569).
+
+### Testing
+
+We will want to gather the top 20 common libraries and implement their behavior inside of `virtual-script` to ensure that we are properly accounting for all these libraries' use cases. This will be our integration test suite.
+
+## Drawbacks
+
+- The element may end up being anywhere from 5kb - 20kb.
```
aputinski, line 195: Is this something that can be extracted as a plugin? Seems unfortunate that I'd have to incur 5-20kb and all the complexities it brings (patching native methods) for a feature that seems intended for platform?

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+If we evaluate the above snippets with `<virtual-script>` instead, then `ga` global value will only be available into the sandbox. So `ga` will not be usable by anyone else in the entire application. We can however, allow for globals to be exposed into the main window. We call this global _exporting_. 
+
+Conversely, if a script running in `virtual-script` needs access to a global variable from the main window, we can _import_ this global. 
+
+A syntax like the following will be provided:
+
+```html
+<virtual-script exported-globals="foo"></virtual-script>
```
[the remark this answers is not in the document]
tedconn: They are read-only by design. @caridy can confirm if they are going to stay that way.

text/0113-virtual-script-container.md
tedconn: This feature will not be a part of core LWC. In fact it's intended to be used only in our b2c products today and will be a separate library.

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+
+* When evaluating code via `<virtual-script>` element, create a new sandbox with patched DOM APIs to simulate a flatten tree.
+  * All APIs that allow you immediate traversal up and down (e.g.: `elm.parentNode`) must be patched.
+  * All Event retargeting mechanism must be unwind to flatten them up.
+  * All deep traversal (e.g.: `elm.querySelector`) must be patched, and DOM filtering must be implemented by simulating the flatten representation of the application. This will require a query selector mechanism that can translate regular selectors into cross shadow selections (which is not trivial work)
+
+* Package the above functionality in a web component which semantically matches the `<script>` tag API.
+
+* Provide an additional mechanism to remap globals when needed.
+    * The sandbox does NOT include any global name defined in the main window, whether it is defined before or after the creation of the sandbox. But the `<virtual-script>` can opt-in to share a global value between the sandbox and the main window, this must be done via the `extra-globals` argument passed to the `<virtual-script extra-globals="dataLayer">` element.
```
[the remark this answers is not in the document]
tedconn: Good catch, I think this case is `imported-globals`, let me update.

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+
+Other considerations for Phase II:
+
+ * the virtual script web component "should" be registered before any shadow root is created, it is not longer a "must", the flip side of this is that pre-existing shadowRoots are just not going to be inspected by virtual scripts.
```
CommanderQ, line 112: Trivial typo: "no longer" rather than "not longer". 😉

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+
+`src`: the address of the resource
+
+**async**
+
+TBD describe asynchronous nature of `virtual-script`.
+
+### Unsupported Attributes
+
+`type`: all `virtual-script` resources will be `text/javascript`. `type="module"` will not be supported.
```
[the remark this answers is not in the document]
tedconn: Yeah I believe so, although I think the request would still happen, just browser cached?

text/0113-virtual-script-container.md
tedconn: One sandbox per page, I'll update

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,213 @@
...
+These libraries are typically added as global scripts to the document root rather than deeper 
+integrations into the components themselves. When every component in the application has a ShadowRoot attached to them, these libraries will not work as expected.
+
+Some examples of these libraries are:
+
+- **Analytics tools**: Google Tag Manager / Google Analytics, Pendo
+- **Personalization platforms**: Optimizely, Evergage, Google Optimize
+- **Instrumentation tools**: New Relic, Sentry.io
+
+The goal of this proposal is to provide a way for application developers to continue to use their existing integrations until library authors support Shadow DOM integration.
```
jbenallen, line 59: This makes sense to enable adoption. It also raises some significant concerns. Once this is released, developers can write any code they want and load that code using a `<virtual-script>` element to bypass the encapsulation of each web component. This a free escape hatch which breaks the principle of encapsulation by its very design. That opens the door to developers using this intentionally to write code which relies on implementation details inside a web component.

1) What is the planned support statement for using `<virtual-script>` when code it executes "breaks" when changes occur to the application around it?
2) Are there plans to support `<virtual-script>` on Salesforce platform?

text/0113-virtual-script-container.md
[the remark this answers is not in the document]
jbenallen: How does the ability to modify the `globalThis` through `export-globals` align with Locker protections? Can `<virtual-script export-globals="...">` be used to pollute/poison the `globalThis` which is used by modules in multiple packages/namespaces on Salesforce platform?

text/0113-virtual-script-container.md
matt-buland-sfdc: Who would be allowed to use `<virtual-script>` ? It seems like an LWC itself is disallowed from using `<script>`, and it's not generally available, even internally. So this would show up in root HTML documents, which means it's not really for LWC, it's for an app / document bootstrap to define.

matt-buland-sfdc: I see this described in the first paragraph. I suppose there are no additional security concerns unless this is made available to a broader audience, which I suspect security experts would be resistant to.

tedconn: > which I suspect security experts would be resistant to

I'm curious what security vulnerabilities you see beyond what's already possible with a regular <script> tag?

text/0113-virtual-script-container.md
tedconn: > That opens the door to developers using this intentionally to write code which relies on implementation details inside a web component.

Developers can already bypass that by traversing the shadow roots, right?

> What is the planned support statement for using <virtual-script> when code it executes "breaks" when changes occur to the application around it?

That is a great question and something @gregwhitworth was alluding to before, but we decided that discussion was outside of the scope of this RFC, and better suited for a security assessment when the usage of `virtual-script` is allowed in a Salesforce product.

> Are there plans to support <virtual-script> on Salesforce platform?

Currently only B2C experiences will support this element, and we will have a corresponding threat analysis. That being said, [we already allow Locker Disablement](https://developer.salesforce.com/docs/atlas.en-us.communities_dev.meta/communities_dev/communities_dev_security_locker.htm) and B2C experiences don't share the same security concerns as the rest of the platform.

text/0113-virtual-script-container.md
matt-buland-sfdc: Beyond script, nothing. I was imagining a solution that met some "ideal" model of secure, such as running script in a separate domain/iframe, and proxying capabilities through some secure channel, so that it would be safe to let any component use/import anything in the context of its shadow. This solution doesn't try to do that, nor does it need to.

text/0113-virtual-script-container.md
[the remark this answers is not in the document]
tedconn: Yes we have talked about having an allow list of globals which can be mutated. I believe `globalThis` is one of those we need to protect against.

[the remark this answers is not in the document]
tedconn: That being said, we are not really setting out to provide tons of protection against malicious scripts, since only application owners can use `virtual-script` and the same security concerns exist with using a regular script tag.

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,214 @@
...
+
+- `virtual-script` container would not have access to globals defined by other scripts in the outer window unless they were imported using `imported-globals`.
+
+- Globals defined in `exported-globals` will be redefined in any `virtual-script` container.
+- Exported globals will also be available to any Locker-created sandboxes
+- Globals defined in Oasis using `imported-globals` will be redefined only in the `virtual-script` container they are imported to.
+
+Note: these global values are controlled side-channels between the main window and the sandbox, and should be used with caution. For example, creating a component that depends on `ga` to be defined, is an anti-pattern.
+
+### Supported Attributes
```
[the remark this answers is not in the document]
tedconn: I think `load` and `error` we will definitely support but beyond that, the answer is "I don't know". I will update the RFC with support for `load` and `error` events but we may need to expand on that in the future as we uncover more use cases.

jbenallen (approved):

matt-buland-sfdc (approved):

text/0113-virtual-script-container.md
caridy: yeah, has described above, see @jbenallen question: this is not focus on security, only on the underlaying technology, who can use it in platform is a different question, certainly another RFC.

text/0113-virtual-script-container.md
caridy: maybe we can invert the statement... saying that with this tool, we are targeting non-critical path tools, and for that reason, we are going to force them to use `async` attribute, to fetch and evaluate when possible.

text/0113-virtual-script-container.md
caridy: I was thinking about this as well... there are other issues. lets chat about it.

text/0113-virtual-script-container.md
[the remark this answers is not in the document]
caridy: @tedconn remember that we now have two equivalent/reflective names: `exported-global-names` and `imported-global-names`.

@matt-buland-sfdc, they are analogous to import and exports in modules, the imported is readOnly inside the sandbox, and the exported is readOnly by the main app. 

@jbenallen in principle, the script evaluated inside virtual script is a high-privilege script, no restrictions on what it can do, therefore, it can attempt to do, and can do damage on the main window (e.g.: full access to cookies, etc). This is not a secure sandbox. neither imported and exported names are live values, they are defined by the person who insert the virtual script in the main window, and cannot be modified causing expansion of their initial value from within the sandbox itself, so, in that regards, it is all fine.

text/0113-virtual-script-container.md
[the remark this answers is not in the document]
caridy: I don't think this is correct. virtual-script does support modules just fine. one thing to highlight though is that since the sandbox has its own module graph, no matter if you load the same module or same script in the main window (raw), the one evaluated by the sandbox is a new instance of it, it is not shared in any way with the main window, other than being cached by the engine.

text/0113-virtual-script-container.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Unsupported Attributes
+
+`type`: all `virtual-script` resources will be `text/javascript`. `type="module"` will not be supported.
+
+`charset`: this attribute is deprecated
+
+`language`: this attribute is deprecated
+
+`nonce`: has special semantics since it is a global attribute and browsers do special behavior with this attribute when a CSP policy is applied. We will have to treat this attribute specially.
```
caridy, line 181: lets chat about this one

tedconn: I put some more information in this issue here: https://github.com/salesforce/oasis/issues/8

caridy (approved):

PR #47, opened 2021-03-31
feat: HTML comments for LWC templates

jodarove:
RFC to support HTML comments in LWC templates.

### DIFF
--- a/text/0000-template-comments.md
+++ b/text/0000-template-comments.md
@@ -1,0 +1,86 @@
+---
+title: HTML comments for LWC templates
+status: DRAFTED
+created_at: 2021-03-31
+updated_at: 2021-03-31
+pr: https://github.com/salesforce/lwc/pull/2263
+---
+
+# HTML comments for LWC templates
+
+## Summary
+
+In most cases, HTML comments in templates only add information for other developers. There are some use cases where HTML comments convey some runtime semantic (e.g., MSO comments). As of today, the LWC template compiler strips HTML comments. This RFC proposes a way to preserve HTML comments in LWC templates.
+
+### Basic example
+
+```html
+<template preserve-comments>
+  <!-- Greeter container -->
+  <div>
+	Hello <b>{name}</b>!
+  </div>
+  <!-- eof: Greeter container -->
+</template>
+```
+```html
+<x-example>
+  # shadow-root
+  |   <!-- Greeter container -->
+  |   <div>
+  |     Hello <b>world</b>
+  |   </div>
+  |   <!-- eof: Greeter container -->
+</x-example>
+```
+
+## Motivation
+
+Microsoft Outlook's desktop email client leverage MSO (Microsoft Office) to render their HTML content. Because of this, the support for common HTML and CSS properties is limited, eg: `<div>`s are not supported.
+
+Email developers rely on [MSO conditionals](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/ms537512(v=vs.85)?redirectedfrom=MSDN)  to improve the UX of emails opened on Microsoft Outlook.
+
+This RFC introduces a new opt-in mechanism at compile time to preserve comments on the rendered component.
+
+## Proposal
+
+### Compile-time behavior
+
+LWC compilation strips HTML comments because the browsers do not use them at runtime. The bundle's size is smaller by stripping the HTML comments, and the rehydration process does not need to take into account comments nodes.
+
+To preserve the performance of the existing components, we propose to introduce one of these:
+
+1. A new attribute (`preserve-comments`) to the root template tag in the component template.
+
+2. Add a compile option (`preserveHTMLComments`) to the template compiler.
+
+Both options will make the compiler invoke a runtime API adding the HTML comments in the component template to the output code that runs in the browser/server. Modifications in each environment's renderers will ensure that the comment node/text is present on the resulting HTML.
+
+### Runtime behavior
+
+A new API "`[co]`mment" will be added to the engine/core. The `co(string commentText): VNode` will receive the comment text and return a Comment vnode.
+
+During rendering cycles, the LWC engine diffing algorithm will show/hide the comment based on the directive containing it (if:true/false, for:each, etc.)
+
+#### Invariants
+
+* HTML comments will be supported based on the opt-in mechanism (TBD, either option 1 or 2).
+* HTML comments will not support bindings (`{foo}`).
+* When supported, comments must be rendered in engine/dom and engine/server.
+* HTML comments inside <slot> elements, will be rendered as part of the slot content.
+* When inside a `<template if:true/false>` will be rendered according to the `if:true/false` directive.
+* HTML comments inside `<template for:each>` will be rendered according to the `for:each` directive.
+
+### Backwards Compatible
+
+This feature is backward compatible because is gated by the new opt-in mechanism.
+
+### Prior art
+
+* React: [dangerouslySetInnerHTML attribute](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml).
+* VueJS: [comments option](https://vuejs.org/v2/api/#comments).
+* Svelte: [preserveComments compile option](https://svelte.dev/docs#svelte_compile)
+
+### How we teach this
+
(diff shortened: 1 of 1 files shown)


text/0000-template-comments.md
```diff
@@ -0,0 +1,86 @@
...
+### Runtime behavior
+
+A new API "`[co]`mment" will be added to the engine/core. The `co(string commentText): VNode` will receive the comment text and return a Comment vnode.
+
+During rendering cycles, the LWC engine diffing algorithm will show/hide the comment based on the directive containing it (if:true/false, for:each, etc.)
+
+#### Invariants
+
+* HTML comments will be supported based on the opt-in mechanism (TBD, either option 1 or 2).
+* HTML comments will not support bindings (`{foo}`).
```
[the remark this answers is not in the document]
pmdartus: In the example above the binding is no on the HTML comment itself. There are 2 HTML comment nodes surrounding a `<i>` element with a `style` attribute binding. Here is as an example of an HTML comment binding:

```html
<template>
  <!-- With dynamic content: {content} -->
</template>
```

The `{content}` will not be interpreted as an expression but as a static text content.

text/0000-template-comments.md
```diff
@@ -0,0 +1,86 @@
...
+
+This RFC introduces a new opt-in mechanism at compile time to preserve comments on the rendered component.
+
+## Proposal
+
+### Compile-time behavior
+
+LWC compilation strips HTML comments because the browsers do not use them at runtime. The bundle's size is smaller by stripping the HTML comments, and the rehydration process does not need to take into account comments nodes.
+
+To preserve the performance of the existing components, we propose to introduce one of these:
```
pmdartus, line 51: What about both options?

The main advantage of the attribute is that it allows platform developers to benefit from this feature since they don't have access to the LWC compiler config. When you have access to the compiler config (eg. OSS usage) a single flag is also convinient.

text/0000-template-comments.md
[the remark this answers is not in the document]
gram25: Am not sure inner html binding would work in this case since we are not injecting html content here. Rather the binding is to set the attribute value of the <i> tag in this example. The actual value thats being rendered today for the binding in the above example is 
`<i style="mso-font-width:-100%;letter-spacing:20px;mso-text-raise:-10px;..>`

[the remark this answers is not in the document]
pmdartus: Here is an example of how declarative inner HTML binding could be used to injected dynamic HTML comments:

```html
<template>
  <div lwc:inner-html={msoComment}></div>
</template>
```

```js

  get msoComment() {
    const iLeftStyle = '... some computed content ...';
    return `<!--[if mso]><i style=${iLeftStyle}>&nbsp;</i><![endif]-->`;
  }
}
```

This main constrain with declarative inner HTML binding is that it is bound to an element and can't be injected directly inside the template body. As far as I understand how MSO comment works, it should be fine to inject it in a `<div>`.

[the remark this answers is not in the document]
gram25: Got it thanks for the example. I see that the inner html RFC is also in draft state. Would it be implemented by then ?

text/0000-template-comments.md
jodarove: I like it, adding it.

text/0000-template-comments.md
```diff
@@ -0,0 +1,86 @@
...
+
+A new API "`[co]`mment" will be added to the engine/core. The `co(string commentText): VNode` will receive the comment text and return a Comment vnode.
+
+During rendering cycles, the LWC engine diffing algorithm will show/hide the comment based on the directive containing it (if:true/false, for:each, etc.)
+
+#### Invariants
+
+* HTML comments will be supported based on the opt-in mechanism (TBD, either option 1 or 2).
+* HTML comments will not support bindings (`{foo}`).
+* When supported, comments must be rendered in engine/dom and engine/server.
```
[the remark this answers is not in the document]
jodarove: resolved

text/0000-template-comments.md
```diff
@@ -0,0 +1,86 @@
...
+* HTML comments inside <slot> elements, will be rendered as part of the slot content.
+* When inside a `<template if:true/false>` will be rendered according to the `if:true/false` directive.
+* HTML comments inside `<template for:each>` will be rendered according to the `for:each` directive.
```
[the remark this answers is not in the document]
jodarove: i joined these 3 invariants into _HTML comments will behave the same way as other nodes with respect to slots and directives (`if:true`, `for:each`, etc.)._ I could still remove it if you want.

text/0000-template-comments.md
```diff
@@ -0,0 +1,90 @@
...
+# HTML comments for LWC templates
+
+## Summary
+
+In most cases, HTML comments in templates only add information for other developers. There are some use cases where HTML comments convey some runtime semantic (e.g., MSO comments). As of today, the LWC template compiler strips HTML comments. This RFC proposes a way to preserve HTML comments in LWC templates.
+
+### Basic example
+
+```html
+<template preserve-comments>
```
nolanlawson, line 18: Would it make sense to namespace this, just to make it clear this is an LWC-ism? (E.g. `<template lwc:preserve-comments>`.) `<template>` is being extended with things like declarative shadow DOM already (`<template shadowroot="open">`).

jodarove: ihmo, no. `preserve-comments` only applies to the root template tag, which is owned by LWC. we usually namespace with `lwc` to attributes that apply to regular elements, like `lwc:dom`.

Note: I'm not aware of `<template shadowroot="open">`, but ill assume that is meant to be used by lwc; in that case, I would say: if the same root template tag has `preserve-comments`, it should also render the comments nodes.

nolanlawson: Sorry for not being clear: `<template shadowroot="open">` comes from [declarative shadow DOM](https://web.dev/declarative-shadow-dom/). I was just pointing out that the browser is adding meaningful attributes to the `<template>` tag.

If LWC owns the root `<template>` then I agree it's not a big deal not to namespace it.

pmdartus: I don't have a strong opinion on this, however, I like the fact that we attempt to scope all the LWC-ism with the `lwc:` prefix. This defines a clear boundary between standard and LWC specific syntax from the developer's perspective.

I am fine either way.

text/0000-template-comments.md
```diff
@@ -0,0 +1,90 @@
...
+### Runtime behavior
+
+A new API "`[co]`mment" will be added to the engine/core. The `co(string commentText): VNode` will receive the comment text and return a Comment vnode.
+
+During rendering cycles, the LWC engine diffing algorithm will show/hide the comment based on the directive containing it (if:true/false, for:each, etc.)
+
+#### Invariants
+
+* HTML comments will be enabled based on the opt-in mechanism, either [option 1 or 2](#compile-time-behavior).
+* HTML comments will not support bindings (`{foo}`).
```
pmdartus, line 74: It would be great to add some details here, stating that if the comment content is dynamic it can be rendered using the `lwc:inner-html` binding (with a link to the proposal).

text/0000-template-comments.md
nolanlawson: Are there any cases where de _don't_ scope the attribute for LWC-isms with `lwc:`? The only ones I can think of are `for:each` and `for:item`, but those are already "scoped" with the `for`, which makes them look very non-web-standards-y since there are few standard cases that web developers will run into like that (`xmlns:xlink` on SVGs is about the only common one I can think of).

jodarove: thanks for sharing that interesting read @nolanlawson! With that, I'm in the same position as @pmdartus, I don't have a strong opinion. However, I think is safer that we go with the `lwc:`. I will update this RFC accordingly 👍 .

> Are there any cases where de don't scope the attribute for LWC-isms with `lwc:`?
on top of those: `if:true|false`, `key`, `iterator:it`?.

pmdartus, line 18:
```suggestion
<template lwc:preserve-comments>
```

pmdartus (approved):
You need to update the example at the top to reflect the attribute name change. Otherwise, this looks great!

PR #48, opened 2021-04-12
adding dynamic component configuration rfc

alejandro-sf:


alejandro-sf:
@jodarove @pmdartus @diervo please consider this RFC for dynamic component configuration (lwc:bind)

nolanlawson:
@alejandro-sf It looks like something is wrong with the formatting? GitHub is not able to render the markdown correctly: [rendered view](https://github.com/salesforce/lwc-rfcs/blob/1a1f0529fabffd73675d692e78e477e431ad731f/text/0114-dynamic-component-configuration.md)

alejandro-sf:
Oh I didn't know the markdown was visible even from the review...I'll update to fix I used the quip export to markdown feature so there's probably something wrong there

alejandro-sf:
Please see the new formatting [here](https://github.com/salesforce/lwc-rfcs/blob/e92dc2b865410120bfd647adabd0aa4b3a342f5b/text/0114-dynamic-component-configuration.md)

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+}
+
+get attributes() { return this.props; }
+
+```
+A new tool must be introduced that will allow component authors to configure their components with information retrieved at runtime without relying on maintaining element references or manually manipulating the DOM.
+
+## Proposal
+
+The proposal involves adding a new directive that captures the relevant information to configure a web component at runtime. The assumption is that this directive will be used in cases where the component module is unknown at runtime (therefore used alongside dynamic imports) and zero or more of the following are unknown at runtime: properties, attributes, or event handlers.
```
pmdartus, line 191:
```suggestion
The proposal involves adding a new directive that captures the relevant information to configure a web component at runtime. The assumption is that this directive will be used in cases where the component constructor is unknown until runtime and zero or more of the following are unknown until runtime: properties, attributes, or event handlers.
```

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+
+The directive `lwc:bind` can retrieve component configuration data via a JavaScript object with a pre-defined shape. Constructor is the only required property in the configuration data.
+
+## General Invariants
```
pmdartus, line 195: There is a mix of actual invariants and design details in this section. 

For example, `Changes to the configuration object must be reflected in the component` is not an invariant of this proposal but rather a design decision based on the proposed solution. I would recommend breaking this section apart and specifying each individual behavior.

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+
+* It must not break the reactivity model or the general framework lifecycle
+
+* After the component is loaded, it should function within the lifecycle of the parent component
+
+* Changes to the configuration object must be reflected in the component (ie changing the properties will cause component properties to be set again)
+
+* Changing the constructor on a parent component will also trigger a cascade re-render of any child components
```
pmdartus, line 203: Changing a constructor requires unmounting the existing custom element and creating a new custom element. It actually doesn't re-render the children elements but creates brand new children elements.

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Drawbacks
+
+Identifying the dependencies for static analysis could be difficult especially if the configurations are not derived from @api properties or statically analyzable strings. Additionally, this proposal would enable true dynamic component creation which has been abused in the past - proper guardrails would need to be implemented to prevent improper use of this pattern.
+
+## Adoption Strategy
+
+This feature will be available in Salesforce Lightning Platform and LWR to certain teams depending on their use cases
+
+## Unresolved Questions
```
pmdartus, line 223: Questions that are not covered by this proposal:

- Can the `lwc:bind` directive be used with other directives, attributes and event handlers?
- If the response to the previous answer is yes, what happens when the `lwc:bind` contains an attribute that is already defined via the template.  For example: `<x-foo lwc:bind={config} class="error"></x-foo>` with `config = { attributes: { class: 'success' } }` ?
- In which order `properties` and `attributes` are set? If the config is defined as `config = { attributes: { class: 'error' }, properties: { className: 'success' }  }` what is the expected output?
- How are attributes values treated? How to add and remove an attribute? What is the semantic of the following attribute values: `null`, `undefined`, `true` and `false`?

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+Object.assign(this.element, this.props);
+
+}
+
+get attributes() { return this.props; }
+
+```
+A new tool must be introduced that will allow component authors to configure their components with information retrieved at runtime without relying on maintaining element references or manually manipulating the DOM.
+
+## Proposal
```
pmdartus, line 189: It would be great to add a section about how other frameworks are approaching this issue.

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+this.loadCtor();
+
+async loadCtor() {
+	const ctor = await import(this.moduleName);
+	this.customCtor = ctor.default;
+}
+
+get configurations() {
+	return {
+		ctor: this.customCtor,
```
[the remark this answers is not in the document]
Josh-Goodman: Does that mean you want 'ctor' to be renamed as 'is' or is there another mechanism for this?

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+
+```html
+<template>
+   <template for:each={chartItems} for:item="chartItem">
+      <c-lazy-cmp lwc:bind={chartItem}></c-lazy-cmp>
+   </template>
+</template>
+```
+```js
+import { LightningElement, api } from "lwc";
```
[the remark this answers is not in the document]
Josh-Goodman: Should we pare this down? I wanted to make sure we had a compelling usecase for each of the 4 things you can configure in 1 place.

text/0114-dynamic-component-configuration.md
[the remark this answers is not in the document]
pmdartus: It's not the `lwc:is` but rather the `lwc:dynamic` that I was referring to above: https://github.com/salesforce/lwc-rfcs/blob/master/text/0110-dynamic-components.md#api

text/0114-dynamic-component-configuration.md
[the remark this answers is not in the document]
pmdartus: I would keep the example in the RFC itself as simple a possible. You can add more complex examples at the end in an appendix for example this way is doesn't distract from the actual content of the RFC.

text/0114-dynamic-component-configuration.md
jodarove: what about special properties, like `innerHtml`, `outerHtml`, `innerText`, etc? Will they be allowed?

There's already a [inner html binding RFC] (https://github.com/salesforce/lwc-rfcs/pull/15), which states some security concerns. Open question: Does this `lwc:bind` proposal overlaps with `lwc:inner-html`?

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,300 @@
...
+export default class LazyCmp extends LightningElement {
+	@api dashboardId;
+	@api dashboardData;
+
+	get chartItem() {
+		return this.chartItem;
+	}
+
+	@wire(myAdapter, { id: 'someDashboardId' }) {
+		chartDefinitions({ data, error }) {
```
jodarove, line 69:
```suggestion
	@wire(myAdapter, { id: 'someDashboardId' }) 
	chartDefinitions({ data, error }) {
```

text/0114-dynamic-component-configuration.md
```diff
@@ -0,0 +1,300 @@
...
+Also, when connetedCallback called on child and the component has been inserted into DOM, but its properties/attributes/eventHandlers have not been initiated, we also need to wait for another cycle to get them rendered with the correct properties. This breaks the framework's life cycle.
+
+```js
+export default class LazyCmp extends LightningElement {
+	element;
+
+	@api moduleName;
+
+	connectedCallback() {
+		const ctor = await import(this.moduleName);
```
jodarove, line 116: invalid js

alejandro-sf: Is this because await must be in an async function?

PR #49, opened 2021-05-11
rfc: mixed shadow mode

ekashida:


text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,175 @@
...
+    </x-child>
+</template>
+
+<!-- x-child -->
+<template>
+    <p>child</p>
+</template>
+```
+
+This observable difference will not be remediated as such a change would be non-trivial and
```
nolanlawson, line 122: Another difference worth mentioning is timing, in the case where the slotted contents are ordered differently from the slots as defined in the slottable component. (See [timing comparison](https://github.com/salesforce/lwc-rfcs/blob/a3f6d7182de687554ef175b1228b0c9e06271619/text/0000-light-dom.md#slots))

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,175 @@
...
+backwards-incompatible with the current synthetic shadow implementation.
+
+### this.shadowRoot vs this.template
+
+`this.shadowRoot` references the component's shadow root in native Shadow DOM while `this.template`
+references the component's shadow root in synthetic Shadow DOM. This observable difference can be
+remediated for components that favor native Shadow DOM by aliasing `this.template` to
+`this.shadowRoot`. There are no plans to enable `this.shadowRoot` for existing components as the
+current LWC-ism is to use `this.template`.
```
nolanlawson, line 131: This section is a bit unclear to me. Are we planning to make `this.template` evaluate to `null` in native shadow mode?

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,175 @@
...
+either the polyfilled API or the native browser API. If the browser does not support Shadow DOM
+(e.g., IE11), then LWC will fallback to synthetic mode regardless of which mode the component
+prefers.
+
+It is likely that components will need to know which mode it is currently operating in, due to
+observable differences between the two. Any indication of this by the framework should only be
+available to components that set the `preferNativeShadow` property to `true`.
+
+In terms of composition, a synthetic mode component can contain a native mode component, but the
+inverse is not allowed. Not only does this make things easier to reason about, but many existing
```
pmdartus, line 57: What is the expected behavior when a parent component opt-in to native shadow while a child component is using synthetic shadow DOM? 

I originally thought that LWC should throw an error. This will cause issues when the flexipage component itself will be migrated to native shadow DOM since we can't ensure that all the components composing the flexipage are using native shadow as well. That said, migrating the flexipage component will certainly cause issues with Aura components as well.

Also I am wondering what would how this restriction would work when mixing light DOM components with shadow DOM components (native or not)

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,175 @@
...
+Applications that rely on instrumentation libraries that don't yet support Shadow DOM are currently
+able to obtain references to descriptors like `addEventListener()` before they are patched and use
+them to override `@lwc/synthetic-shadow` polyfills where needed. Such workarounds would not work for
+events originating from components that prefer native Shadow DOM.
+
+### Styling
+
+Applications currently rely on global stylesheets to apply deeply throughout the DOM. With native
+Shadow DOM, this would not be possible. All of a component's styles would have to be generated and
+added to the component bundle.
```
pmdartus, line 78: These sections should probably be moved to "Observable differences".

text/0000-mixed-shadow-mode.md
pmdartus: >  while `this.template` references the component's shadow root in synthetic Shadow DOM.

This isn't entirely true, when `@lwc/synthetic-shadow` isn't loaded `this.template` refers to the native shadow root.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,175 @@
...
+component's perspective, there are two choices of shadow semantics: synthetic mode or native
+mode. The concept of a mixed mode should be transparent and the framework should guarantee this
+through integration testing, which is further discussed below.
+
+A component can signal that it prefers native Shadow DOM by setting a static property called
+`preferNativeShadow` to `true`.
+
+```js
+export default class extends LightningElement {
+    static preferNativeShadow = true;
```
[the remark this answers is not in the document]
ekashida: > IMO, the engine should warn and fall back to native shadow in this case

There's nothing the component author can do if `@lwc/synthetic-shadow` is not loaded by the application and I would like to avoid adding unactionable warnings to the logs. Since `false` is the default value and the component is expected to handle both native and synthetic modes (note the intentional use of the term "prefer"), do you think it helps to add a warning during runtime?

text/0000-mixed-shadow-mode.md
ekashida: Fixed, thanks.

text/0000-mixed-shadow-mode.md
ekashida: Great point, thanks!

text/0000-mixed-shadow-mode.md
ekashida: > What is the expected behavior when a parent component opt-in to native shadow while a child component is using synthetic shadow DOM?

The current plan is to throw an exception during runtime.

> I originally thought that LWC should throw an error. This will cause issues when the flexipage component itself will be migrated to native shadow DOM since we can't ensure that all the components composing the flexipage are using native shadow as well. That said, migrating the flexipage component will certainly cause issues with Aura components as well.

If we do change our stance on this, it would be relaxing a restriction so I think we can deal with it once we have to.

> Also I am wondering what would how this restriction would work when mixing light DOM components with shadow DOM components (native or not)

Good question, I'll add it to the list of unresolved questions and think about it some more.

nolanlawson: FWIW yesterday we discussed light children inside of synthetic shadow parents and decided we wouldn't try to support style scoping for this. (I.e. `<style>`s will be inserted into the light child as-is; no effort is made to scope them to their shadow parent.) Not sure about other implications, but probably light DOM will work best with native shadow.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Detailed design
+
+Components will opt-in to this new feature and the default mode will be synthetic mode. From the
+component's perspective, there are two choices of shadow semantics: synthetic mode or native
+mode. The concept of a mixed mode should be transparent and the framework should guarantee this
+through integration testing, which is further discussed below.
+
+A component can signal that it prefers native Shadow DOM by setting a static property called
+`preferNativeShadow` to `true`.
```
nolanlawson, line 39: I'd just like to point out, since it came up in the light DOM discussion, that we are currently defaulting this to `false` but someday (presumably) it will default to `true`. So we can't really avoid the "booleans should default to `false`" problem here.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,196 @@
...
+Another difference is that for synthetic mode, lifecycle hooks are invoked in the order of
+appearance after they are assigned, whereas in native mode, they are invoked in the order of
+appearance in the template.
+
+### this.shadowRoot vs this.template
+
+`this.shadowRoot` and `this.template` reference the component's shadow root in native Shadow DOM
+while `this.template` references the component's shadow root in synthetic Shadow DOM. There are no
+plans to enable `this.shadowRoot` for existing components that only support synthetic Shadow DOM as
+the current LWC-ism is to use `this.template`.
```
nolanlawson, line 130: Does this mean that, moving forward, we will recommend `this.shadowRoot` over `this.template`? (I think this may actually be a good idea, since one is standard and the other is an LWC-ism.)

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,196 @@
...
+Applications that rely on instrumentation libraries that don't yet support Shadow DOM are currently
+able to obtain references to descriptors like `addEventListener()` before they are patched and use
+them to override `@lwc/synthetic-shadow` polyfills where needed. Such workarounds would not work for
+events originating from components that prefer native Shadow DOM.
+
+### Styling
+
+Applications currently rely on global stylesheets to apply deeply throughout the DOM. With native
+Shadow DOM, this would not be possible. All of a component's styles would have to be generated and
+added to the component bundle.
```
nolanlawson, line 162: Worth mentioning here is that native shadow components will not require `lwc:dom="manual"` (as I understand it), since we won't need a MutationObserver to add the shadow token attributes for style scoping.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,196 @@
...
+Note that the current proposal for mixed mode considers polyfills as all-or-nothing.  LWC is not in
+the business of providing workarounds for discrepancies in browser implementations, unless it
+affects the functionality of the framework itself.
+
+# How we teach this
+
+Developers should know about mixed mode and the ability for a component to favor native Shadow DOM
+before they start implementing new components. This would eliminate any potential tech debt before
+they get started. Any unresolved observable differences should be documented to assist in deciding
+whether a component is a candidate for native Shadow DOM.
```
nolanlawson, line 191: Thinking about the developer experience, I wonder if it would be useful to offer customers a way to enable native shadow DOM mode globally for all their components. You can imagine that some shop may eagerly migrate all their components to use `preferNativeShadow = true`, but now they can't ensure that any _new_ components will be written in native shadow. And for developers, it's easy to forget to add the static property. Whereas if they could control the default value, then they could ensure that they're futureproof before we switch the framework default from `false` to `true`.

That said, I'm not sure technically how to achieve this, or how to scope the flag properly.

text/0000-mixed-shadow-mode.md
ekashida: I think you mean to say that many components will set this property to `true` but I don't think we can ever default it to `true`...at least without versioning. There are thousands of LWC components in existence today and we can't expect all of them to support mixed mode. My guess is that applications that are ready for a change like defaulting this to `true` will simply stop including the polyfill.

text/0000-mixed-shadow-mode.md
ekashida: Actually, this section states that we will not enable `this.shadowRoot` for components unless they prefer native mode.

text/0000-mixed-shadow-mode.md
ekashida: I think that in the case you described, they could just remove the synthetic shadow polyfill.

text/0000-mixed-shadow-mode.md
nolanlawson: That's a good point. I guess I mean that, at some point in the future, LEX will be 100% native shadow DOM, correct? Which means that effectively the default will be `true` from the component author's perspective?

text/0000-mixed-shadow-mode.md
nolanlawson: I'm thinking of this from the perspective of, say, a managed package author. If someone is off-core and owns the whole stack, then yeah, they can just remove the polyfill.

text/0000-mixed-shadow-mode.md
nolanlawson: Gotcha. In that case, backporting `this.shadowRoot` to synthetic (in which case it just refers to `this.template`) would seem sensible to me, since it's one less thing for developers to have to think about when converting components between native/synthetic shadow.

text/0000-mixed-shadow-mode.md
ekashida: Yeah, I see what you mean. I guess that open source applications would just exclude the synthetic shadow polyfill, but we would probably need to do something different for platform as the majority of components start using native. We could probably do something smart in the platform compiler.

text/0000-mixed-shadow-mode.md
ekashida: I would prefer to wait and see if this is an actual pain point before backporting this feature. I suspect it won't be because conversions would just be from synthetic to native and `this.template` would work in both modes. If we do decide to make the change later, it would be a simple and straightforward backwards-compatible change, so we probably won't need to reach consensus to move forward. I will add this as an open question to get some more feedback.

text/0000-mixed-shadow-mode.md
ekashida: This is a great point. I think we will need to keep the restriction of using `lwc:dom="manual"` even for native shadow, if components need to support both native and synthetic. We'll probably also need to list out the observable differences when manipulating DOM via DOM APIs. Thoughts?

text/0000-mixed-shadow-mode.md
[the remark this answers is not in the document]
pmdartus: You are right. Now that I am thinking about this again it will create too much noise.

text/0000-mixed-shadow-mode.md
pmdartus: > I think we will need to keep the restrictions

Are you speaking about the [following restrictions](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/restrictions.ts#L102-L158)?

> We'll probably also need to list out the observable differences when manipulating DOM via DOM APIs.

The only difference between synthetic shadow + `lwc:dom=manual` and native shadow regarding styling is the timing when styles are applied. In synthetic shadow, we are using a mutation observer to track sub-tree changes to add the styling attribute. It means that in synthetic shadow DOM, styles are applied to elements a microtask later than with native shadow DOM. 

The only way I see this to break is if developers are querying for styles or layout using `Window.getComputedStyle` or `Element.prototype.getBoundingClientRect` right after they mutate the DOM but right before the mutation observer is notified. Which is quite unlikely IMO. I don't think it worth adding a section in the documentation discussion this.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,196 @@
...
+
+A potential (non-verified) drawback is that existing components may experience a slight performance
+degradation due to the fact that mixed mode requires logic-forking whenever a Shadow DOM API is
+invoked.
+
+## Alternatives
+
+Other alternatives have not been considered but contributions to this proposal are welcome.
+
+## Adoption strategy
```
pmdartus, line 174: To make the adoption of mixed-mode successful we will need to provide developers a way to test their components in both modes. 

As of today synthetic shadow is [always loaded](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/src/setup.js#L1) in `@lwc/jest-preset`. It would be great if we were to have a way to conditionally load synthetic shadow (https://github.com/salesforce/lwc-test/issues/93). We would then recommend running the same tests with and without synthetic shadow to make sure the component works as expected in both modes.

text/0000-mixed-shadow-mode.md
pmdartus: An alternative solution would be to introduce a new `prefer-native-shadow` linting rule. This linting rule could check if all the classes extending from `LightningElement` have a `static preferNativeShadow = true`.

text/0000-mixed-shadow-mode.md
ekashida: > > I think we will need to keep the restrictions

> Are you speaking about the following restrictions?

@pmdartus Yes

> It would be nice if we could remove the lwc:dom=manual requirement for native-shadow components (it's a nice carrot to encourage migration to native shadow), but I understand if we need to keep it for now because of the way the code is structured.

The problem is that without that directive, the component would work fine in native mode but throw exceptions in synthetic mode.

nolanlawson: Point taken. To be fair though, isn't that also the case with calling `this.shadowRoot`? (Works fine in native mode, fails in synthetic mode.)

ekashida: Thanks, added the addition of `this.shadowRoot` to synthetic mode as an open question.

text/0000-mixed-shadow-mode.md
ekashida: Thinking about this more, I think my original intention was to expose `this.shadowRoot` in synthetic mode, iff `preferNativeShadow` is set to `true`. This makes sense to me so I'll make it more explicit in this section:

> There are no plans to enable `this.shadowRoot` for existing components that **_only_** support synthetic Shadow DOM as the current LWC-ism is to use `this.template`.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+of these polyfills can result in broken experiences due to dependencies on certain "escape hatches"
+that are not possible in a native Shadow DOM environment (e.g., instrumentation, styling, etc, which
+are discussed below).
+
+Applications that wish to use LWC components in a native shadow context simply need to omit the
+inclusion of the `@lwc/synthetic-shadow` polyfill; however, this is not a realistic option for
+large applications that cannot afford a rewrite.
+
+This proposal opens up an incremental migration path for applications moving to native Shadow DOM by
+allowing the usage of both native and synthetic Shadow DOM in the same document.
```
caridy, line 29: `... same document` is probably the wrong term, maybe `...same application.`

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Detailed design
+
+Components will opt-in to this new feature and the default mode will be synthetic mode. From the
```
caridy, line 33: this first sentence is confusing in the sense that it looks like the component is choosing synthatic mode by default, but that's only true if the app chooses to load synthetic shadow, which platform always does.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+component's perspective, there are two choices of shadow semantics: synthetic mode or native
+mode. The concept of a mixed mode should be transparent and the framework should guarantee this
+through integration testing, which is further discussed below.
+
+A component can signal that it prefers native Shadow DOM by setting a static property called
+`preferNativeShadow` to `true`.
+
+```js
+export default class extends LightningElement {
+    static preferNativeShadow = true;
```
caridy, line 43: what an ugly property name :)

caridy: maybe a enum: `static defaultShadowMode = 'native'`

caridy: btw, I realized it is probably too late for me to bikeshed on this... but I can't help myself! lol

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+}
+```
+
+If this property is not defined, a default value of `false` will be used. The value of this property
```
caridy, line 47: No, it can't be at the construction time... this should be cached before the first instance of this constructor is created, and cached at the def level for the future.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+will be read once during component construction and cached for the lifetime of the component. The
+LWC engine will throw an exception if the value of `preferNativeShadow` is not a boolean value.
+
+All polyfills applied by `@lwc/synthetic-shadow` will look to this property as a signal to invoke
```
caridy, line 51: this sentence if also confusing because synthatic-shadow has no access to the component constructor by design, so it can't really inspect that value, it is the responsibility of LWC to signal what shadow to attach based on the information associated to the instance/constructor.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+either the polyfilled API or the native browser API. If the browser does not support Shadow DOM
+(e.g., IE11), then LWC will fallback to synthetic mode regardless of which mode the component
+prefers.
+
+Due to observable differences between the two modes, it is likely that components that prefer native
```
caridy, line 56: what will be such hint?

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+will need to know whether it is currently operating in native or synthetic Shadow DOM. Any hints
+provided by the framework will only be available to components that set the `preferNativeShadow`
+property to `true` because this information is not useful to components that only support synthetic
+Shadow DOM.
+
+In terms of composition, a synthetic mode component can contain a native mode component, but the
```
caridy, line 62: I love this paragraph!

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+As mentioned previously, in synthetic mode, slotted elements that are never assigned to a slot are
+not rendered. This means that their lifecycle hooks are never invoked.
+
+Another difference is that for synthetic mode, lifecycle hooks are invoked in the order of
+appearance after they are assigned, whereas in native mode, they are invoked in the order of
+appearance in the template.
+
+### this.shadowRoot vs this.template
+
+`this.shadowRoot` and `this.template` will both reference the component's shadow root in native
```
caridy, line 127: I disagree with this as a signal because this will put in danger any attempt to merge component and elements into a single instance in the future because the getter for shadowRoot is subject to the mode used for the shadow, which is orthogonal to this proposal.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+Shadow DOM while `this.template` currently references the component's shadow root in synthetic
+Shadow DOM.  When a component specifies a preference for native Shadow DOM, `this.shadowRoot` will
+be made available as an alias to `this.template`.
+
+There are no plans to enable `this.shadowRoot` for existing components that only support synthetic
+Shadow DOM as the current LWC-ism is to use `this.template`.
+
+### this.shadowRoot instanceof ShadowRoot
+
+Components that favor native Shadow DOM will observe that `this.shadowRoot instanceof ShadowRoot`
```
caridy, line 137: If this will be remediated, which it should, why are we arguing that this statement is false?

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,206 @@
...
+### Listening for non-composed events above the root LWC node
+
+There currently exists logic that allows listeners to handle non-composed events outside of the root
+LWC node if the event originates from a non-LWC component in the subtree. This behavior exists for
+legacy reasons, is only possible to allow in a synthetic Shadow DOM environment, and cannot be
+preserved in a native Shadow DOM context.
+
+### Accessibility
+
+The most common accessibility issue in LWC is related to id-referencing across shadow boundaries.
```
caridy, line 151: I'm not sure I understand this sentence.

text/0000-mixed-shadow-mode.md
ekashida: I like it! I don't know anyone who has a strong preference for `preferNativeShadow`. I'll bring it up with the rest of the team and make the changes if nobody objects.

text/0000-mixed-shadow-mode.md
ekashida: How about making it available as a boolean value on `this.template.native`? This would resolve to the falsy value `undefined` for existing components, which works well in terms of backcompat.

text/0000-mixed-shadow-mode.md
ekashida: I'm not sure I understand. Are you saying that we should avoid exposing `this.shadowRoot` because this is something we will never do for LWC components?

text/0000-mixed-shadow-mode.md
pmdartus: I like it as well.

text/0000-mixed-shadow-mode.md
pmdartus: We should align this with static `renderMode` as well. Both values are currently read at construction time.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Drawbacks
+
+A potential (non-verified) drawback is that existing components may experience a slight performance
+degradation due to the fact that mixed mode requires logic-forking whenever a Shadow DOM API is
+invoked.
+
+## Alternatives
+
+Other alternatives have not been considered but contributions to this proposal are welcome.
```
nolanlawson, line 169: I think it would be worth spelling out the (somewhat naïve) alternative: just switching from 100% synthetic to 100% native. Here's why I don't think this is an obviously "bad" idea:

- Many folks will wait until the last second to upgrade their components anyway (given human nature), so for them, their experience will basically be the same as switching from 100% synthetic to 100% native.
- We will have to add code to handle mixed mode, and this code will need to be thrown away when we go 100% native.
- There are complexities with supporting mixed mode (native within synthetic, event bubbling and retargeting, styling, etc.), and this could potentially add its own maintenance burden, e.g. if customers start depending on some quirk of native-within-synthetic.

I'm not saying this is what we _should_ be doing, but I think it's a reasonable alternative, so we should have an explanation of why we didn't go this route.

text/0000-mixed-shadow-mode.md
caridy: construction time is the wrong time to compute this, it should be extract the first time LWC inspect the class for the first time.

text/0000-mixed-shadow-mode.md
caridy: No, I'm saying that if we ever merge this instance with the element instance, it will be the element instance the one deciding whether or not that is null based on the attach shadow mode.

text/0000-mixed-shadow-mode.md
ekashida: I was just chatting about this with @nolanlawson and we were a little concerned about how similar it sounds to open/closed modes of shadow roots.

text/0000-mixed-shadow-mode.md
nolanlawson: It looks like `this.shadowRoot` is also not settable, even for closed-mode shadow elements. So we can't override the default behavior if we merge the two classes.

Screen Shot 2021-06-03 at 3 58 52 PM

Although I suppose even in the case where we merge the two classes and support closed shadow mode, we could just allow `shadowRoot` to be null? Would this be a problem?

text/0000-mixed-shadow-mode.md
pmdartus: Good point, this alternative is worth adding to the RFC. 

I would like to nuance the alternative with the following cons:
- Because Salesforce experiences are composed of components authored by multiple vendors (internal Salesforce devs, customers, ISV, etc), it is nearly impossible to coordinate such kind of switch. 
- Giving a smooth transition path would prevent custom disruption by letting them prepare the native shadow DOM rollout.

text/0000-mixed-shadow-mode.md
nolanlawson: @ekashida and I chatted about this. Since exposing `this.shadowRoot` would introduce new security concerns in the SA (since `this.template` is allowlisted, whereas `this.shadowRoot` allows unfiltered access), and since it's orthogonal to the mixed mode feature, we're going to punt on this for now. We can always consider how to expose `this.shadowRoot` later.

text/0000-mixed-shadow-mode.md
caridy: well, my initial reaction to this sort of feature is that when exposed, people will create a forking logic depending on the mode in this case, is that something we want? I will like to understand better, maybe with a realistic example, why people will need to know about the mode, for what?

caridy (changes requested):
LGTM, but the question around inspection time vs creation time is still a big one for me, let's resolve that before we merge this.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,199 @@
...
+will need to know whether it is currently operating in native or synthetic Shadow DOM. Any hints
+provided by the framework will only be available to components that set the `defaultShadowMode`
+property to `native` because this information is not useful to components that only support
+synthetic Shadow DOM.
+
+In terms of composition, a synthetic mode component can contain a native mode component, but the
+inverse is not allowed. Not only does this make things easier to reason about, but many existing
+components rely on workarounds in synthetic mode that are not possible to support in native mode.
+These observable differences are further discussed below. As this invariant cannot be asserted
+during compile-time, the engine will assert it during runtime.
```
nolanlawson, line 68: We're starting to discuss the complexities of implementing the synthetic-within-native restriction (e.g. how to handle synthetic-within-light-within-native, how to handle dynamic DOM node insertion), so I think it would be good to spell out exactly what benefits we get from restricting synthetic-within-native.

Also: why restrict synthetic-within-native, but not native-within-synthetic?

caridy: this is not an arbitrary restriction, this is a technical driven restriction, any slot element printed by a synthetic shadow will be inherited (owned) by the nearest document fragment, which is going to be native, and mess up with the distribution for that native.

nolanlawson: Thanks, this makes it much clearer why we should restrict it! 🙂  I can't think of a workaround, unless we were to transform `<slot>` into something else at compile time when it's a synthetic shadow component.

caridy: which will be a non-backward compatible change for that context, breaking some components, so better to just restrict it.

text/0000-mixed-shadow-mode.md
ekashida: The idea is that they might need to fork their logic due to the observable differences listed below. I agree that this is something we don't want to expose unless we have to so it doesn't need to go in the first iteration. Should I just remove this mention for now?

nolanlawson:
A question that came up to do is what to do about `lwc:dom="manual"` restrictions (e.g. cannot call `div.appendChild()` unless the `div` is in manual mode). It seems to me that mixed mode should continue enforcing these, even for components that are running in native mode, because we don't want component authors to deliberately break compat with synthetic shadow.

If `@lwc/synthetic-shadow` is never loaded though (100% native), then these restrictions don't need to be enforced IMO.

pmdartus:
> If `@lwc/synthetic-shadow` is never loaded though (100% native), then these restrictions don't need to be enforced IMO.

There is currently no guarantee that `@lwc/synthetic-shadow` will never be loaded. I think once we are further down the process of deprecating synthetic shadow DOM we should start removing those restrictions. Existing components will keep working but we won't surface the error when `lwc:dom="manual"` isn't set.

text/0000-mixed-shadow-mode.md
```diff
@@ -167,30 +164,29 @@ added to the component bundle.
...
-for existing components.
-
-Note that the current proposal for mixed mode considers polyfills as all-or-nothing. LWC is not in
-the business of providing workarounds for browser implementation issues, unless it affects the
-functionality of the framework itself.
+When setting `shadowMode` to `any`, it will be the responsibility of the component author to ensure
+that all components in the subtree are compatible with native Shadow DOM. Due to the nature of this
+feature, it is expected that initial usage will start at the leaf components and work its way up the
+component tree. If a component in the subtree does not support native Shadow DOM, the component
+author should work with the other component author to make the appropriate updates.
```
nolanlawson, line 179: So right now we either have

```js
shadowMode = undefined
shadowMode = 'any'
```

Is this right?

ekashida: Yes, that's correct. I didn't make it explicit because it was obvious to me, but I'll add it in there since you asked.

text/0000-mixed-shadow-mode.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+This mode will result in the whole component subtree to operate in either native shadow mode or
+synthetic shadow mode. If the browser supports Shadow DOM the shadow mode will be native. If the
+browser does not support Shadow DOM (e.g., IE11) the shadow mode will be synthetic. The test to
+determine Shadow DOM support will be to check whether `window.ShadowRoot` is defined.
+
+Setting `shadowMode` to `undefined` is equivalent to not setting the property at all. A value of
+`undefined` results in the default behavior where the shadow mode of the component depends only on
+the inclusion of the `@lwc/synthetic-shadow` polyfill.
```
nolanlawson, line 59: @ekashida 

```suggestion
Setting `shadowMode` to `undefined` is equivalent to not setting the property at all. A value of
`undefined` results in the default behavior where the shadow mode of the component depends only on
the inclusion of the `@lwc/synthetic-shadow` polyfill. The only valid values are `undefined` and `"any"` – any other value will throw a compiler error.
```

nolanlawson: Actually on second thought, I think we may need to introduce an explicit value like `"default"` in addition to `undefined` and `"any"`. The problem is here:

```js
class Super extends LightningElement {
  static shadowMode = 'any'
}

class Sub extends Super {
  static shadowMode = undefined
}
```

With `renderMode`, [we allow subclasses to inherit the `static renderMode` from their superclass](https://github.com/salesforce/lwc/blob/8ddcfc44118d7ae12b3c1dcba131ba78cc66d693/packages/%40lwc/engine-core/src/framework/def.ts#L159-L167). We also currently do this for `preferNativeShadow`.

But in this case, it's ambiguous whether `Sub` should inherit the `shadowMode` from `Super` or not.

text/0000-mixed-shadow-mode.md
```diff
@@ -11,80 +11,88 @@ pr: https://github.com/salesforce/lwc-rfcs/pull/49
...
 ```
 
-Using this property to opt in will result in the whole component subtree to operate in native shadow
-mode. and it is the responsibility of the component author to ensure that all components are
-compatible with native Shadow DOM. Due to the nature of this feature, it is expected that initial
-usage will start at the leaf components and work its way up the component tree.
+This mode will result in the whole component subtree to operate in either native shadow mode or
+synthetic shadow mode. If the browser supports Shadow DOM the shadow mode will be native. If the
+browser does not support Shadow DOM (e.g., IE11) the shadow mode will be synthetic. The test to
+determine Shadow DOM support will be to check whether `window.ShadowRoot` is defined.
```
pmdartus, line 55: Bikeshed: The term `any` doesn't represent well the behavior described here. What do you think of: `prefer-native`, `native-if-available` ?

text/0000-mixed-shadow-mode.md
```diff
@@ -11,80 +11,88 @@ pr: https://github.com/salesforce/lwc-rfcs/pull/49
...
+ability for a component to access native Shadow DOM APIs, even while the synthetic shadow polyfill
+is applied. Mixed mode should be transparent to components and the framework should guarantee this
 through integration testing, which is further discussed below.
 
 ### Opting in
 
-A component can signal that it prefers native Shadow DOM by setting a static property called
-`preferNativeShadow` to `true`.
+A component can signal that it supports both synthetic and native Shadow DOM by setting a static
+property called `shadowMode` to `any`.
```
pmdartus, line 44: Bikeshed: The name `shadowMode` can easily be confused with the shadow root mode (`open` or `closed`).

text/0000-mixed-shadow-mode.md
nolanlawson: @pmdartus It's hard to think of names that capture everything going on here. We have two modes:

1. Put me in native if the environment supports it, or if I'm off-core and there's no `@lwc/synthetic-shadow`
2. Put me in native if my ancestor forces me (by choosing option 1 above), or if I'm off-core and there's no `@lwc/synthetic-shadow`

text/0000-mixed-shadow-mode.md
nolanlawson: Maybe in the future we can call this "encapsulation mode" to avoid confusion? That's [what MDN calls it](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#parameters).

ekashida: Yeah, [I also had this concern](https://github.com/salesforce/lwc-rfcs/pull/49#discussion_r645177988) but I couldn't come up with a better name. I used `shadowSupportMode` internally, what does everyone think about that?

tedconn: What about reusing `renderMode` from light dom?

nolanlawson: I think it would be overloading `renderMode` a bit too much. Also ideally in the far future we will deprecate `shadowSupportMode` (or whatever it's called), but we will still support `renderMode`.

ekashida: Changed to `shadowSupportMode`

PR #44, opened 2021-03-17
RFC: Light DOM Support

abdulsattar:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/abdulsattar/light-dom/text/0000-light-dom.md)

Based on the initial RFC #40

PR #50, opened 2021-05-14
RFC: light DOM scoped styles

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/light-dom-scoped-styles/text/0000-light-dom-scoped-styles.md)

Builds on #44.

### DIFF
--- a/text/0000-light-dom-scoped-styles.md
+++ b/text/0000-light-dom-scoped-styles.md
@@ -1,0 +1,173 @@
+---
+title: Light DOM Scoped Styles
+status: DRAFTED
+created_at: 2021-05-14
+updated_at: YYYY-MM-DD
+champion: Nolan Lawson (nolanlawson), Abdulsattar Mohammed (abdulsattar)
+pr: https://github.com/salesforce/lwc-rfcs/pull/50
+---
+
+# Light DOM Scoped Styles
+
+## Summary
+
+This proposal adds a mechanism for light DOM components ([#44](https://github.com/salesforce/lwc-rfcs/pull/44)) to have co-located CSS files that only apply to that component.
+
+## Basic example
+
+Since there is already `*.css` for unscoped light DOM CSS files, we add `*.scoped.css` for scoped light DOM CSS:
+
+```html
+<!-- x/foo.html -->
+<template>
+  <div>Hello</div>
+</template>
+```
+
+```js
+// x/foo.js
+import { LightningElement } from 'lwc'
+
+export default class Foo extends LightningElement {
+  static shadow = false
+}
+```
+
+```css
+/* x/foo.scoped.css */
+div {
+  color: green;
+}
+```
+
+The result will be:
+
+```html
+<x-app>
+  <style class="x-foo_foo" type="text/css">
+    div.x-foo_foo { color: green; }
+  </style>
+  <div class="x-foo_foo">Hello</div>
+</x-app>
+```
+
+## Motivation
+
+For Light DOM ([#44](https://github.com/salesforce/lwc-rfcs/pull/44)), the default assumption is that styles are unscoped. In other words, a light DOM component essentially just contains `<style>` tags that are inserted into the DOM _in situ_. (The actual implementation may be slightly different, but this is the basic mental model.)
+
+Many frameworks, however, have a concept of _scoped styles_, even without using shadow DOM. These offer good developer ergonomics, because developers can concentrate on the CSS co-located with a particular component, without worrying how that CSS might affect other components.
+
+Prior art:
+
+- [Vue: `<style scoped>`](https://vue-loader.vuejs.org/guide/scoped-css.html)
+- [Svelte: scoped styles](https://svelte.dev/docs#style)
+- [React: CSS Modules in `create-react-app`](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/)
+- [Stencil: scoped styles in light DOM](https://stenciljs.com/docs/styling#scoped-css)
+- [BEM: Block, Element, Modifier](https://en.bem.info/)
+
+## Detailed design
+
+### Scoping token
+
+For the purposes of this document, a _scoping token_ is some string that we use to scope CSS selectors to a particular region of the DOM.
+
+In general, there are two approaches for applying scoping tokens to the DOM: classes or attributes. Both have [the same CSS specificity](https://alistapart.com/article/braces-to-pixels/#section4), but [historically](https://github.com/threepointone/glamor/issues/339) and [presently](https://github.com/salesforce/lwc/pull/2329), classes are faster than attributes in terms of the browser's [style calculation](https://developers.google.com/web/fundamentals/performance/rendering/#the_pixel_pipeline) process. So we prefer classes.
+
+This means that, with scoped styles, all CSS selectors inside of the `*.scoped.css` file will have an added class, which scopes the rules to that particular component.
+
+For example, the user might author:
+
+```css
+div > .bar {}
+```
+
+This is compiled to:
+
+```css
+div.x-foo_foo > .bar.x-foo_foo {}
+```
+
+Any existing classes in the template will be merged with the scoping token classes.
+
+### Comparisons with synthetic shadow DOM
+
+The current design of LWC's synthetic shadow scoped styles has several features that we do not want to emulate with light DOM scoped styles:
+
+- Using attributes over classes.
+- Allowing `@import` to import other CSS files.
+- Evaluating scoped CSS at runtime rather than compile time.
+- Using `lwc:dom="manual"` and `MutationObserver` to dynamically update the scoping token for new DOM nodes.
+
+The classes vs attribute issue is already addressed above. So let's cover the other differences.
+
+#### `@import` and runtime scoping
+
+On the server, synthetic shadow scoped styles are compiled to a JavaScript function which takes the scoping tokens as input and outputs a string. This means that styles are scoped at runtime rather than compile time.
+
+The main reason for this is that two component templates can contain `*.css` files that `@import` the same shared CSS file. This shared CSS file would need to be separately scoped for the two components. Since these shared CSS files could be very large (e.g. a CSS framework), and since they may `@import` further CSS files, it would be impractical to scope all of these files at compile time for each component. It would lead to a lot of duplicated code being generated on the server and sent to the client.
+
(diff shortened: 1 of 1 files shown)


text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,232 @@
...
+/* parent.scoped.css */
+x-child {}
+```
+
+```css
+/* child.scoped.css */
+x-child {}
+```
+
+This will result in two scoping tokens being applied to the `<x-child>` element – one from the parent, and another from the child. Precendence order is not guaranteed.
```
tedconn, line 166:
```suggestion
This will result in two scoping tokens being applied to the `<x-child>` element – one from the parent, and another from the child. Order of precedence is not guaranteed.

```

text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,233 @@
...
+
+Any existing classes in the template will be merged with the scoping token classes.
+
+### Comparisons with synthetic shadow DOM
+
+The current design of LWC's synthetic shadow scoped styles has several features that we do not want to emulate with light DOM scoped styles:
+
+- Using attributes over classes.
+- Allowing `@import` to import other CSS files.
+- Evaluating scoped CSS at runtime rather than compile time.
```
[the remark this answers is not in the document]
nolanlawson: I mean the fact that we generate a function that takes the scoping token as input and outputs a string, rather than just generating a string server-side. I'll make it clearer here.

text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,233 @@
...
+
+- [Vue: `<style scoped>`](https://vue-loader.vuejs.org/guide/scoped-css.html)
+- [Svelte: scoped styles](https://svelte.dev/docs#style)
+- [React: CSS Modules in `create-react-app`](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/)
+- [Stencil: scoped styles in light DOM](https://stenciljs.com/docs/styling#scoped-css)
+- [BEM: Block, Element, Modifier](https://en.bem.info/)
+
+## Detailed design
+
+### Scoping token
```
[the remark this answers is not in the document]
nolanlawson: I'll add something about this. I agree that in my own mental model, applying the scope token to the root element also felt kind of weird. But it felt weirder _not_ to have it, since otherwise `x-foo` doesn't work.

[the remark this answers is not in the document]
nolanlawson: Done, added a section.

text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,233 @@
...
+
+Prior art:
+
+- [Vue: `<style scoped>`](https://vue-loader.vuejs.org/guide/scoped-css.html)
+- [Svelte: scoped styles](https://svelte.dev/docs#style)
+- [React: CSS Modules in `create-react-app`](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/)
+- [Stencil: scoped styles in light DOM](https://stenciljs.com/docs/styling#scoped-css)
+- [BEM: Block, Element, Modifier](https://en.bem.info/)
+
+## Detailed design
```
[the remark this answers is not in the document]
nolanlawson: > I would personally favor a strong signal

This seems unnecessary to me, since we don't already do this for existing `*.css` files. Developers can include or omit the `*.css` file, and the compiler will pick it up automatically. Why shouldn't it be the same for `*.scoped.css` files? If we're going to have a new system for scoped CSS, I would prefer extending it back to regular CSS as well – even if it's a breaking change, it would be better for consistency.

I admit `<style scoped>` would be more preferable from the perspective of maintaining Rollup/Jest/etc. plugins, but again, it's a totally different system from existing CSS, and it wouldn't be clear to developers why they need one system for one and another system for the other. Also it would raise the question of why we don't allow "single file" components ala Vue/Svelte - e.g. a `<script>` tag containing the JS directly inlined in the HTML.

text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,330 @@
...
+
+Light DOM components are already opt-in (using `static renderMode = 'light'`), and scoped light DOM styles would also be opt-in, using `*.scoped.css`.
+
+# How we teach this
+
+Conceptually, scoped light DOM styles will have to be bundled up into a larger discussion of light DOM versus shadow DOM. Because switching between the two will never be as simple as flipping a boolean flag, and because the differences between the two can have a wide-ranging impact, we have to be careful about how we communicate the differences to developers.
+
+# Unresolved questions
+
+There are no unresolved questions at this time.
```
nolanlawson, line 330: To add: should `*.css` be the global, or should it be the scoped one, and `*.global.css` is the global one? In Vue, global is the default, whereas in Svelte, scoped is the default. There are different intuitions about this, so it's not clear one or the other is the best.

pmdartus (approved):

text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,366 @@
...
+updated_at: YYYY-MM-DD
+champion: Nolan Lawson (nolanlawson), Abdulsattar Mohammed (abdulsattar)
+pr: https://github.com/salesforce/lwc-rfcs/pull/50
+---
+
+# Light DOM Scoped Styles
+
+## Summary
+
+This proposal adds a mechanism for components ([#44](https://github.com/salesforce/lwc-rfcs/pull/44)) to have co-located CSS files that are scoped to that component, but without relying on shadow DOM.
```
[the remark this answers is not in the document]
nolanlawson: @pmdartus I explicit removed the bit about "light DOM components" since this feature now works in shadow DOM components as well.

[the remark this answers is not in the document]
nolanlawson: I'll say "components, including light DOM components"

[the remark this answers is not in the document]
pmdartus: Perfect.

text/0000-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,369 @@
...
+
+Prior art:
+
+- [Vue: `<style scoped>`](https://vue-loader.vuejs.org/guide/scoped-css.html)
+- [Svelte: scoped styles](https://svelte.dev/docs#style)
+- [React: CSS Modules in `create-react-app`](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/)
+- [Stencil: scoped styles in light DOM](https://stenciljs.com/docs/styling#scoped-css)
+- [BEM: Block, Element, Modifier](https://en.bem.info/)
+
+Furthermore, scoped styles are useful even for shadow DOM. Consider a shadow parent component with a light child component: the parent's styles "bleed" into the child, which may be surprising for developers who are accustomed to the Vue/Svelte model.
```
brandonferrua, line 69: This paragraph is confusing me a bit, are you saying scoped styles would prevent the bleeding of styles?

```html
<x-app>
  <style class="x-foo_foo" type="text/css">
    div.x-foo_foo { color: green; }
  </style>
  <x-bar>
    #shadowroot
      <style>
        h2 { color: red; }
      </style> 
      <h2 class="x-foo_foo">Hello</h2>
  </x-bar>
</x-app>
```

The light DOM element with scoped styles would still render as red

nolanlawson: @brandonferrua I'm talking about this case:

```html
<x-shadow>
  #shadowroot
    <style type="text/css">
      div { color: green; }
    </style>
    <x-light>
      <div>Hello</div>
    </x-light>
</x-shadow>
```

In this case, the light DOM component's `<div>` will have `color: green`, which may surprise developers who think of shadow DOM styles as "scoped." (It's "scoped," but not in the same way most JS frameworks talk about scoping!)

text/0000-light-dom-scoped-styles.md
brandonferrua: Is there any support for `slot` in this world?

brandonferrua: Anything different about custom properties? 

Thinking of this extreme case, what would the color of the slotted `span` be? `cornflowerblue` or `blue`?

```html
<x-app>
  <style>
    :root {
      --color-red: red;
      --color-blue: blue;
    }
  </style>
  <style class="x-foo_foo" type="text/css">
    div.x-foo_foo { 
      --color-red: maroon; 
      color: var(--color-blue); 
    }
    div.x-foo_foo span {
      --color-blue: cornflowerblue; 
    }
  </style>
  
  <div class="x-foo_foo">
    <h2>Hello</h2>
    <x-bar>
      #shadowRoot
        <style>
          h2 {
            color: var(--color-red);
          }
          ::slotted(span) {
            color: var(--color-blue)
          }
        </style>
        <h2>Hello</h2>
        <slot>↦ span</slot>
      #lightdom
      <span>slotted element</span>
    </x-bar>
  </div>
</x-app>
```

nolanlawson: This was a bit of a brain-teaser, so I wrote [a demo](https://github.com/nolanlawson/lwc-barebone/commit/167c105edb0a20ed89b0152426966099144de184). 🙂 

Looks like the answer is cornflower blue. It's a bit confusing because both `color` and custom properties are inherited, but basically yes, the `--color-blue` inherits down from the `<div>`.

Screen Shot 2021-09-30 at 12 47 48 PM

(Using Firefox but it's the same in Chrome; just easier to see in Firefox since Chrome doesn't visualize constructable stylesheets very well.)

brandonferrua:
Just asking this again cause I didn't see anything about it in the RFC. @nolanlawson 

> Is there any support for `slot` in this light dom world? If so, would we rely on `:slotted` to style the slotted content?

text/0116-light-dom-scoped-styles.md
```diff
@@ -0,0 +1,369 @@
...
+
+With global light DOM styles, the `:host` selector is inserted as-is. This means that a light DOM child component can use `:host` to target its shadow parent's host container.
+
+With scoped styles, the intuition around `:host` should be a bit different. Rather than inserting the CSS as-is, scoped styles imitate the encapsulation of shadow DOM components. So it makes sense for `:host` to follow shadow DOM-like semantics. This means that `:host` should refer to the root element of the light DOM component – i.e., the root of the scoped DOM region.
+
+So for instance:
+
+```css
+/* light.css */
+:host {
```
gonzalocordero, line 201: @nolanlawson I find this a little counter-intuitive: 

"I'm a shadow friendly component, my styles are robust and encapsulated, I do allow others externally to apply some styles to the `host` element, as I understand I'm only a guest for a certain point and time in the cycle, but I would never expect a child component of my own to be able to mess with my styles"

PR #51, opened 2021-05-18
RFC Template for ePrivacyConsentCookie component

hemantpandey17:
This LWC component provide utility functions which allow teams to incorporateCookie Consent mechanism in their components. This abstracts all the cookie fetching, reading, writing, updating logic and provide library functions to achieve the functionality.


gonzalocordero:
@hemantpandey17 this needs to be moved into the following RFC repo instead: https://github.com/salesforce/lbc-rfcs/blob/master/0000-component.spec.md

Thanks!

PR #52, opened 2021-07-05
feat: add `lwc:spread` Directive RFC

abdulsattar:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/abdulsattar/dynamic-attrs/text/0000-dynamic-attributes.md)

### DIFF
--- a/text/0000-dynamic-attributes.md
+++ b/text/0000-dynamic-attributes.md
@@ -87,8 +87,9 @@ template directly. In the above example, `c-child` is always passed `World` as `
 
 ## Locker Integration
 
-`lwc:spread` will only support LWS. Usage of `lwc:spread` will mark the element as risky and add the `renderer` key to the element
-in the template.
+Since `lwc:spread` allows setting any property, there is a risk of component authors trying to assign properties like `innerHTML`.
+To prevent this, usage of `lwc:spread` will mark the element as risky and add the `renderer` key to the element
+in the template so that locker can sanitize those properties.
 
 ## Drawbacks
 
--- a/text/0000-dynamic-attributes.md
+++ b/text/0000-dynamic-attributes.md
@@ -7,10 +7,10 @@ champion: Mohammed Abdul Sattar (abdulsattar)
 pr: https://github.com/salesforce/lwc-rfcs/pull/52
 ---
 
-# Passing Dynamic Attributes/Properties
+# `lwc:spread` Directive
 
 ## Summary
-This proposal adds a mechanism to pass a dynamic attributes/properties/listeners to elements in
+This proposal adds a mechanism to pass a dynamic properties to elements in
 an LWC template using a new directive `lwc:spread`.
 
 ## Example
@@ -19,13 +19,7 @@ an LWC template using a new directive `lwc:spread`.
 export default class MyComponent extends LightningElement {
     childProps = { 
         name: 'My Child',
-        class: 'my-class',
-        onclick: this.handleClick,
     };
-    
-    handleClick() {
-        console.log('Clicked')
-    }
 }
 ```
 ```html
@@ -35,12 +29,12 @@ export default class MyComponent extends LightningElement {
 </template>
 ```
 
-The new `lwc:spread` directive passes `childProps` as properties/attributes/event listeners to `x-child`.
+The new `lwc:spread` directive passes `childProps` as properties to `x-child`.
 
 ## Motivation
 
-In LWC, there is no way to pass props/attributes to a child component when the component author
-doesn't know in advance all the props/attributes that are needed.
+In LWC, there is no way to pass props to a child component when the component author
+doesn't know in advance all the props that are needed.
 
 One such usecase is `lwc:dynamic` which allows component authors to lazily instantiate components. However,
 the props that we pass to the component can't be dynamic: we need to pass them in the template HTML itself.
@@ -52,7 +46,7 @@ We can, of course, not pass any props and once the element renders, query the
 element and set the props programmatically. We lose reactivity this way and also force
 the component to handle `null` props which might not be desirable.
 
-This proposal allows not just `lwc:dynamic` but all other elements to accept an object which is bound as props/attrs/listeners
+This proposal allows not just `lwc:dynamic` but all other elements to accept an object which is bound as props
 at runtime.
 
 ### Prior Art
@@ -63,22 +57,47 @@ at runtime.
 
 ## Detailed Design
 
-A new LWC directive, `lwc:spread` will be introduced that can be used to spread the props/attributes/listeners
-to children elements.
+A new LWC directive, `lwc:spread` will be introduced that can be used to spread the props to children elements.
 
 ### Structure
-It accepts an object with keys as the attribute/property/listener names and values as the corresponding attribute/property/listener values.
+It accepts an object with keys as the property names and values.
 
-This object is merged with props/attrs/listeners passed in the template (essentially `Object.assign`). However, since template
-compiler splits all the props passed in the template into attributes, properties and event listeners, we do the same thing at runtime. We use the same algorithm that the template compiler uses and whatever the [engine does for native custom elements](https://rfcs.lwc.dev/rfcs/lwc/0000-third-party-web-components-support#proposed-solution-2).
+This object is merged with props passed in the template (essentially `Object.assign`).
 
 There is one difference however: the set of properties passed in the template doesn't change but it changes in `lwc:spread` i.e.
-an attribute can't be present in one render and absent in the next render if it is defined in the template. With
-`lwc:spread` though, it can be present in one render and not be present in the next.
+a property can't be present in one render and absent in the next render if it is defined in the template. With
+`lwc:spread` though, it can be present in one render and not be present in the next. See example below.
 
-- attrs/listeners are added/removed based on their presence/absence in `lwc:spread`.
-- properties can't be "removed" if they are not present in subsequent renders. So, we set them to `undefined`. New properties are
-set as usual.
+**Example**
+
+```html
+<template>
+    <x-child name={name}></x-child> <!-- in first render -->
+    <x-child></x-child> <!-- in second render, name is missing. This is not possible with template compiler -->
+</template>
+```
+
+```html
+<template>
+    <x-child lwc:spread={props}>
+</template>
+```
+
+```js
+export default class App extends LightningElement {
+    handleClick() {
+        this.props = showName ? {name: 'lwc'} : {}; // in one render we're passing `name` prop, but in another one there's no `name` property.
+    }
+
+}
+```
+
+In cases where a property from earlier render is missing, we have two options:
+1. Explicity set it to `undefined` or `null`.
+2. Leave it untouched. Meaning we just assign the properties we get in the current render without touching any of the earlier set properties.
+
+We'll go ahead with Option 1. We'll explicity set them to `undefined`. The advantage of this is that it guarantees that at any point
+the properties on an element are the ones that are present in the template and the ones present in `lwc:spread` in that render.
 
 An interesting side effect of the above behavior can be observed when properties have default values:
 
@@ -105,79 +124,21 @@ export default class App extends LightningElement {
 ```
 Customers should be educated on the above behavior.
 
-### Reactivity
-There's no precendent in LWC for change in the set of attributes/properties/listeners. An element can't have a property in one
-render and not have it in another render. LWC Engine actually throws when it detects such a change. In the case of `lwc:spread`,
-we'll have to relax this validation.
-
-In every render, `lwc:spread` will merge with the older value of `lwc:spread`. Any keys not present in the new `lwc:spread` but
-present in the old `lwc:spread` will be set to `null` explicitly in `lwc:spread`. With this merging logic, all the new
-attributes/listeners will be set and missing attributes/listeners will be removed. Missing properties will be explicitly set to `null`. This results in a consistent, logical behavior for users. Please see the following example.
-
-### Example
-```html
-// x/test.html
-<template>
-    <x-child lwc:spread={props}></x-child>
-    <button onclick={toggle}>Toggle</button>
-</template>
-```
-
-```js
-import { LightningElement } from 'lwc';
-
-export default class Test extends LightningElement {
-    props = {
-        attrA: 'a',
-        attrB: 'b',
-
-        propA: 'a',
-        propB: 'b',
-
-        listenerA() { },
-        listenerB() { },
-    }
-    toggle() {
-        this.props = {
-            attrA: 'a', // remains same
-            attrC: 'c', // new attribute added using setAttribute
-            // attrB is removed using `removeAttribute`
-
-            propA: 'a', // remains same
-            propC: 'c', // new property assigned using elm.propC = 'c'
-            // propB is "removed" using elm.PropB = null;
-
-            listenerA() { }, // remains same
-            listenerC() { }, // new listener added using `addEventListener`
-            // listenerB is removed using `removeEventListener`
-        }
-    }
-}
-```
-
 ### Casing
 LWC Template Compiler converts kebab-case attrs/props to camelCase to match the Javascript counterparts. Since props `lwc:spread`
 are bound in Javascript, they need to be defined in camelCase. This means that `lwc:spread` won't do any case transformation.
 
-### Disallowed keys
-Some keys (e.g. 'for:each') alter the compiled HTML output itself, so they are not allowed.
-`lwc:spread` does not allow LWC directives in general.
-
-### Overriding attributes/props
+### Overriding props
 ```html
 <template>
-  <c-child name="Hello" lwc:spread={childArgs}></c-child>
+  <c-child name="Hello" lwc:spread={childProps}></c-child>
 </template>
 ```
 ```js
-childArgs = { name: "World" };
+childProps = { name: "World" };
 ```
-`lwc:spread` will take precedence to whatever props/attrs are declared in the template directly. In the above example,
+`lwc:spread` will take precedence to whatever props are declared in the template directly. In the above example,
 `c-child` is always passed `World` as `name`. Additionally, there can be only one `lwc:spread` on a directive.
-
-### Locker Integration
-
-The usage of `lwc:spread` will be considered risky and the element will be injected with a `renderer` property that will be overridden by locker.
 
 ## Drawbacks
 
--- a/text/0000-dynamic-attributes.md
+++ b/text/0000-dynamic-attributes.md
@@ -1,5 +1,5 @@
 ---
-title: Passing Dynamic Attributes/Properties
+title: `lwc:spread` Directive
 status: DRAFTED
 created_at: 2021-07-05
 updated_at: YYYY-MM-DD
@@ -10,8 +10,8 @@ pr: https://github.com/salesforce/lwc-rfcs/pull/52
 # Passing Dynamic Attributes/Properties
 
 ## Summary
-This proposal adds mechanism to pass a dynamic attributes or properties bag to elements in
-an LWC template.
+This proposal adds a mechanism to pass a dynamic attributes/properties/listeners to elements in
+an LWC template using a new directive `lwc:spread`.
 
 ## Example
 ```js
@@ -19,7 +19,6 @@ an LWC template.
 export default class MyComponent extends LightningElement {
     childProps = { 
         name: 'My Child',
-        style: 'width: 50px',
         class: 'my-class',
         onclick: this.handleClick,
     };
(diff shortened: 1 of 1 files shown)


text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+}
+```
+```html
+// x/myComponent.html
+<template>
+    <x-child lwc:spread={childProps}></x-child>
+</template>
+```
+
+The new `lwc:spread` directive passes `childProps` as properties/attributes to `x-child`.
```
abdulsattar, line 39: Couldn't think of a better name.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+
+In LWC, there is no way to pass props/attributes to a child component when the component author
+doesn't know in advance all the props/attributes that are needed.
+Adding this feature can be helpful in creating wrapper components that extend the functionality
+of an existing component while being oblivious to all the props that component requires.
```
pmdartus, line 46: I am not sold on how this proposal would make it easier to create wrapper components to extends existing component capabilities. This proposal solves the issue when you want to pass a set of properties/attributes instead of discrete bindings via the template. But it still requires the wrapper component to add all the public properties to be exposed via `@api`.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+A new LWC directive, `lwc:spread` will be introduced that can be used to spread the props/attributes
+to the children elements.
+
+### Structure
+It accepts an object with keys as the attribute names and values as the
+corresponding attribute values.
+
+LWC Template compiler separates all attributes into:
+1. Attributes
+2. Properties
```
pmdartus, line 65: Missing event listener.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+3. `style` attribute
+4. `class` attribute
+
+The same algorithm will be used for separating those attributes at runtime.
```
pmdartus, line 69: Just to confirm, the passed object is a flat object and the LWC engine will have to figure out at runtime if an object property is either an attribute, a property, or an event listener. Correct?

A couple of questions on this topic:
- What happens when a property is present in a rendering cycle and removed in the next rendering cycle (eg. `{ name: 'foo' }` --> `{ }`)? What does it mean for properties, attributes, and event listeners?
- What do `null` and `undefined` map to?
- Until now all the event listeners are automatically bound to the component (`this` value being the component instance)? What is the expected behavior with this proposal?

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+  <c-child name="Hello" lwc:spread={childArgs}></c-child>
+</template>
+```
+```js
+childArgs = { name: "World" };
+```
+
+The order in which attributes are specified matters. If the explicit attribute is passed
+before the attribute bag, the attribute bag takes precedence. Otherwise, the explicit attribute
+takes precedence. Much like `Object.assign`.
```
pmdartus, line 87: Attribute order in HTML doesn't matter in general. This proposal would be a departure from this. I don't have a strong opinion against this, but I wanted to call this out.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+    
+    handleClick() {
+        console.log('Clicked')
+    }
+}
+```
+```html
+// x/myComponent.html
+<template>
+    <x-child lwc:spread={childProps}></x-child>
```
pmdartus, line 35: One of the requirements I have with this proposal, it that it should work on both standard and custom elements.

text/0000-dynamic-attributes.md
abdulsattar: There wasn't a chance for having duplicate attributes till now (LWC throws a compile time error when it detects duplicate attributes).

I looked at Vue and Svelte: In Svelte, the order matters. Vue always gives precedence to the attributes explicitly assigned.

text/0000-dynamic-attributes.md
abdulsattar: > Just to confirm, the passed object is a flat object and the LWC engine will have to figure out at runtime if an object property is either an attribute, a property, or an event listener. Correct?

Correct.

Answers to your questions:
- Attributes and event listeners are removed. Not clear what to do with properties. I put it as an open question.
- Maybe we can use `null` and `undefined` to remove the properties?
- We should keep the same behavior.

text/0000-dynamic-attributes.md
abdulsattar: That's the same expectation I have as well. I'll add an example for standard element to make it clear.

text/0000-dynamic-attributes.md
abdulsattar: React as well does `Object.assign` way of doing stuff.

abdulsattar: Svelte and React use the spread operator syntax for props: `<p {...props}>`. They also allow multiple spreads like: `<p {...props1} {...props2}>`. It's closer to `Object.assign`.

Vue on the other hand does similar to this proposal: `v-bind="props"` and discards the order.

We can choose what Vue does.

text/0000-dynamic-attributes.md
nolanlawson: I was curious what the other frameworks do here. In the case of Svelte, Vue, and React, they all seem to agree that

```js
{
  foo: '',
  bar: null,
  baz: undefined
}
```

should render as `<div foo=""></div>`. This makes sense to me.

Screen Shot 2021-07-06 at 12 12 57 PM
Screen Shot 2021-07-06 at 12 02 23 PM
Screen Shot 2021-07-06 at 12 10 17 PM

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+## Example
+```js
+// x/myComponent.js
+export default class MyComponent extends LightningElement {
+    childProps = { 
+        name: 'My Child',
+        style: 'width: 50px',
+        class: 'my-class',
+        onclick: this.handleClick,
+    };
```
nolanlawson, line 25: I take it that for strings like `ariaDisabled`, this will automatically convert to `aria-disabled`? Will we support the kebab-cased versions as well, or only the camelCased versions?

There are also certain odd attributes like `aria-posinset` which [would look most natural in camelCase as `ariaPosInSet`](https://wicg.github.io/aom/spec/aria-reflection.html#idl_attr_exceptions), but I suppose we should prefer `ariaPosinset` to avoid having to keep a list of exceptions?

nolanlawson: Depending on what we decide here, this may also influence event handers, e.g. `onclick` versus `onClick`.

abdulsattar: We can just leave it as is: `childProps = { 'aria-posinset': '5' };` without any conversion.

text/0000-dynamic-attributes.md
abdulsattar: I looked at what LWC does currently with `null`, `undefined`. I'd prefer if we keep the same behavior.

image

pmdartus: As a side note, Svelte uses runtime reflection to know if a value should be set as property or as attribute: https://github.com/sveltejs/svelte/blob/b46432060753c4fa3db11da8641f67e49c0016ce/src/runtime/internal/dom.ts#L239

text/0000-dynamic-attributes.md
pmdartus: Here is an example to better explain the point above. Let say that you want to extend the functionality of an `x-button` component with 3 public properties:

```js
// x-button.js

import { LigthningElement, api } from 'lwc';

export default class Buttton extends LightningElement {
  @api label;
  @api color; 
  @api active;
}
```

In the wrapper button component template you can use the newly introduced directive to pass down the public property it receives:

```html
<!-- x-wrapper-button.html -->
<template>
  <x-button lwc:spread={props}></x-button>
</template>
```

The main limitation with this proposal is that `x-wrapper-button` has to define the same set of public properties `x-button` defines for this work. This greatly limits the capability of extending capabilities for complex components like `lightning-input` with 50+ public properties and methods.

```js
// x-button.js

import { LigthningElement, api } from 'lwc';

export default class WrapperButtton extends LightningElement {
  // Indentical public properties than in x-button
  @api label;
  @api color; 
  @api active;
}
```

abdulsattar: Right, the only situation I see this helps is when we use `lwc:dynamic` (which is a totally different proposal).

text/0000-dynamic-attributes.md
abdulsattar: > For removed properties, I don't think we need to do anything special. Unlike all the frameworks listed above, in LWC, properties are passed individually and there is no guarantee a given property is passed or not via the template. In this regard, I don't think components should be notified that a property present in the previous rendering cycle is not present anymore.

The issue with this is:
```
<lightning-input lwc:spread={inputProps}></x-img>
```
If we start with `inputProps` with `{}`, then add `{disabled: true}` and then set it back to `{}`, we are saying `disabled` will be `true`.

We also can't say we'll explicitly mark all the missing props in the next cycle `undefined` because, when we do set `inputProps` to `{}`, the initial value of `disabled` might be `false` (as declared in the JS class). If we later set it to `{disabled: true}` and then to `{}` (and LWC marks it as `undefined`), it's inconsistent.

pmdartus: You are right, we should make it super explicit in the documentation that LWC won't do any magic with props. Unlike attributes, there is no way to know what is the initial value for a given prop.

caridy:
In principle, I don't agree with this direction, I believe this get us into the realm of dynamism just for the sake of sugar. You will never be able to migrate to a DOM template engine if you add this feature now.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+<template>
+    <x-child lwc:spread={childProps}></x-child>
+</template>
+```
+
+The new `lwc:spread` directive passes `childProps` as properties/attributes to `x-child`.
+
+## Motivation
+
+In LWC, there is no way to pass props/attributes to a child component when the component author
```
caridy, line 43: this statement is incorrect. It is perfectly possible to do it, but programatically. So, if you want to update this to say that in LWC Templates, then sure... but it is not really something that LWC itself prevents, but the HTML language itself doesn't have that capability.

text/0000-dynamic-attributes.md
nolanlawson: I see, so we wouldn't do any camelCase -> kabeb-case conversion? That's probably the simplest solution.

text/0000-dynamic-attributes.md
pmdartus: > So, if you want to update this to say that in LWC Templates, then sure... but it is not really something that LWC itself prevents, but the HTML language itself doesn't have that capability.

While I agree with your statement, by going the route of programmatically creating a component, setting the props, and inserting it into the DOM you are automatically opting out of the LWC reactivity and rendering model.

One of the main issues with this approach is that it forces developers to bridge LWC reactivity model with the programmatically created component. Another downside is that it makes the component unable to be server-side renderable.

This proposal might not be the best approach to solve this issue, however, I do believe LWC should provide some new primitives. I will add a point for the next dev sync to make sure we talk through this.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,91 @@
...
+1. Attributes
+2. Properties
+3. `style` attribute
+4. `class` attribute
+
+The same algorithm will be used for separating those attributes at runtime.
+
+### Disallowed keys
+Some keys (e.g. 'for:each') alter the compiled HTML output itself, so they are not allowed.
+`lwc:spread` does not allow LWC directives in general.
```
kevinv11n, line 73: Is this valid?

```html
<template>
  <template for:each={many} for:item={one}>
    <c-child lwc:spread={one}></c-child>
  </template>
</template>
```

kevinv11n (commented):
At one time we were evaluating a link-phase validation of reference attributes/properties. That is, if a component looks like:

```html
<template>
  <c-child x="Hello"></c-child>
</template>
```

During the compilation cycle we'd validate that `c-child` declares an `@api x`. There were many boundary cases this couldn't handle (eg a prototype chain defined at runtime) but in the common case where the prototype chain is fixed (and often very shallow) we can discover issues before runtime. 

What does `lwc:spread` mean to such a feature?

pmdartus:
> What does `lwc:spread` mean to such a feature?

When using `lwc:spread` directive the component is explicitly giving up on linking validation and also any other analyses relying on LWC  template metadata (eg. Komaci).

The mental model that I have about this proposal is that it allows expressing dynamic properties, attributes, and event handlers in a declarative fashion. it is possible to transpose the first [example](https://github.com/salesforce/lwc-rfcs/blob/ca98b64f5121748757ec02f98f60608407265e2e/text/0000-dynamic-attributes.md#example) with programmatic API (with some nuance): 

```js
// x/myComponent.js
export default class MyComponent extends LightningElement {
    handleClick() {
        console.log('Clicked')
    }

    renderedCallback() {
        const child = this.template.querySelector('x-child');
        child.name = 'My Child';
        child.setAttribute('style', 'width: 50px');
        child.classList.add('my-class');
        child.addEventListener('click', (evt) => this.handleClick(evt));
    }
}
```

```html
// x/myComponent.html
<template>
    <x-child></x-child>
</template>
```

The linking validation will not be capable of analyzing the properties and attribute in the programmatic case either. The added benefit of this proposal, when it comes to metadata collection, is that `lwc:spread` directive clearly indicates that a component can't be fully analyzed.

The usage of the `lwc:spread` directive is not something we want to promote in general. In most cases, this directive will be associated with the `lwc:dynamic` directive to dynamically set the component constructor at runtime. The soil usage of the `lwc:dynamic` directive would also prevent any link time checks.

text/0000-dynamic-attributes.md
jonskeen: What if you already have map of the values in the wrapper component? For example you're building a `UL` using `for:each` to loop over an array of objects and construct `LI`s. If the `LI` contains a component expecting the values from the object, it would be so much nicer to spread the object into the internal component. Currently one must list each prop, which makes adding public properties to the inner component a bigger hassle than it needs to be.

```
<ul>
    <template for:each={items} for:item="item">
    <li>
        <some-component lwc:spread={item}></some-component>
    </li>
    </template>
</ul>
```

abdulsattar: @jonskeen what you mentioned is syntactic sugar. It's not like it's impossible to do it currently, it's just that `lwc:spread` will make passing those attributes a little easier. `lwc:spread` also comes at a cost: we can't do static analysis on `some-component` to see which properties are being passed to it.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+created_at: 2021-07-05
+updated_at: YYYY-MM-DD
+champion: Mohammed Abdul Sattar (abdulsattar)
+pr: https://github.com/salesforce/lwc-rfcs/pull/52
+---
+
+# Passing Dynamic Attributes/Properties
+
+## Summary
+This proposal adds a mechanism to pass a dynamic attributes/properties/listeners to elements in
```
caridy, line 13: why distinguishing or mentioning attributes and listeners here? the way I see it... we only give them support for properties... and it is the responsibility of the developer to know what props to use, we don't try to make sense of them.

text/0000-dynamic-attributes.md
caridy: I think this has changed since the last conversation, to use aria, you must use the prop name, otherwise it does nothing.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+```html
+// x/myComponent.html
+<template>
+    <x-child lwc:spread={childProps}></x-child>
+</template>
+```
+
+The new `lwc:spread` directive passes `childProps` as properties/attributes/event listeners to `x-child`.
```
caridy, line 38: same thing... just props.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Prior Art
+- React ([JSX Spread Operator](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes))
+- Vue ([`v-bind`](https://v3.vuejs.org/api/directives.html#v-bind))
+- Svelte ([Spread Props](https://svelte.dev/tutorial/spread-props))
+- Lit ([Discussion for support](https://github.com/lit/lit/issues/923), [PR](https://github.com/lit/lit/pull/1960))
+
+## Detailed Design
+
+A new LWC directive, `lwc:spread` will be introduced that can be used to spread the props/attributes/listeners
```
caridy, line 66: same... only props.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+to children elements.
+
+### Structure
+It accepts an object with keys as the attribute/property/listener names and values as the corresponding attribute/property/listener values.
```
caridy, line 70:
```suggestion
It accepts an object with key value pairs, each of which will be applied to the element.
```

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+
+This object is merged with props/attrs/listeners passed in the template (essentially `Object.assign`). However, since template
```
caridy, line 72: I disagree with this paragraph entirely. It is NOT related to what we do today... it is just equivalent to Object.assign() on the element itself. No heuristic is used.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+compiler splits all the props passed in the template into attributes, properties and event listeners, we do the same thing at runtime. We use the same algorithm that the template compiler uses and whatever the [engine does for native custom elements](https://rfcs.lwc.dev/rfcs/lwc/0000-third-party-web-components-support#proposed-solution-2).
+
+There is one difference however: the set of properties passed in the template doesn't change but it changes in `lwc:spread` i.e.
+an attribute can't be present in one render and absent in the next render if it is defined in the template. With
+`lwc:spread` though, it can be present in one render and not be present in the next.
+
+- attrs/listeners are added/removed based on their presence/absence in `lwc:spread`.
```
caridy, line 79: I also disagree with this statement... if you pass `onclick`, we will assign the `onclick` and we will NOT wrap it, or do anything to preserve the context (`this` value) in any ways. If you replace that value with a new `onclick` value, thats up to you, but if you decide to not pass it next time... it doesn't remove the previous value.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+        this.props = {name: "aura"}; // `name` in child is "aura";
+        this.props = {}; // `name` in child is `undefined`;
+    }
+}
+
+```
+Customers should be educated on the above behavior.
+
+### Reactivity
+There's no precendent in LWC for change in the set of attributes/properties/listeners. An element can't have a property in one
```
caridy, line 109: I don't understand this paragraph.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+render and not have it in another render. LWC Engine actually throws when it detects such a change. In the case of `lwc:spread`,
+we'll have to relax this validation.
+
+In every render, `lwc:spread` will merge with the older value of `lwc:spread`. Any keys not present in the new `lwc:spread` but
```
caridy, line 113: I also desagree with the semantics described here. I don't think we should diff the obj, that's too easy to trick.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,189 @@
...
+            listenerA() { }, // remains same
+            listenerC() { }, // new listener added using `addEventListener`
+            // listenerB is removed using `removeEventListener`
+        }
+    }
+}
+```
+
+### Casing
+LWC Template Compiler converts kebab-case attrs/props to camelCase to match the Javascript counterparts. Since props `lwc:spread`
```
caridy, line 159: The first sentence is half true... because we only do that for CE, built-ins are using attrs most of the time. I don't think it is relevant... but maybe just saying that they must be camelCase for properties, and no attributes, should be sufficient.

text/0000-dynamic-attributes.md
caridy: Yes, I think the paragraph is confusing... but essentially it is saying that `{one}` object cannot have a key called `for:each`, because that's not a valid property name... I think that should be implicit if we say we only support props... directives are NOT props... but I think it is worth noticing, just make it more clear in the paragraph.

text/0000-dynamic-attributes.md
caridy: I don't think we should talk here about what we do... but the semantics of the new feature instead. What we do now can change tomorrow, but the semantics of this feature will remain intact. Basically, just saying that the value on the spread needs to use prop names in a camelCase. Maybe also listing a couple of exceptions to the rule to ilustrate the semantics.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,150 @@
...
+    props = {};
+    renderedCallback() {
+        // in initial render props is {} and `name` in child is 'lwc'
+        this.props = {name: "aura"}; // `name` in child is "aura";
+        this.props = {}; // `name` in child is `undefined`;
+    }
+}
+
+```
+Customers should be educated on the above behavior.
```
caridy, line 125: I didn't quite understood this example.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+In cases where a property from earlier render is missing, we have two options:
+1. Explicity set it to `undefined` or `null`.
+2. Leave it untouched. Meaning we just assign the properties we get in the current render without touching any of the earlier set properties.
+
+We'll go ahead with Option 1. We'll explicity set them to `undefined`. The advantage of this is that it guarantees that at any point
+the properties on an element are the ones that are present in the template and the ones present in `lwc:spread` in that render.
+
+An interesting side effect of the above behavior can be observed when properties have default values:
```
caridy, line 102: also I didn't really understood why default values are important? I don't think they are. Default values in HTML are defined as initialization only, and they have side effects once, the first time they are set. I don't think we are talking about the same thing here... just confused.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,150 @@
...
+export default class App extends LightningElement {
+    handleClick() {
+        this.props = showName ? {name: 'lwc'} : {}; // in one render we're passing `name` prop, but in another one there's no `name` property.
+    }
+
+}
+```
+
+In cases where a property from earlier render is missing, we have two options:
+1. Explicity set it to `undefined` or `null`.
```
caridy, line 96: setting a prop to undefined might be problematic... some properties will try to stringify that, and will result on `"undefined"` as the value. I think we should stay away from prescribing anything here, just saying that whatever it is in the object, will be set into the element, in a one way street, no diffing, no segregation of values, etc. just straight Object.assign().

I also think that having the desugaring syntax would help a lot.

abdulsattar: The problem with this approach is it's dependent on the previous render as well, not just what is passed to `lwc:spread`.

E.g.
```html
<x-child lwc:spread={props}></x-child>
<button onclick={handleClick1}>Button 1</button>
<button onclick={handleClick2}>Button 2</button>
<button onclick={handleClick3}>Button 3</button>
<button onclick={handleClick4}>Button 4</button>
```
```js
handleClick1() {  this.props = {name: "alice"}; }
handleClick2() {  this.props = {age: 10};}
handleClick3() {  this.props = {name: "bob"};}
handleClick4() {  this.props = {age: 37};}
```
In the above component, depending on the _order_ in which the buttons are clicked, we get different behavior.

caridy (approved):
LGTM, only one note.

text/0000-dynamic-attributes.md
```diff
@@ -0,0 +1,101 @@
...
+```js
+childProps = { name: "World" };
+```
+
+`lwc:spread` will always be applied last. That means it will take precedence to whatever props are declared in the
+template directly. In the above example, `c-child` is always passed `World` as `name`. Additionally, there can be only one `lwc:spread` on a directive.
+
+## Locker Integration
+
+`lwc:spread` will only support LWS. Usage of `lwc:spread` will mark the element as risky and add the `renderer` key to the element
```
caridy, line 90: I don't understand the first sentence here.

caridy (approved):
LGTM

PR #53, opened 2021-09-10
RFC: SSR rehydration

divmain:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/divmain/ssr-rehydration/text/0000-ssr-rehydration.md).


text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+### The Happy Path
+
+With regard to SSR, there exists a happy path (Fig 1a) where the rendering output for a given request is identical on both the server and the client; identical VDOM is generated for the entire document in both environments.
+
+Depending upon the actual implementation, exceptions to this happy path include:
+
+- **Exception A:** One or more components are never rendered on the server (Fig 1b). Instead, a placeholder is rendered on the server, and later replaced on the client during CSR/hydration.
+  - **Example:** The page contains a cart icon with a dynamic number of items in cart.
+  - **Example:** The page contains a user avatar.
+- **Exception B:** One or more components are rendered only on the server (Fig 1c). No JS is sent to the client for these components and the original SSR'd HTML remains untouched by hydration and later client-side renders.
```
tedconn, line 40: > No JS is sent to the client

If no JS is sent to the client then it's not a web component, so it's just an `UnknownHTMLElement`?

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+
+However, rendering on the server can itself introduce performance issues, working against the original intent. Namely, if DOM that was generated from SSR HTML is not preserved during the initial CSR, this will result in potentially expensive DOM manipulation and unwanted visual artifacts.
+
+The aim of rehydration is to avoid the client-side performance issues of a naive SSR implementation, reusing original DOM wherever possible.
+
+### The Happy Path
+
+With regard to SSR, there exists a happy path (Fig 1a) where the rendering output for a given request is identical on both the server and the client; identical VDOM is generated for the entire document in both environments.
+
+Depending upon the actual implementation, exceptions to this happy path include:
```
tedconn, line 35: I can think of another exception, where for example, you want to render a dummy component on the server, for example `<router-container></router-container>`. On the client I would expect this to get fully connected and rendered but on the server I'll render a dummy implementation. It's like an expected mismatch.

text/0000-ssr-rehydration.md
tedconn: It sounds like this is basically server-side templating which won't be convenient for SPAs but great for MPAs?

divmain: Yes, this is conceptually comparable to React server components. There are a lot of downsides to using something like this, especially (as you point out) if you want a true SPA with page-level transitions handled by JS. In LWC, you would end up with `UnknownHTMLElements` on the page, yes.

text/0000-ssr-rehydration.md
divmain: Can you elaborate on how this is distinct from Exception A?  I think, perhaps, the exception description can be improved and that is where the gap is. Essentially, one thing (likely a placeholder of some kind) is rendered on the server and another thing is swapped in on the client.

There are actually two ways this could go from there:

1. You render something different on the client for the first render during rehydration. This results in an "expected mismatch" to use your words, and the patch takes care of swapping in the new content.
2. You render the same thing on the client for the first CSR during rehydration but hook into `connectedCallback` and update your state to render something else asynchronously on a subsequent/early CSR.

That does bring to mind something else worth considering. Should there exist another lifecycle for when a component is attached to pre-existing DOM during rehydration? I'm not yet aware of how lifecycle hooks are widely used by LWC devs, and I wouldn't want to overload `connectedCallback` and break existing apps.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+
+- **Exception A:** One or more components are never rendered on the server (Fig 1b). Instead, a placeholder is rendered on the server, and later replaced on the client during CSR/hydration.
+  - **Example:** The page contains a cart icon with a dynamic number of items in cart.
+  - **Example:** The page contains a user avatar.
+- **Exception B:** One or more components are rendered only on the server (Fig 1c). No JS is sent to the client for these components and the original SSR'd HTML remains untouched by hydration and later client-side renders.
+  - **Example:** The page contains a footer that never changes.
+  - **Example:** The landing page is mostly static and is cached on a CDN.
+- **Exception C:** Render-critical state is changed between the time of SSR and CSR rehydration, causing an unexpected mismatch (Fig 1d).
```
blittle, line 43: This should only be possible on temporal based data (like dates). Generally API data should not be a problem.

text/0000-ssr-rehydration.md
divmain: That's a fair question.  To me, this incongruity is a natural outcome of marrying SSR (which happens outside the browser) with LWC (which "should eventually fade away to browser standards").

However, I hear you that having custom elements in the DOM without actually defining them is odd, and possibly not the right direction to go. This is a real use-case that we should either handle or explicitly _not_ support. I'll be interested to hear what other folks have to say.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+
+- a new section in the [LWC Guide](https://lwc.dev/guide/introduction), explaining the SSR-specific APIs and demonstrating how to get SSR up and running
+
+- an addition to the [LWC Recipes](https://recipes.lwc.dev/) or an example application utilizing SSR
+
+# Survey of Prior Art
+
+A handful of influential web frameworks and state management libraries were examined in relation to rehydration and adjacent concerns. The findings are summarized as follows:
+
+![0000-ssr-rehydration-fig3.png](../assets/0000-ssr-rehydration-fig3.png)
```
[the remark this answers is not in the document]
divmain: > I think that the survey should include end-to-end react frameworks, such as nextjs, blitzjs, remix, etc. The inclusion of data store libraries like redux/mobx doesn't add much value.

I'm happy to take another look. A lot of what these frameworks provide is out of scope for LWC - and live more naturally in LWR, for example. But if the survey missed some functionality that would require changes to LWC, I do want to capture that. If you can point me to anything you think might be important to consider, that'd be appreciated!

> VueJS SSR does not support streaming if the app relies on context data

I can add that caveat, thanks!

> React does rehydrate state, you just pass in that state as props to the base component. It just isn't opinionated on where you get that state.

I suppose that's true, although I think of that as being a solution outside of React itself.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+<figure>
+  <img src="../assets/0000-ssr-rehydration-fig2.png" alt="Fig. 2" width="60%">
+  <figcaption align = "center">
+    <b>Fig.2: the proposed implementation</b>
+  </figcaption>
+</figure><br /><br />
+
+In roughly chronological order with respect to a single page request, the following steps shall be taken as part of rehydration:
+
+- As part of SSR, serialize render-critical state and bake it into the HTML response returned from the server. For each `LightningElement`, attach corresponding state to its root DOM node as `data-ssr-state`.
```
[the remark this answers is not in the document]
divmain: Please correct me if I misunderstand, and perhaps the language here needs to be refined. A store could be serialized using whatever method desired on the server and rehydrated on the client before `hydrateElement` is called, outside of LWC altogether.

Are you looking for an LWC-provided mechanism to serialize and rehydrate state that is not defined with `@api` on a `LightningElement`?

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+
+The aim of rehydration is to avoid the client-side performance issues of a naive SSR implementation, reusing original DOM wherever possible.
+
+### The Happy Path
+
+With regard to SSR, there exists a happy path (Fig 1a) where the rendering output for a given request is identical on both the server and the client; identical VDOM is generated for the entire document in both environments.
+
+Depending upon the actual implementation, exceptions to this happy path include:
+
+- **Exception A:** One or more components are never rendered on the server (Fig 1b). Instead, a placeholder is rendered on the server, and later replaced on the client during CSR/hydration.
```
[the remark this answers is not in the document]
divmain: Yes, as noted in a previous comment, this needs to be rephrased. The client-side component is never rendered on the server, so it is true but misleading.

text/0000-ssr-rehydration.md
tedconn: >  Can you elaborate on how this is distinct from Exception A?

So there's really three scenarios we're discussing:

1. The DOM isn't the same and it's unexpected and we don't know what to do so we abort.
2. The DOM is intentionally different with the expectation that it's ok because it will rehydrate the web component.
3. The DOM is intentionally different because it branched its component logic by interrogating the render state (client vs server).

I think as long as, in all three scenarios we can recover from the mismatch and resume rehydration at a component level, then there's no distinction that needs to be made.

> Should there exist another lifecycle event for when a component is attached to pre-existing DOM during rehydration?

When the component is attached to pre-existing DOM wouldn't that effectively be the `render` phase? Thus, `renderedCallback` might be suitable?

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+After the proposed change, if a developer wants to take advantage of rehydration, a developer would instead do:
+
+```javascript
+import { hydrateElement } from 'lwc';
+import MyApp from './my-app';
+
+hydrateElement(MyApp, document.querySelector('#root')); 
+```
+
+### Server-only Component Support
```
pmdartus, line 105: It feels out of the scope for this first iteration of rehydration. What do you think about moving this section to a different proposal?

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+
+```javascript
+import { LightningElement } from 'lwc';
+import { isNodeEnv } from './my-utils';
+import tmplServerPlaceholder from './templateServer.html';
+import tmplClient from './templateClient.html';
+
+export default class ElementWithSSRPlaceholder extends LightningElement {
+    render() {
+        return this.isNodeEnv ? tmplServerPlaceholder : tmplClient;
```
pmdartus, line 199: How does this pattern address exception C? By using two different templates, there is a high chance the engine will be running in an unexpected mismatch.

For SSR placeholder components, I would envisioning the following pattern:

```js

export default class Placeholder extends LightningElement {
    showPlaceholder = true;

    connectedCallback() {
        if (!process.env.SSR) {
            setTimeout(() => this.showPlaceholder = false);
        }
    }
}
```

```html
<template>
    <template if:false={showPlaceholder}>
        <x-client-side-only></x-client-side-only>
    </template>

    <template if:true={showPlaceholder}>
        I am SSR placeholder
    </template>
</template>
```

By having the `showPlaceholder` property set to true by default, the placeholder will always be rendered on the server and the client. This avoids the unexpected mismatch during rehydration. After the component is rehydrated, the `showPlaceholder` property is set false in the following macro task to finally render the `<x-client-side-only>` element.

text/0000-ssr-rehydration.md
tedconn: I am not opposed to that. That is to say, we will definitely not require that for any MVP.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+    render() {
+        /*
+          On the server, the following line is transformed to:
+            return 'server' === 'server'
+          On the client, it is transformed to:
+            return 'client' === 'server'
+        */
+        return process.env.BUILD_TARGET === 'server'
+            ? tmplServerPlaceholder
+            : tmplClient;
```
nolanlawson, line 227: The convention I've seen for this in the past is [`process.browser`](https://www.npmjs.com/package/process#browser-implementation). There is also [the `"browser"` field in `package.json`](https://github.com/defunctzombie/package-browser-field-spec) and more recently [conditional exports](https://nodejs.org/api/packages.html#packages_conditional_exports).

It seems to me we should align on `process.browser` since it's no weirder than `process.env.NODE_ENV === 'production'` (which we already emit in our build). Also, anything that relies on `package.json` results in trickier code for authors to write (need two separate files for browser vs Node code), and for us to support in Rollup/Webpack/Jest/etc. If another convention has gained more traction or seems more appropriate, though (`import.meta.browser`?) then I'm happy to align on that instead.

Another alternative is to force users to rely on sniffing (e.g. [`browser-or-node`](https://www.npmjs.com/package/browser-or-node)), but the downside is that we can't strip it out in Rollup/minification.

text/0000-ssr-rehydration.md
nolanlawson: I'm trying to think of a case where `UnknownHTMLElement` would cause problems for a static component, but nothing comes to mind. Even `delegatesFocus` apparently has [a counterpart](https://github.com/mfreed7/declarative-shadow-dom/blob/a52e70f560e30c8aac2543396b1706e21445c2db/README.md#-behavior) in declarative shadow DOM.

text/0000-ssr-rehydration.md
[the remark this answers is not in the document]
nolanlawson: I would be interested to know what Lit is up to. They are currently implementing SSR, and unlike the other frameworks listed here, they are web component-based like LWC.

text/0000-ssr-rehydration.md
divmain: > As far as I understand how rehydration work in other frameworks, the 2 examples above will make the framework bails out and re-render the page from scratch. Minor differences can be recovered, like text change. But DOM structural changes can't be recovered during rehydration because it would require a full DOM diffing.

There is some variation across frameworks for how this is handled and it has changed over time. Historically, for example, React was resilient to mismatches during SSR rehydration. React 16 changed that behavior for performance reasons.

However, this isn't a binary thing (resilient vs not resilient, full DOM diffing vs complete bailtout) and we don't have to treat it that way. When considering how mismatches are handled during rehydration, the following options come to mind:

1. Bail out of rehydration entirely when a mismatch is detected, throw away the original HTML, and force a full re-render.
2. Handle easy mismatches (text nodes, certain attrs, etc.) falling back to number 1 for anything more extreme.
3. Bail out of rehydration for subtrees, rooted at a mismatch.
4. Bail out of rehydration for subtrees, rooted at the component where the mismatch occurred.
5. Try to repair as much of the SSR-derived DOM as possible, throwing nothing away explicitly.
6. Assume no mismatches exist, and break silently.

When you refer to full DOM diffing, I interpret that to mean #5. However, we could choose #4, and may get good mileage out of that approach.

So what I'm thinking is:

- We detect a mismatch
- We determine if that mismatch is easily recoverable (e.g. we have allowances for text-node mismatches)
- If it is not easily recoverable, we determine the closest LWC ancestor to that mismatch (anywhere from the root component in a SPA to a leaf component for a small widget).
- We invalidate the ancestor's subtree, replacing SSR-derived DOM with CSR DOM _for that component subtree only_.

It seems that mismatches in LWC will be either A) component-level mismatch due to non-serializable state, B) ephemeral state (e.g. current time in a text node), or C) something broke unexpectedly and we need to recover.

The approach I'm suggesting would handle case A in a reasonable way most times, without incurring horrible performance degredation. If the state for the root component of a huge tree is non-serializable, the action taken would be be roughly comparable to a full bailout, which is probably the best we can do, anyway.

If we make allowances for "easy" mismatches, case B is handled without any notable cost.

The best approach for case C is difficult to know without seeing how SSR is utilized in the real world. But, worst case, you bail out at the root level. In the best case, you bail out for part of the application, minimizing the performance impact.

divmain: > > Should there exist another lifecycle event for when a component is attached to pre-existing DOM during rehydration?

> Ideally, developers shouldn't have to differentiate between pure client-side rendering and rehydration. @divmain Do you have a use case in mind?

I have seen developers commonly abuse lifecycle events, so I may be overly cautious here. But let's say a developer wants to create a websocket connection to some backend, and they want that connection attached to an instance of a component.

`connectedCallback` makes sense for that. But in a non-SSR world, a component won't have any child elements when `connectedCallback` is invoked. In an SSR world, it would.

I can't think of an example of where that _should_ matter, but I'm thinking [Hyrum's Law](https://www.hyrumslaw.com/) may apply.

text/0000-ssr-rehydration.md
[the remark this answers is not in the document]
divmain: Thanks @nolanlawson! I will take a look.

text/0000-ssr-rehydration.md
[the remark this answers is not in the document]
divmain: I don't think there's a great answer to this - it is a headache when doing SSR regardless of the framework. Explicit documentation can help. Build tooling (linting, type checking) can help. I'm open to ideas here, for sure.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+  </figcaption>
+</figure><br /><br />
+
+In roughly chronological order with respect to a single page request, the following steps shall be taken as part of rehydration:
+
+- As part of SSR, serialize render-critical state and bake it into the HTML response returned from the server. For each `LightningElement`, attach corresponding state to its root DOM node as `data-ssr-state`.
+
+- Rather than calling `LWC#createElement`, required that `LWC#hydrateElement` is called to mount the root Lightning Web Component on the client.
+
+- Generate server-side VDOM from the SSR HTML on the client.
```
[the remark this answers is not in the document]
divmain: Looking again at snabbdom documentation, this step is unnecessary. I'll make the edit.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+
+- As the initial client-side render progresses, extract state from `data-ssr-state` attributes, instantiate `LightningElements` with extracted state, and generate client-side VDOM.
```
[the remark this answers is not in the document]
divmain: My thought is that it could be present on any component in the subtree. The other approach with the same aim is to extract a state tree that mirrors the component tree during SSR, and serialize state in a single place. However, serializing the state at the root node of each component means the state is easily accessible during traversal in rehydration, which is attractive from the implementation side. Open to feedback here, if you think there are pitfalls I haven't accounted for.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,299 @@
...
+While the initial patch progresses, each SSR VDOM node must be checked for a `data-ssr-preserve` boolean attribute. When identified, the SSR VDOM subtree rooted at that node must be preserved for subsequent client-side renders. Frameworks built on top of LWC can utilize this feature to provide server-only components.
+
+<details open>
+  <summary><b>Example:</b></summary>
+
+An LWC-based framework could provide a new `LightningServerElement` class. Its behavior and underlying implementation could be switched out at build time, as a function of the intended runtime environment.
+
+On the server, a `LightningServerElement` could behave identically to a `LightningElement` with one exception: it would attach a `data-ssr-preserve` attribute to its root node.
+
+For a client build, a Babel plugin could identify any class declaration that extends `LightningServerElement` and replace the entire declaration with `NoopElement`.  That would look something like the following:
```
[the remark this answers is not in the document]
divmain: Yes, this would be entirely independent of LWC, really. But documentation/ideas/examples will be important to answer: "How do I make "server components" in LWC like React has?"

text/0000-ssr-rehydration.md
divmain: I'm happy to push this to a different proposal.

text/0000-ssr-rehydration.md
divmain: I will re-word this. I wasn't intending to say that this approach solves exception C. I was saying that if we solve for exception C, implementing placeholder components (exception A) becomes straightforward.

text/0000-ssr-rehydration.md
jodarove: Should we include a section on handling mismatches (exception c), including spec-like steps?

> As far as I understand how rehydration work in other frameworks, the two examples above will make the framework bails out and re-render the page from scratch. Minor differences can be recovered, like text change. But DOM structural changes can't be recovered during rehydration because it would require a full DOM diffing.

I like this alternative because we do the checks at dev time, bailing out the entire dom being hydrated. For prod, we could assume everything is correct and just hydrate without any verification, recovering minor differences.

> So what I'm thinking is:
We detect a mismatch
We determine if that mismatch is easily recoverable (e.g. we have allowances for text-node mismatches)
If it is not easily recoverable, we determine the closest LWC ancestor to that mismatch (anywhere from the root component in a SPA to a leaf component for a small widget).
We invalidate the ancestor's subtree, replacing SSR-derived DOM with CSR DOM for that component subtree only.

I also consider this attractive: it allows us to reuse more from SSR DOM. But, If we do it only on dev, it is throwaway code because perf is not as important, so the first approach is as good as this one. The benefit may be when used in prod, but there's a perf hit of "mismatch detection" when compared to the previous solution.

jodarove: Open question: How to recover from a lifecycle error in a component while hydrating? (ex: an error on the connected callback)

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+<figure>
+  <img src="../assets/0000-ssr-rehydration-fig2.png" alt="Fig. 2" width="60%">
+  <figcaption align = "center">
+    <b>Fig.2: the proposed implementation</b>
+  </figcaption>
+</figure><br /><br />
+
+In roughly chronological order with respect to a single page request, the following steps shall be taken as part of rehydration:
+
+- As part of SSR, serialize render-critical state and bake it into the HTML response returned from the server. For each `LightningElement`, attach corresponding state to its root DOM node as `data-ssr-state`.
```
diervo, line 67: What is `state`? What will be serializable? How it will look like?

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+
+Alternately, if it is desirable _not_ to ship the placeholder in the client-side bundle, one could utilize a Babel transform or `@rollup/plugin-replace` to make the necessary changes at build time.
+
+<details open>
+  <summary>Example:</summary>
+
+```javascript
+import { LightningElement } from 'lwc';
+import { isNodeEnv } from './my-utils';
+// ↶ removed from client builds during minification
```
diervo, line 199: Let's not get into this as it will have non-trivial implications on runtime lifecycle (ex. LWR)

text/0000-ssr-rehydration.md
diervo: Where that data-ssr-state comes from? Is it user-land driven? Or is the compiler producing it?

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+State that is contained outside of a LWC (e.g. in a store of some kind) will need to be serialized manually during SSR and rehydrated prior to invocation of `hydrateElement` on the client.
+
+### Bailing out
+
+There is a spectrum of choices possible when encountering an unexpected mismatch during rehydration. The following is a sampling from that spectrum:
+
+1. Bail out of rehydration entirely when a mismatch is detected, throw away the original HTML, and force a full re-render.
+2. Handle easy mismatches (text nodes, certain attrs, etc.) falling back to number 1 for anything more extreme.
+3. Bail out of rehydration for subtrees, rooted at a mismatch.
+4. Bail out of rehydration for subtrees, rooted at the component where the mismatch occurred.
```
diervo, line 101: Later on you say that the way a user can opt-in/out-out to have a different template on Client vs Server is in user land by swapping the whole template. I think it will be interesting to also put an example where the granularity is at the statement level. (ex. `<template if:true={inServer}>FOO<else>BAR`). How how that would affect the bailing out decision.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+After the proposed change, if a developer wants to take advantage of rehydration, a developer would instead do:
+
+```javascript
+import { hydrateElement } from 'lwc';
+import MyApp from './my-app';
+
+hydrateElement(MyApp, document.querySelector('#root')); 
+```
+
+### Constructable stylesheets
```
diervo, line 149: IMHO we should have waited to introduce constructible stylesheets (as we were hiding the internal anyway) if this is creating a path of resistance for hydration. But I guess "Everything is obvious once you know the answer"

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+1. Swap in constructable stylesheets for vanilla stylesheets.
+2. Permanenly disable constructable stylesheets for the lifetime of the `document`.
+
+It is believed that swapping in constructable stylesheets could obviate the performance wins that constructable stylesheets are intended to bring. Furthermore, any theoretical benefit is outweighed by the implementation complexity of swapping in constructable stylesheets.
+
+For these reasons, it is proposed that constructable stylesheets be disabled for the lifetime of the `document` in a post-rehydration context.
+
+### Server-only components
+
+Addressing [Exception B to the happy path](#the-happy-path) will involve a solution roughly comparable to [React server components](https://reactjs.org/blog/2020/12/21/data-fetching-with-react-server-components.html). While the bulk of this can be implemented in user-land, at a level of abstraction above LWC, the `hydrateElement` patch implementation will need to support the use-case.
```
diervo, line 164: What I'm really concerned here is on who has the saying on what parts are rehydratable or not. My intuition tells me that we might want to have some lever at the application level...

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+
+```javascript
+import { LightningElement } from 'lwc';
+import { isNodeEnv } from './my-utils';
+import tmplServerPlaceholder from './templateServer.html';
+import tmplClient from './templateClient.html';
+
+export default class ElementWithSSRPlaceholder extends LightningElement {
+    render() {
+        return this.isNodeEnv ? tmplServerPlaceholder : tmplClient;
```
diervo, line 185: I know this is pseudo-code, but I think it will be important to understand who is responsible for providing the host-specific environment setup. I do think we need to spec it out (plus its invariants - like should be immutable), and it should not be "user-land" defined.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+  <summary>Example:</summary>
+
+```javascript
+import { LightningElement } from 'lwc';
+import { isNodeEnv } from './my-utils';
+// ↶ removed from client builds during minification
+import tmplServerPlaceholder from './templateServer.html';
+import tmplClient from './templateClient.html';
+
+export default class ElementWithSSRPlaceholder extends LightningElement {
```
diervo, line 203: Per my previous comment I would add an example of more fine grained condition wihin the template (like wit ha getter) and see how that reconciles with the bailing out mechanism

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,266 @@
...
+
+For performance reasons, Vue will also not repair mismatching HTML during rehydration. However, this is less of an issue in practice, because Vue will take a snapshot of state and ship it alongside the rendered application as `window.__INITIAL_STATE__`.
+
+Most framework authors have indicated that non-support for rehydration resilience is due to performance reasons. We should measure the performance impact of SSR rehydration in a large LWC app to either 1) validate our design decision, or 2) point us in a new direction.
+
+Additionally, if we determine that the performance impact is too great, a solution will need to be devised to address [Exception A to the happy path](#the-happy-path).
+
+# Unresolved questions
+
+- **How do we handle invocations of `customElements.define` with SSR-serialized custom elements already in the DOM?**
```
diervo, line 231: The whole invariant around what happens with dynamic or lazy custom elements should be resolved. At least define an ambiguous behaviour

text/0000-ssr-rehydration.md
divmain: I've clarified that this is an example, only intended to show what is possible, not to prescribe a particular implementation.

.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
jodarove, line 1: since this is dependant on your environment, ill put this as a global .gitignore in your local setup.

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,271 @@
+---
+title: SSR Rehydration
+status: DRAFTED
+created_at: 2021-08-31
+updated_at: 2021-08-31
+pr: https://github.com/salesforce/lwc-rfcs/pull/53
```
jodarove, line 6: i believe this field should point to an implementation PR.

```suggestion
pr: https://github.com/salesforce/lwc/pull/2442
```

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Server-only components
+
+Addressing [Special Case B](#the-happy-path) will involve a solution roughly comparable to [React server components](https://reactjs.org/blog/2020/12/21/data-fetching-with-react-server-components.html). While the bulk of this can be implemented in user-land, at a level of abstraction above LWC, the `hydrateElement` patch implementation will need to support the use-case.
+
+However, we consider support for this case to be out-of-scope for the initial implementation. We will explore further as part of a separate RFC.
+
+### SSR placeholders
+
+Implementation of placeholder components, as described in [Special Case A](#the-happy-path) should require no extra effort, so long as unexpected subtree mismatches (Special Case C) are dealt with properly.
```
jodarove, line 162: another alternative that won't generate mismatches is to rely on the fact that on SSR, the `renderedCallback` is not invoked, ex:

```
<template>
    <template if:true={isFirstRender}>
        <p>content placeholder on the server</p>
    </template>
    <template if:false={isFirstRender}>
        <p>rendered only on the client</p>
    </template>
</template>

export default class ClientOnly extends LightningElement {
    isFirstRender = true;
    renderedCallback() {
        this.isFirstRender = false;
    }
}
```

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,271 @@
...
+The introduction of constructable stylesheets to LWC presents a small wrinkle. Since support for this browser API is not yet ubiquitous, we cannot know during SSR whether it will be supported on the client. For that reason, constructable stylesheets are disabled during SSR.
+
+From there, we have two options if support for constructable stylesheets is detected during rehydration:
+
+1. Swap in constructable stylesheets for vanilla stylesheets.
+2. Permanenly disable constructable stylesheets for the lifetime of the `document`.
+
+It is believed that swapping in constructable stylesheets could obviate the performance wins that constructable stylesheets are intended to bring. Furthermore, any theoretical benefit is outweighed by the implementation complexity of swapping in constructable stylesheets.
+
+For these reasons, it is proposed that constructable stylesheets be disabled for the lifetime of the `document` in a post-rehydration context.
```
jodarove, line 152:
```suggestion
For these reasons, it is proposed that constructable stylesheets be disabled during hydration, and during the lifetime of the component instance.
```

text/0000-ssr-rehydration.md
```diff
@@ -0,0 +1,271 @@
...
+3. Any props that should be supplied to that root component as part of the initial client-side render.
+
+### Constructable stylesheets
+
+The introduction of constructable stylesheets to LWC presents a small wrinkle. Since support for this browser API is not yet ubiquitous, we cannot know during SSR whether it will be supported on the client. For that reason, constructable stylesheets are disabled during SSR.
+
+From there, we have two options if support for constructable stylesheets is detected during rehydration:
+
+1. Swap in constructable stylesheets for vanilla stylesheets.
+2. Permanenly disable constructable stylesheets for the lifetime of the `document`.
```
jodarove, line 148:
```suggestion
2. Disable constructable stylesheets during hydration, for the lifetime of the component instance.
```

Explanation: The stylesheets are created, and cached in the VM the first time the VM renders ([code](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/template.ts#L144-L185)).

PR #56, opened 2021-09-30
Programmatic style injection

pmdartus:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/pmdartus/css-extensibility/text/0000-programmatic-style-injection.md)

> This proposal describes how LWC components can programmatically associate CSS stylesheets to an LWC component.

### DIFF
--- a/text/0000-programmatic-style-injection.md
+++ b/text/0000-programmatic-style-injection.md
@@ -133,9 +133,11 @@ An alternative to exporting LWC stylesheet module as an opaque object is to have
 
 As mentioned above, the LWC compiler implicitly loads the stylesheet associated with each template. Internally, the template compiler generates an import statement with the template file name and the `.css` extension. Since stylesheets are optional, the LWC module resolver currently ignores non-existent stylesheet imports.
 
-This behavior is problematic for explicit stylesheet imports as the compilation doesn't fail if the imported stylesheet can't be resolved. To solve this issue, the LWC compiler should treat component's implict stylesheet import as optional and other explicit stylesheet imports as required. For DX improvement, LWC compiler should fail compilation or at least warn developers if an explicit stylesheet import cannot be resolved before treating non-existent stylesheet imports as empty imports. 
+This behavior is problematic for explicit stylesheet imports as the compilation doesn't fail if the imported stylesheet can't be resolved. To solve this issue, the LWC compiler should treat component's implicit stylesheet import as optional and other explicit stylesheet imports as required. For DX improvement, LWC compiler should fail compilation or at least warn developers if an explicit stylesheet import cannot be resolved before treating non-existent stylesheet imports as empty imports. 
 
-An alternative solution is that a new `optional` query string can be added to the import identifier to indicate whether or not the LWC compiler should report an error when the file is missing.
+A straightforward implementation is to check whether a stylesheet import is implicit by checking if the importee's path is the same as importer's path. 
+
+Another possible implementation is that a new `optional` query string can be added to the import identifier to indicate whether or not the LWC compiler should report an error when the file is missing.
 
 ```js
 // Required LWC stylesheet import. Fails during compilation when missing.
--- a/text/0000-programmatic-style-injection.md
+++ b/text/0000-programmatic-style-injection.md
@@ -137,6 +137,14 @@ import styles from "./styles.css?optional=true";
 
 This query string should only be used by the LWC template compiler to differentiate between the optional stylesheet imported from templates and the required stylesheet explicitly imported by developers.
 
+The import assertion proposal leaves the the door open for [potential future extensions](https://github.com/tc39/proposal-import-assertions#follow-up-proposal-evaluator-attributes) to avoid bloating the query string.
+
+```js
+// This syntax was discussed as part of the import assertion proposal and might be added in a
+// follow-up proposal.
+import styles from "./styles.css" assert { type: "lwc/css" } with { optional: true };
+```
+
 ## Drawbacks
 
 With this proposal, developers have now the capability to extends any other component to inject their own styles. This is an issue if those components aren't meant to be extended. 
--- a/text/0000-programmatic-style-injection.md
+++ b/text/0000-programmatic-style-injection.md
@@ -1,0 +1,158 @@
+---
+title: Programmatic style injection
+status: DRAFTED
+created_at: 2021-09-16
+updated_at: 2021-09-30
+pr: (leave this empty until the PR is created)
+---
+
+# Programmatic style injection
+
+## Summary
+
+This proposal describes how LWC components can programmatically associate CSS stylesheets to an LWC component.
+
+## Basic example
+
+```js
+import { LightningElement } from 'lwc';
+import myStyles from './styles.css';
+
+export default Example class extends LightningElement {
+    static stylesheets = [
+        myStyles
+    ];
+}
+```
+
+## Motivation
+
+CSS custom properties and CSS `::part()` are the standard mechanisms by which applications can influence the look and feel of shadow DOM components. Component authors are required to explicitly define those extension points that can be styled from outside the shadow tree. While this approach gives a certain level of control over the component appearance, in certain cases developers might want to take full control over the styles injected in the shadow tree. And this without compromising the component encapsulation. 
+
+Component libraries are a prime example of this use case. Component authors want to expose high-quality base components with minimal styles and let application developers add their branding.
+
+LWC implicit template and stylesheet resolution system makes it impossible today to extend from a base LWC component and override its styles at the same time. This proposal addresses this issue by exposing a mechanism to associate stylesheets directly with the LWC component class. 
+
+## Detailed design
+
+A new `stylesheets` static property is added on the `LightningElement` constructor. This property allows developers to define which stylesheets the component should use. It accepts an array of [LWC stylesheets](#lwc-stylesheets-modules). The base `LightningElement.stylesheets` static property returns an empty array.
+
+The `stylesheets` property works on both light DOM and shadow DOM LWC components. This stylesheet injection mechanism complements the existing implicit template/stylesheet resolution mechanism.
+
+The LWC engine caches the static `stylesheets` value for the lifetime of the application during the component class definition. Modifying the `stylesheets` property after the component code is evaluated will not affect the stylesheet injected by this component. 
+
+For component inheritance, the LWC engine doesn't attempt to merge ancestor components stylesheets. Developers are in charge of merging (or not) the ancestors' stylesheets with their stylesheets.
+
+```js
+import { LightningElement } from 'lwc';
+import ancestorStylesheet from './ancestor.css';
+import childStylesheet from './child.css';
+
+class Ancestor extends LightningElement {
+    static stylesheets = [
+        ancestorStylesheet
+    ];
+}
+
+class Descendent extends Ancestor {
+    static stylesheets = [
+        ...super.stylesheets, 
+        childStylesheet
+    ];
+}
+```
+
+### Injection order
+
+The LWC engine injects stylesheets in the following order: first, the implicit stylesheet associated with the template and then all the stylesheets associated with the `stylesheets` property on the component class. This order allows the styles that are programmatically associated with the component to override the styles that are implicitly loaded via the template.
+
+In the following example the `<x-foo>` stylesheets will be injected in the following order: `foo.css`, `custom-a.css` and `custom-b.css`.
+
+```
+x/
+└── foo/
+    ├── custom-a.css
+    ├── custom-b.css
+    ├── foo.js
+    ├── foo.html
+    └── foo.css
+```
+
+```js
+// x/foo.js
+import { LightningElement } from 'lwc';
+import customStyleA from './custom-a.css';
+import customStyleB from './custom-b.css';
+
+export default Foo class extends LightningElement {
+    static stylesheets = [
+        customStyleA,
+        customStyleB
+    ];
+}
+```
+
+### LWC stylesheets modules
+
+Programmatically importing a stylesheet module from JavaScript using an import declaration is not a new concept introduced by this RFC. Such import is already supported by the LWC compiler. In fact, the template compiler is already using this mechanism to import the stylesheet implicitly associated with the template.
+
+However, until now, the compile-time behavior and runtime behavior for stylesheet modules have never been fully specified. This section fills this specification gap.
+
+Stylesheets modules are identifiable by their `.css` extension. This kind of module only exposes a single default export. The exported value is an opaque object, in other words, the internal implementation of the stylesheet is not exposed to developers. Keeping the exported value opaque is necessary until all the browser vendors agree on [Constructable Stylesheets](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet) implementation.
+
+```js
+import stylesheetA from './styles-a.css';
+import stylesheetB from './styles-b.css';
+```
+
+#### Comparison with CSS module script
+
+LWC stylesheet modules work similarly to native [CSS module script](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/css-modules-v1-explainer.md). This feature isn't fully standardized yet and is today only implemented in Chromium-based browsers. There are few differences between the two proposals worth calling out:
+- LWC stylesheet modules can be imported using a standard import statement. The LWC compiler infers from the `.css` file extension that the file has to be treated as a CSS module. On the other hand, CSS module scripts require the use of [import assertions](https://github.com/tc39/proposal-import-assertions).
+- LWC stylesheet modules default export is an opaque object. CSS module scripts default export is an instance of [`CSSStyleSheet`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet).
+
+```js
+// LWC stylesheet modules
+import styles from "./styles.css";
+console.log(styles + '') // [object LWCCSSStyleSheet]
+
+// CSS modules
+import styles from "./styles.css" assert { type: "css" };
+console.log(styles + '') // [object CSSStyleSheet]
+```
+
+#### LWC stylesheet import enforcement
+
+As mentioned above, the LWC compiler implicitly loads the stylesheet associated with each template. Internally, the template compiler generates an import statement with the template file name and the `.css` extension. Since stylesheets are optional, the LWC module resolver currently ignores non-existent stylesheet imports.
+
+This behavior is problematic for explicit stylesheet imports as the compilation doesn't fail if the imported stylesheet can't be resolved. To solve this issue a new `optional` query string can be added to the import identifier to indicate whether or not the LWC compiler should report an error when the file is missing.
+
+```js
+// Required LWC stylesheet import. Fails during compilation when missing.
+import styles from "./styles.css";
+
+// Optional LWC stylesheet import. Resolve to an empty stylesheet if missing.
+import styles from "./styles.css?optional=true";
+```
+
+This query string should only be used by the LWC template compiler to differentiate between the optional stylesheet imported from templates and the required stylesheet explicitly imported by developers.
+
+## Drawbacks
+
+With this proposal, developers have now the capability to extends any other component to inject their own styles. This is an issue if those components aren't meant to be extended. 
+
+## Alternatives
+
+TBD
+
+## Adoption strategy
+
+This proposal solves a narrow use-case. It is not expected for this API to be widely adopted.
+
+# How we teach this
+
+A new section covering this topic will be added to the documentation.
+
+# Unresolved questions
+
+None
--- a/text/0000-programmatic-style-injection.md
+++ b/text/0000-programmatic-style-injection.md
@@ -98,11 +98,17 @@ Programmatically importing a stylesheet module from JavaScript using an import d
 
 However, until now, the compile-time behavior and runtime behavior for stylesheet modules have never been fully specified. This section fills this specification gap.
 
-Stylesheets modules are identifiable by their `.css` extension. This kind of module only exposes a single default export. The exported value is an opaque object, in other words, the internal implementation of the stylesheet is not exposed to developers. Keeping the exported value opaque is necessary until all the browser vendors agree on [Constructable Stylesheets](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet) implementation.
+Stylesheets modules are identifiable by their `.css` extension. This kind of module only exposes a single default export. The exported value is an opaque object, in other words, the internal implementation of the stylesheet is not exposed to developers. Developers shouldn't rely on the shape of those objects as it may change.
 
 ```js
 import stylesheetA from './styles-a.css';
 import stylesheetB from './styles-b.css';
+```
+
+Scoped stylesheet modules work similarly than standard stylesheet modules. They can be imported using the `.scoped.css` extension.
+
+```js
+import scopedStylesheet from './styles.scoped.css';
 ```
 
 #### Comparison with CSS module script
--- a/text/0000-programmatic-style-injection.md
+++ b/text/0000-programmatic-style-injection.md
@@ -127,11 +127,15 @@ import styles from "./styles.css" assert { type: "css" };
 console.log(styles + '') // [object CSSStyleSheet]
 ```
 
+An alternative to exporting LWC stylesheet module as an opaque object is to have a utility function `registerStylesheet` that registers compiled stylesheets so that component authors cannot modify or replace stylesheet functions. This would be a similar approach on component templates ([`isTemplateRegistered`](https://github.com/salesforce/lwc/blob/1b85b4606f8cc8a18c368d9ce3b9c3616043f10c/packages/@lwc/engine-core/src/framework/secure-template.ts#L17)).
+
 #### LWC stylesheet import enforcement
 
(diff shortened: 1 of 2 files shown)


text/0000-programmatic-style-injection.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+### LWC stylesheets modules
+
+Programmatically importing a stylesheet module from JavaScript using an import declaration is not a new concept introduced by this RFC. Such import is already supported by the LWC compiler. In fact, the template compiler is already using this mechanism to import the stylesheet implicitly associated with the template.
+
+However, until now, the compile-time behavior and runtime behavior for stylesheet modules have never been fully specified. This section fills this specification gap.
+
+Stylesheets modules are identifiable by their `.css` extension. This kind of module only exposes a single default export. The exported value is an opaque object, in other words, the internal implementation of the stylesheet is not exposed to developers. Keeping the exported value opaque is necessary until all the browser vendors agree on [Constructable Stylesheets](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet) implementation.
```
nolanlawson, line 101: This implies that someday the opaque object will be a constructable stylesheet. I'm not sure we can ever get to that future, due to light DOM scoped CSS and the need to evaluate the scope token at runtime, as well as `:host`, which varies based on the context of light vs shadow.

I would just say it's an opaque object, and developers shouldn't rely on the shape of that object as it may change. Either that, or we should make it truly opaque in a way that is only accessible to the engine.

text/0000-programmatic-style-injection.md
```diff
@@ -0,0 +1,158 @@
...
+
+This behavior is problematic for explicit stylesheet imports as the compilation doesn't fail if the imported stylesheet can't be resolved. To solve this issue a new `optional` query string can be added to the import identifier to indicate whether or not the LWC compiler should report an error when the file is missing.
+
+```js
+// Required LWC stylesheet import. Fails during compilation when missing.
+import styles from "./styles.css";
+
+// Optional LWC stylesheet import. Resolve to an empty stylesheet if missing.
+import styles from "./styles.css?optional=true";
+```
```
nolanlawson, line 136: Not sure if we should call this out as an implementation detail or not. Potentially there are other ways we could solve this problem in the future. (Fancier imports, for instance, `import styles from "./styles.css" assert { type: "lwc/css" } with { optional: true }`.)

nolanlawson (changes requested):
This LGTM, my only feedback is that we should make some mention of how scoped styles interact with this. I assume they behave exactly the same as normal template styles, but it's worth calling out.

pmdartus:
FYI @brandonferrua

jodarove (approved):

brandonferrua:
Any progress on this RFC? Would love to take advantage of this feature

nolanlawson:
@brandonferrua It made the line for 242.

text/0000-programmatic-style-injection.md
```diff
@@ -0,0 +1,172 @@
...
+
+LWC implicit template and stylesheet resolution system makes it impossible today to extend from a base LWC component and override its styles at the same time. This proposal addresses this issue by exposing a mechanism to associate stylesheets directly with the LWC component class. 
+
+## Detailed design
+
+A new `stylesheets` static property is added on the `LightningElement` constructor. This property allows developers to define which stylesheets the component should use. It accepts an array of [LWC stylesheets](#lwc-stylesheets-modules). The base `LightningElement.stylesheets` static property returns an empty array.
+
+The `stylesheets` property works on both light DOM and shadow DOM LWC components. This stylesheet injection mechanism complements the existing implicit template/stylesheet resolution mechanism.
+
+The LWC engine caches the static `stylesheets` value for the lifetime of the application during the component class definition. Modifying the `stylesheets` property after the component code is evaluated will not affect the stylesheet injected by this component.
```
nolanlawson, line 42: @pmdartus It's unclear to me whether dynamically modifying the `stylesheets` _array_ is supported or not:

```js
this.constructor.stylesheets = [foo] // definitely not supported

this.constructor.stylesheets.push(foo) // supported?
```

I'd rather go ahead and say that the entire array is frozen (i.e. throws an error on modification).

However if we say this, then I guarantee some people are going to want to dynamically modify their stylesheets on-demand. 🙂 And sadly people have found a (hacky) way to do this, which (AFAICT) this proposal does not cover.

text/0000-programmatic-style-injection.md
```diff
@@ -98,13 +98,19 @@ Programmatically importing a stylesheet module from JavaScript using an import d
...
 
 ```js
 import stylesheetA from './styles-a.css';
 import stylesheetB from './styles-b.css';
 ```
 
+Scoped stylesheet modules work similarly than standard stylesheet modules. They can be imported using the `.scoped.css` extension.
+
+```js
+import scopedStylesheet from './styles.scoped.css';
```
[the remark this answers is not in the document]
nolanlawson: @pmdartus Re-evaluating this: I think approach # 2 would be problematic for making scoped styles default in light DOM (https://github.com/salesforce/lwc-rfcs/pull/65).

If scoped styles are the default for light DOM, then `foo.css` means "scoped," and the only way to communicate that is with the query param. So I think we should instead keep the query param, and transform JS files to append `?scoped=true` to imports of `*.scoped.css`. The fact that the import occurred in a user-authored JS file is our signal that the file should be treated as scoped.

Unfortunately, there is now a weird situation like this:

```
- foo.js
- foo.css
- foo.html (light DOM)
```

In the above scenario, `foo.css` is treated as scoped when imported by `foo.html`, and treated as unscoped when imported by `foo.js`.

Maybe this isn't such a big deal, though, because people will probably not name it `foo.css` alongside `foo.html` and then import from `foo.js`? If they're doing programmatic style injection, they can name the CSS file whatever they want, because it's not implicitly tied to the HTML file.

/cc @divmain

forcePanda:
Hi @nolanlawson,
Is this already available on the platform? I'm trying to use this but no success.

I'm referencing this [doc](https://github.com/salesforce/lwc-rfcs/blob/pmdartus/css-extensibility/text/0000-programmatic-style-injection.md)

My component structure:
template1.html
template2.html
cmp.js
cmp.html
cmp.js-meta.xml
util.css

cmp.js
`import utilCSS from './util.css';

export default class CMP extends LightningElement {

    static stylesheets = [
        utilCSS
    ];
}
`

pmdartus:
> Is this already available on the platform? I'm trying to use this but no success.

@forcePanda This feature is still at the draft RFC (request for comments) stage. It is yet to be implemented in LWC if/when we are able to reach a consensus.

forcePanda:
> > Is this already available on the platform? I'm trying to use this but no success.
> 
> @forcePanda This feature is still at the draft RFC (request for comments) stage. It is yet to be implemented in LWC if/when we are able to reach a consensus.

Ah fish! I had a good use case for it. Thanks for the update @pmdartus. 
Much appreciated! :)

text/0000-programmatic-style-injection.md
```diff
@@ -0,0 +1,172 @@
...
+    static stylesheets = [
+        ...super.stylesheets, 
+        childStylesheet
+    ];
+}
+```
+
+### Injection order
+
+The LWC engine injects stylesheets in the following order: first, the implicit stylesheet associated with the template and then all the stylesheets associated with the `stylesheets` property on the component class. This order allows the styles that are programmatically associated with the component to override the styles that are implicitly loaded via the template.
```
nolanlawson, line 67: In the case of multi-template components, we cannot guarantee that the implicit stylesheets associated with the template apply before the stylesheets from the `stylesheets` property on the component class. This is because we never remove templates and instead only append them.

So for instance, if you have `static stylesheets = [X]` and you `render()` template A with stylesheet A, then later `render` template B with stylesheet B, you would get this ordering:

1. `A`
2. `X`
3. `B`

To me, it is more consistent and predictable if instead the ordering were:

1. `X`
2. `A`
3. `B`

However, if we change this ordering, then it would no longer be true that you can "override the styles that are implicitly loaded via the template." So I am a bit torn on this.

text/0000-programmatic-style-injection.md
nolanlawson: Thinking back to [how component authors are able to dynamically inject stylesheets today using a workaround](https://github.com/salesforce/lwc-test/pull/147), it occurs to me that the current RFC does not actually solve this use case, because it's not possible to modify stylesheets dynamically during the component lifecycle. So component authors will still be tempted to use the workaround.

I am beginning to wonder if it makes sense to have a slightly different design. Something like:

```js
render() {
  return {
    template: template,
    stylesheets: [stylesheetX, stylesheetY]
  }
}
```

In the above case, the `render()` function could return either 1) a template, as is done today, or 2) an object with a `template` property and and optional `stylesheets` property.

In the above design, we would not need to specify stylesheet ordering, because it would work exactly the same as the current `render()` function works today. If a stylesheet is not injected, it is injected on-demand.

OTOH, it would not solve the problem of multi-template components that render A, then B, then A. (The stylesheets from B would still have been injected after A, so B will take precedence.) To fix that, we would need to overhaul how stylesheet injection is done (either [remove stylesheets as we attempted once](https://github.com/salesforce/lwc/pull/2797), or bump stylesheets later in the list when they are rendered).

nolanlawson (changes requested):
After giving this a re-read, I am concerned about how to handle multi-template components. I'm also concerned that this RFC doesn't fully replace the [current hacks](https://github.com/salesforce/lwc-test/pull/147) that component authors are using. I would like to get some more feedback, especially from folks like @brandonferrua and @gonzalocordero, on the issues I raised above.

text/0000-programmatic-style-injection.md
brandonferrua: > In the case of multi-template components, we cannot guarantee that the implicit stylesheets associated with the template apply before the stylesheets from the `stylesheets` property on the component class. This is because we never remove templates and instead only append them.
> 
> So for instance, if you have `static stylesheets = [X]` and you `render()` template A with stylesheet A, then later `render()` template B with stylesheet B, you would get this ordering in the DOM:
> 
> 1. `A`
> 2. `X`
> 3. `B`
> 
> To me, it is more consistent and predictable if instead the ordering were:
> 
> 1. `X`
> 2. `A`
> 3. `B`
> 
> However, if we change this ordering, then it would no longer be true that you can "override the styles that are implicitly loaded via the template." So I am a bit torn on this.
> 
> As I note [here](https://github.com/salesforce/lwc-rfcs/pull/56#discussion_r981693420), there is also the issue that the order will never change on subsequent updates. (E.g. if we re-`render()` template A with stylesheet A, then none of the stylesheets in the DOM will change.)

If I'm thinking of the cascade, I think we want this order:

1. `A`
2. `B`
3. `X`

Provided the following example, i'd expect `X` stylesheet to to be apply last in the order:
```css
/* Template A CSS */
.a {
  color: var(--a-color);
}
```

```css
/* Template B CSS */
.b {
  color: var(--b-color);
}
```

```css
/* Template X CSS */
--a-color: red;
--b-color: blue;
```

Result is the `color` of `.a` is `red` and `.b` is `blue`.

nolanlawson: @brandonferrua In your example, the order doesn't matter, correct? [Codepen](https://codepen.io/nolanlawson-the-selector/pen/oNdEpKR)

Screen Shot 2022-09-28 at 10 47 42 AM

No matter what order the `<style>`s are in, the result should be the same.

brandonferrua: It would if I have CSS property declarations, for example https://codepen.io/brandonferrua/pen/abGqqNM/512551f3612bdde4e0c228d1b5bbaf6d

nolanlawson: @brandonferrua Yes, that makes sense. I just wanted to make sure I understand what your use case is.

In order to get the order you want (`A B X`), we would need to insert stylesheets at a particular point in the DOM/`adoptedStyleSheets` rather than just blindly appending. (I.e. insert `B` between `A` and `X`.) Also as I mentioned, it doesn't solve the problem of "Rendering A, then B, then A again." (In that case, the order would still be `A B X`, with `B` after `A`.)

I am starting to wonder if the only way to truly solve this problem is to [remove unused stylesheets](https://github.com/salesforce/lwc/pull/2797). In that system, rendering B would remove A's stylesheets, and vice versa. (It was a small perf hit, but the PR proved it was doable.) Either that, or when running `render()`, we could change the order of stylesheets.

brandonferrua: Ah yes, I see the concern about "A, then B, then A again".

Removing the unused stylesheets seems like an elegant idea and a cleaner experience than dealing with `render()`. 

Could both implementations exist? Where the static property always is injected last and you remove unused stylesheets or someone chooses to be explicit, constructing their templates and stylesheets using the `render()` method for full control.

nolanlawson: > Could both implementations exist? Where the static property always is injected last and you remove unused stylesheets or someone chooses to be explicit, constructing their templates and stylesheets using the `render()` method for full control.

Sure, but in that case, would you still need the static property? You could always append whatever stylesheet you want to the end of the `stylesheets` you return from `render()`:

```js
render() {
  return {
    template,
    stylesheets: [otherStylesheet, stylesheetIWantLast]
  }
}
```

In my mind, the ordering for the above would be:

- `template`'s inherent stylesheets
- `otherStylesheet`
- `stylesheetIWantLast`

brandonferrua: If I'm looking at how much code the author would need to write to accomplish the same thing, I'd prefer the static property.

If I Subclass an ancestor Class:
```
import barStylesheets from 'bar.css';

class Bar extends Foo {
  static stylesheets = [...stylesheets, barStylesheets];
}
```

I don't need to import the HTML template plus the inherited stylesheet, I get that implicitly so that's nice.

If I could only use the render method, it gets a bit clunky:
```
import barTemplate from 'bar.html'
import fooStylesheets from '../foo/foo.css';
import barStylesheets from 'bar.css';

class Bar extends Foo {
  render() {
    return {
      template: barTemplate,
      stylesheets = [fooStylesheets, barStylesheets]
    }
  }
}
```
It's fine and it works but I feel like I'm having to "reset" things I otherwise should have gotten through inheritance.

nolanlawson: @brandonferrua Interesting. In my `render` proposal, though, I believe this should work?

```js
class Bar extends Foo {
  render() {
    const fooStylesheets = super.render().stylesheets
    return {
      template,
      stylesheets: [...fooStylesheets, ...barStylesheets]
  }
}
```

I agree though that it's less elegant than `static stylesheets = [...super.stylesheets, ...barStylesheets]`.

nolanlawson: Another wrinkle to add to this problem:

Even if we removed unused stylesheets, there is still a problem if we have multiple components of the same class on the same page.

For instance, imagine two multi-template components. The first may be in state A and the second in state B. If they are using a global styling system (synthetic shadow or light-DOM-scoped), then we can't have an ordering between the A and B stylesheets that satisfies the cascade needs for both components.

In terms of programmatic stylesheets, we would have the same issue if two components of the same class rendered different non-template stylesheets (i.e. ` render() { return { template, stylesheets } }`) in my proposal.

The only way to fully solve this may be to 1) have unique scope tokens per-instance of a component, and 2) remove unused stylesheets.

nolanlawson: Another thought: this problem already exists today with multi-template components. Perhaps we need a separate RFC to address it.

And even without addressing that use case: leveraging the `render()` function to return the `stylesheets` should (at least) replace the [current hacks](https://github.com/salesforce/lwc-test/pull/147), and it would still work for multi-template components with no conflicts between the different stylesheets.

text/0000-programmatic-style-injection.md
pmdartus: I like the approach. The only downside with this is that it requires importing the template manually.

text/0000-programmatic-style-injection.md
```diff
@@ -0,0 +1,172 @@
...
+
+Component libraries are a prime example of this use case. Component authors want to expose high-quality base components with minimal styles and let application developers add their branding.
+
+LWC implicit template and stylesheet resolution system makes it impossible today to extend from a base LWC component and override its styles at the same time. This proposal addresses this issue by exposing a mechanism to associate stylesheets directly with the LWC component class. 
+
+## Detailed design
+
+A new `stylesheets` static property is added on the `LightningElement` constructor. This property allows developers to define which stylesheets the component should use. It accepts an array of [LWC stylesheets](#lwc-stylesheets-modules). The base `LightningElement.stylesheets` static property returns an empty array.
```
nolanlawson, line 38: > The base `LightningElement.stylesheets` static property returns an empty array.

This does not seem possible to implement. The problem is that the `LightningElement` object is [frozen and sealed](https://github.com/salesforce/lwc/blob/06355248d748d9b14a4bfc837c08a2ec3236c542/packages/%40lwc/engine-dom/src/apis/lightning-element.ts#L49-L50).

So we cannot _both_ set it to an empty array (allow people to do `...super.stylesheets`), and also allow `static stylesheets` in the element itself. It throws an error: `Cannot assign to read only property 'stylesheets'`.

I would recommend making `LightningElement.stylesheets` undefined. And we can tell developers that using `...super.stylesheets` is fine, unless you inherit directly from `LightningElement`. This has some precedent in the fact that `super.connectedCallback()` will throw an error unless you inherit from some non-`LightningElement` class.

pmdartus: > This does not seem possible to implement. The problem is that the `LightningElement` object is [frozen and sealed](https://github.com/salesforce/lwc/blob/06355248d748d9b14a4bfc837c08a2ec3236c542/packages/%40lwc/engine-dom/src/apis/lightning-element.ts#L49-L50).

I am not certain of this. The following snippet works as expected ([fiddle](https://jsfiddle.net/g6j8kqhr/)):

```js
'use strict';

class Base {
  static stylesheets = [];
}

Object.freeze(Base);
Object.seal(Base.prototype);

class Foo extends Base {
  static stylesheets = ['foo'];
}

console.log(Base.stylesheets);
console.log(Foo.stylesheets);
```

I think the issue comes from the fact that LWC component class fields are currently transformed with [loose](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties#loose) set to `true` ([compiler](https://github.com/salesforce/lwc/blob/06355248d748d9b14a4bfc837c08a2ec3236c542/packages/%40lwc/compiler/src/transformers/javascript.ts#L45)). With this flag on, the class fields are transformed into assignments instead of property definitions ([playground](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=true&code_lz=OQVwzgpgBGAuBOBLAxrYBuAUJ5AbAhmGFAEKHQDemAkHPrCjLAJ64RgAWEEsxAvFADaAXSwBfbAHkARgCsIqAHQAzeNwBeEABRlIASiwz5SyPlw7yigA7wA9rHvMrEA9jyFiAMVu2oEAB6wEAB2ACbEupQ0dAzITKzsXDz8QsDKPsCimBI4tsFgtmyKuLYA5haQinAJnNy8rsh5BUUl5d62VSxstckGQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env&prettier=false&targets=&version=7.19.6&externalPlugins=&assumptions=%7B%7D)).

nolanlawson: @pmdartus You must be correct. I was seeing the loose transformation when I was debugging this.

Do you think it's a big problem if developers need to do:

```js
static stylesheets = [...(super.stylesheets || []), stylesheets]
```

It's ugly, but I'm not terribly bothered by it. 🤷

pmdartus: It's far from being ideal. But it's OK.

text/0000-programmatic-style-injection.md
```diff
@@ -0,0 +1,172 @@
...
+}
+```
+
+### LWC stylesheets modules
+
+Programmatically importing a stylesheet module from JavaScript using an import declaration is not a new concept introduced by this RFC. Such import is already supported by the LWC compiler. In fact, the template compiler is already using this mechanism to import the stylesheet implicitly associated with the template.
+
+However, until now, the compile-time behavior and runtime behavior for stylesheet modules have never been fully specified. This section fills this specification gap.
+
+Stylesheets modules are identifiable by their `.css` extension. This kind of module only exposes a single default export. The exported value is an opaque object, in other words, the internal implementation of the stylesheet is not exposed to developers. Developers shouldn't rely on the shape of those objects as it may change.
```
nolanlawson, line 101: @pmdartus After discussion with @rui-rayqiu, I'm not sure we need to make the stylesheet objects opaque. Instead, I would propose we use a similar system to the one we currently use for templates (`isTemplateRegistered`).

In short: there is not much benefit to making the object truly opaque. Sure, developers will see something like `LWCCSSStyleSheet` in the console rather than a function. But we don't do this for templates, so why bother for stylesheets? At the end of the day, we just want to disallow people from using arbitrary string-returning-functions, and we can do this with a `registerStylesheet()` utility.

We already [experimented with this](https://github.com/salesforce/lwc/pull/2708), but we didn't do it out of fear of breakages. If we could do it in a non-observable way (e.g. warn rather than throw an error), then I think this is feasible.

pmdartus: > At the end of the day, we just want to disallow people from using arbitrary string-returning-functions, and we can do this with a `registerStylesheet()` utility.

My idea was to future-proof the stylesheet function by converting it into an object and at the same time making it opaque. By turning it into an object we could do a brand-check via the `instanceof` operator instead of storing this into a `WeakMap` and expose new methods on this object in the future if necessary. 

At the end of the day, in JavaScript, a function is just a special kind of object. Is it a deal breaker? I don't think so. We can certainly keep what we already have in place today. 

I don't have a very strong opinion on this specific aspect of the proposal. If you feel migration overhead and the potential cost of dealing with breakage outweigh the benefits, I am fine with stripping this from the proposal.

pmdartus: @rui-rayqiu @nolanlawson What approach would you like to pursue?

text/0000-programmatic-style-injection.md
```diff
@@ -127,11 +127,15 @@ import styles from "./styles.css" assert { type: "css" };
...
 
-This behavior is problematic for explicit stylesheet imports as the compilation doesn't fail if the imported stylesheet can't be resolved. To solve this issue a new `optional` query string can be added to the import identifier to indicate whether or not the LWC compiler should report an error when the file is missing.
+This behavior is problematic for explicit stylesheet imports as the compilation doesn't fail if the imported stylesheet can't be resolved. To solve this issue, the LWC compiler should treat component's implict stylesheet import as optional and other explicit stylesheet imports as required. For DX improvement, LWC compiler should fail compilation or at least warn developers if an explicit stylesheet import cannot be resolved before treating non-existent stylesheet imports as empty imports. 
+
+An alternative solution is that a new `optional` query string can be added to the import identifier to indicate whether or not the LWC compiler should report an error when the file is missing.
```
nolanlawson, line 138: I don't think `?optional=true` is an alternative solution. Instead, I see it as an under-the-hood implementation detail of `To solve this issue, the LWC compiler should treat component's implict stylesheet import as optional and other explicit stylesheet imports as required.`. @rui-rayqiu

text/0000-programmatic-style-injection.md
nolanlawson: 1. Add reporting to catch when someone is mutating a stylesheet (either setting `stylesheet.$scoped$` or using their own function-that-returns-a-string)
2. Mark this behavior as deprecated (e.g. in console warnings)
3. Roll out a "registered" and `Object.freeze`'d stylesheet function which disallows mutations

We already have a similar plan for the "frozen template": https://github.com/salesforce/lwc/issues/2826

nolanlawson:
@pmdartus @rui-rayqiu It just occurred to me (after reading some typos in the comments for #76) that we probably should have called this `static styleSheets` rather than `static stylesheets`. That's the way the browser APIs write it:

- [`document.styleSheets`](https://developer.mozilla.org/en-US/docs/Web/API/Document/styleSheets)
- [`document.adoptedStyleSheets`](https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets)
- [`CSSStyleSheet`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet)

Otherwise I worry people will type `styleSheets` and be confused that it doesn't work.

This feature has been shipped for a couple weeks, but we can change it in a backwards-compatible way. (Warn if using `stylesheets`, remove it later.) What do you think?

rui-rayqiu:
> Alternatively, we can call it `styles` and avoid the case confusion altogether.

@nolanlawson I like `styles` for avoiding confusion. But would calling it `styles` causing issues if some existing components are already using a property `styles` on the component?

nolanlawson:
> But would calling it `styles` causing issues if some existing components are already using a property `styles` on the component?

Yes, but only if it's a `static styles`. And we already have this problem with `static stylesheets` – we just ignore it if it's not an array of functions.

nolanlawson:
After internal discussion, we've decided to stick with `stylesheets`. `styles` is confusing because it looks similar to `Element.prototype.style`, and we can fix confusion about the casing by using linters.

PR #58, opened 2021-11-01
Native Web Component Support

ekashida:


text/0000-native-web-component-support.md
```diff
@@ -0,0 +1,143 @@
...
+possibility of rendering custom elements that are `HTMLUnknownElement`, which is basically an
+extension of `HTMLElement` without any additional fields or methods.
+
+To ensure that only off-platform applications can use this feature, it should be implemented behind
+a feature flag (e.g., `ENABLE_CUSTOM_ELEMENT_REGISTRY_RESOLUTION`).
+
+### Platform restriction
+
+In addition to the OSS restrictions outlined above, in order for LWC to support native web
+comopnents for platform applications, the following needs to happen.
```
kevinv11n, line 93:
```suggestion
components for platform applications, the following needs to happen.

```

text/0000-native-web-component-support.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Detailed design
+
+### OSS restriction
+
+In order for LWC to support native web components for OSS applications, the following needs to
+happen.
+
+1. Prevent LWC components from using the same custom element as a native web component.
+2. Prevent native web components from using the same custom element as an LWC component.
```
kevinv11n, line 73: This wording confuses me. Do you mean that LWC and native web components must not collide on names?

text/0000-native-web-component-support.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Native Web Component Support
```
pmdartus, line 2: To be more precise can we rename this proposal to Native Custom Element support.

```suggestion
title: Native Custom Element Support
```

text/0000-native-web-component-support.md
```diff
@@ -0,0 +1,143 @@
...
+
+```html
+<html>
+  <body>
+    <hello-world></hello-world>
+  </body>
+</html>
+```
+
+The usage of native web components is generally not allowed on the platform so the main use case is
```
pmdartus, line 30:
```suggestion
The usage of native custom element is generally not allowed on the platform so the main use case is

```

text/0000-native-web-component-support.md
```diff
@@ -0,0 +1,143 @@
...
+
+### OSS restriction
+
+In order for LWC to support native web components for OSS applications, the following needs to
+happen.
+
+1. Prevent LWC components from using the same custom element as a native web component.
+2. Prevent native web components from using the same custom element as an LWC component.
+3. Remove the restriction that LWC must know about all available components before rendering.
+
```
kevinv11n, line 75: Also to solve:
4. Passing attributes vs properties
5. Passing non-primitive HTML types

text/0000-native-web-component-support.md
```diff
@@ -0,0 +1,143 @@
...
+To ensure that only off-platform applications can use this feature, it should be implemented behind
+a feature flag (e.g., `ENABLE_CUSTOM_ELEMENT_REGISTRY_RESOLUTION`).
+
+### Platform restriction
+
+In addition to the OSS restrictions outlined above, in order for LWC to support native web
+comopnents for platform applications, the following needs to happen.
+
+1. Prevent registration of custom elements in a namespace different from the current one.
```
pmdartus, line 95: This restriction is too aggressive IMO. This will prevent platform developers to reuse existing web component libraries and greatly reduce the usefulness of this proposal.

I also don't think LWC should be in the business of restricting custom elements from being registered based on the current namespace. Locker should be in charge of this.

text/0000-native-web-component-support.md
nolanlawson: Looking at the test suite for [Custom Elements Everywhere](https://custom-elements-everywhere.com/), it looks like they recommend:

- Arrays/objects should be passed as properties
- Booleans/numbers/strings can be passed as either properties or attributes

In order to pass this test suite, we will also need to handle events with `names-like-this` or `NamesLikeThis`, which LWC currently doesn't support. (At least, in the declarative template syntax, which is what the test suite explicitly tests for.) I.e.:

```html
<!-- currently does not work -->
<some-element onkebab-case={foo}></some-element>
<another-element onPascalCase={foo}></another-element>
```

text/0000-native-web-component-support.md
pmdartus: > I take it you think Locker should hijack the global `customElements.define()` and handle the namespacing there?

Correct

text/0000-native-web-component-support.md
pmdartus: > In order to pass this test suite, we will also need to handle events with `names-like-this` or `NamesLikeThis`, which LWC currently doesn't support.

For this, we will certainly need to drop `parse5` and use a custom HTML parser. This will be quite a large undertaking.

text/0000-native-web-component-support.md
nolanlawson: Assuming that other frameworks are able to handle custom elements at runtime (props/attrs/events/etc.) without intercepting the `define()` process, my assumption is that LWC should be able to do the same.

nolanlawson:
Is this RFC superseded by https://github.com/salesforce/lwc-rfcs/pull/66 ?

PR #59, opened 2021-12-15
RFC: template refs

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/refs/text/0000-refs.md)

### DIFF
--- a/text/0000-refs.md
+++ b/text/0000-refs.md
@@ -200,7 +200,23 @@ The impact of _not_ doing this is that it's much trickier to find the right elem
 
 ## Adoption strategy
 
-This is not a breaking change, and there is no need to replace existing uses of `querySelector` / `querySelectorAll` / etc. Developers are free to adopt refs incrementally as they write new components, or to modify existing components if they wish.
+This is potentially a breaking change in a very edge-casey scenario where the component relies on `this.refs` _not_ being defined in advance. For instance:
+
+```js
+export class Example extends LightningElement {
+  renderedCallback() {
+    if (this.refs) {  // <-- Will always return true
+      this.refs.div.textContent += '[updated]!';
+    } else {
+      this.refs = { div: this.template.querySelector('div') };
+    }
+  }
+}
+```
+
+This use case is probably not very common, though.
+
+Also, there is no need to replace existing uses of `querySelector` / `querySelectorAll` / etc. Developers are free to adopt refs incrementally as they write new components, or to modify existing components if they wish.
 
 A codemod may be tricky, but not impossible, to implement. (Search for existing usages of `querySelector`, search the template for a matching element, replace with refs.)
 
--- a/text/0000-refs.md
+++ b/text/0000-refs.md
@@ -196,7 +196,7 @@ The impact of _not_ doing this is that it's much trickier to find the right elem
 
 ## Adoption strategy
 
-Technically this is a breaking change, because developers may have already defined some object named `refs` in their component. We will need to do some analysis of potential breakage and find workarounds (e.g. ESLint rules).
+Technically this is a breaking change, because developers may have already defined a property named `refs` in their component. We will need to do some analysis of potential breakage and find workarounds (e.g. ESLint rules).
 
 # How we teach this
 
--- a/text/0000-refs.md
+++ b/text/0000-refs.md
@@ -1,5 +1,5 @@
 ---
-title: Refs
+title: Template refs
 status: DRAFTED
 created_at: 2021-12-15
 updated_at: 2022-06-16
--- a/text/0000-refs.md
+++ b/text/0000-refs.md
@@ -1,0 +1,209 @@
+---
+title: Refs
+status: DRAFTED
+created_at: 2021
+---
+
+# Refs
+
+## Summary
+
+This proposal offers a way to explicitly reference DOM nodes defined in a template from within a component's JavaScript.
+
+## Basic example
+
+```html
+<!-- component.html -->
+<template>
+    <div lwc:ref="foo">Hello world!</div>
+</template>
+```
+
+```js
+// component.js
+import { LightningElement } from 'lwc';
+export default class extends LightningElement {
+  renderedCallback() {
+    console.log(this.refs.foo.textContent); // 'Hello world!'
+  }
+}
+```
+
+## Motivation
+
+As of today, the only way to access DOM nodes defined in the template via JavaScript is using standard DOM APIs on `this.template` (shadow DOM) or `this` (light DOM):
+
+```js
+// shadow DOM
+this.template.querySelector('div');
+
+// light DOM
+this.querySelector('div');
+```
+
+This works, and it follows vanilla web component conventions, but it's cumbersome. It forces the developer to think of a query selector (or another technique, like `.firstChild`) to locate an element in the DOM.
+
+In the context of light DOM components, there is an additional difficulty: `this.querySelector()` may return elements from outside of the immediate template (e.g. light DOM children). So developers have to choose their selectors carefully.
+
+### Prior art
+
+The concept of "refs" exist in many other frameworks:
+
+- [React: `createRef()`](https://reactjs.org/docs/refs-and-the-dom.html)
+- [Vue: template refs](https://v3.vuejs.org/guide/composition-api-template-refs.html)
+- [Svelte: `bind:this`](https://svelte.dev/docs#template-syntax-element-directives-bind-this)
+- [Lit: `createRef()`](https://lit.dev/docs/api/directives/#ref)
+- [Angular: `elementRef`](https://angular.io/api/core/ElementRef)
+- [Stencil: `ref`](https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element)
+
+By building on these conventions, we offer a simple and ergonomic way to solve the problem of locating elements at runtime.
+
+## Detailed design
+
+The basic design of refs is the use of the `lwc:ref` directive in templates:
+
+```html
+<template>
+    <div lwc:ref="foo"></div>
+</template>
+```
+
+...which can be accessed at runtime in a component using `this.refs`:
+
+```js
+export default class extends LightningElement {
+  renderedCallback() {
+    console.log(this.refs.foo); // `<div>` DOM node
+  }
+}
+```
+
+### `lwc:ref`
+
+`lwc:ref` can only refer to a static string. For instance, this will throw a template compiler error:
+
+```html
+<template>
+    <div lwc:ref={dynamic}></div> <!-- Not allowed! -->
+</template>
+```
+
+`lwc:ref` cannot be applied to `<template>`s, or to `<slot>`s in light DOM:
+
+```html
+<template>
+    <template if:true={foo} lwc:ref="foo"></template> <!-- Not allowed! -->
+</template>
+```
+
+```html
+<template lwc:render-mode="light">
+    <slot lwc:ref="foo"></slot> <!-- Not allowed! -->
+</template>
+```
+
+### `this.refs`
+
+`this.refs` is a plain object, with string keys and DOM node values. If a ref does not exist, then `this.refs["unknown"]` returns `undefined`.
+
+The object is read-only, and attempting to add, modify, or delete its properties from within the component will result in a runtime error.
+
+In the case of shadow DOM components, `this.refs` refers to elements inside of the shadow DOM. Whereas with light DOM components, it refers to elements inside of the light DOM. This makes sense, since in either case, the element must be defined in the template owned by that component in order to be accessible with `this.refs`.
+
+### Multiple templates
+
+In the case of multiple templates, the same rules as for `this.template.querySelector()` or `this.querySelector()` apply. Once the new template is rendered, `this.refs` refers to elements defined in the new template:
+
+```js
+import a from './a.html';
+import b from './b.html';
+
+export default class extends LightningElement {
+  count = 0;
+  
+  render() {
+    return this.count % 2 === 0 ? a : b;
+  }
+  
+  renderedCallback() {
+    console.log(this.refs);
+  }
+  
+  increment () {
+    this.count++;
+  }
+}
+```
+
+```js
+// Logs `this.refs` for a.html
+const cmp = createElement('x-component', { is: Component });
+
+// Logs `this.refs` for b.html
+cmp.increment();
+
+// Logs `this.refs` for a.html
+cmp.increment();
+```
+
+### Duplicate `lwc:ref`s
+
+If duplicate `lwc:ref`s are defined in the same template:
+
+```html
+<template>
+    <div lwc:ref="foo"></div>
+    <div lwc:ref="foo"></div>
+</template>
+```
+
+... then the second one "wins" (in terms of depth-first tree order). The reason for allowing duplicates is explained below.
+
+### Conditional elements
+
+In some cases, it may be useful to have elements that are conditionally defined based on `<template if:true>`. For example:
+
+```html
+<template>
+    <template if:true={darkMode}>
+        <div lwc:ref="foo"></div>      
+    </template>
+    <template if:false={darkMode}>
+        <div lwc:ref="foo"></div>
+    </template>
+</template>
+```
+
+In this case, it's convenient to be able to call `this.refs.foo` and to not care whether the element is inside the first `<template>` or the second.
+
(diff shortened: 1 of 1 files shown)


text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+
+Also, refs are non-standard, whereas `querySelector` and friends are standard.
+
+## Alternatives
+
+The impact of _not_ doing this is that it's much trickier to find the right elements, in light DOM in particular.
+
+## Adoption strategy
+
+Technically this is a breaking change, because developers may have already defined a property named `refs` in their component. We will need to do some analysis of potential breakage and find workarounds (e.g. ESLint rules).
```
caridy, line 199: I don't think it is a breaking change if the value is configurable/writable. This will be defined in the `LightningElement.prototype`, meaning if anyone is declaring that, or attempting to redefining it, should be able to do so, overruling the default behavior, meaning it will not interfere with whatever they are doing today.

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+title: Refs
+status: DRAFTED
+created_at: 2021
+---
+
+# Refs
+
+## Summary
+
+This proposal offers a way to explicitly reference DOM nodes defined in a template from within a component's JavaScript.
```
caridy, line 11: In principle, I'm ok with this proposal

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+<template lwc:render-mode="light">
+    <slot lwc:ref="foo"></slot> <!-- Not allowed! -->
+</template>
+```
+
+### `this.refs`
+
+`this.refs` is a plain object, with string keys and DOM node values. If a ref does not exist, then `this.refs["unknown"]` returns `undefined`.
+
+The object is read-only, and attempting to add, modify, or delete its properties from within the component will result in a runtime error.
```
caridy, line 109: So, I have some concerns about these two paragraphs. I don't know yet how do you plan to implement this, but here are my questions:

1. what is the nature of `this.refs` property? is it a configurable accessor?
2. `this.refs === this.refs`?
3. `this.refs === this.oldRefs` assuming `oldRefs` was cached in the previous call to `renderedCallback`? If the `refs` object is reused in between render calls, that will not work, because you you will have to mark it as non-extensible, meaning switching templates will break the object invariants.
4. what will `Reflect.ownKeys(this.refs)` return? is it going to be all `lwc:ref` names of the current selected template? or only those rendered in the DOM during the last rendering cycle?

caridy: Here is my recommendation:

* `LightningElement.prototype.refs` property to be a writable and configurable data property. (this will be easier for backward compatibility.
* the `refs` object value to be defined as non-extensible, with non-configurable, non-writable data entries, and it is produced after each rendering cycle. I'm not 100% sure that this is the right approach though! Maybe a throw away object that is mutable, and its entries are configurable is better, thinking about inheritance is the problem here to preserve some kind of integrity.

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+    </template>
+    <template if:false={darkMode}>
+        <div lwc:ref="foo"></div>
+    </template>
+</template>
+```
+
+In this case, it's convenient to be able to call `this.refs.foo` and to not care whether the element is inside the first `<template>` or the second.
+
+Because the template compiler cannot know which conditions will cause an element to exist or not, we cannot support this use case while also disallowing duplicates. Therefore, the simplest solution for duplicates is to define whichever element is last (in terms of depth-first tree order) to be the one that "wins."
```
caridy, line 179: I'm fine with this, being the last makes sense, but the problem I'm having is to imagine how are you going to implement this in LWC? the diffing algo is tricky in terms of preserving the order of traversal for the vnode tree.

caridy (commented):
Added few comments, looking good, but I will like to see few more clarifications.

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+### Prior art
+
+The concept of "refs" exists in many other frameworks:
+
+- [React: `createRef()`](https://reactjs.org/docs/refs-and-the-dom.html)
+- [Vue: template refs](https://v3.vuejs.org/guide/composition-api-template-refs.html)
+- [Svelte: `bind:this`](https://svelte.dev/docs#template-syntax-element-directives-bind-this)
+- [Lit: `createRef()`](https://lit.dev/docs/api/directives/#ref)
+- [Angular: `elementRef`](https://angular.io/api/core/ElementRef)
+- [Stencil: `ref`](https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element)
```
kevinv11n, line 57: Aura has the concept, too.
```suggestion
- [Stencil: `ref`](https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element)
- [Aura: `aura:id`](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/js_cb_find_by_id.htm)

```

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+
+```html
+<template lwc:render-mode="light">
+    <slot lwc:ref="foo"></slot> <!-- Not allowed! -->
+</template>
+```
+
+### `this.refs`
+
+`this.refs` is a plain object, with string keys and DOM node values. If a ref does not exist, then `this.refs["unknown"]` returns `undefined`.
```
kevinv11n, line 107: > If a ref does not exist, then `this.refs["unknown"]` returns `undefined`.

This reads like if a ref does exist then `"unknown" in this.refs` returns `false`, and if a ref (which one?) then "unknown" in this.refs` returns `true`.  Would it be more clear to write this?
> If a ref does not exist it returns `undefined`. For example, `this.refs.unknown` returns `undefined`.

kevinv11n: > with string keys and DOM node values

Why DOM node values and not DOM element values? How do you attach `lwc:ref` to nodes that are not elements?

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+
+The object is read-only, and attempting to add, modify, or delete its properties from within the component will result in a runtime error.
+
+In the case of shadow DOM components, `this.refs` refers to elements inside of the shadow DOM. Whereas with light DOM components, it refers to elements inside of the light DOM. This makes sense, since in either case, the element must be defined in the template owned by that component in order to be accessible with `this.refs`.
+
+### Multiple templates
+
+In the case of multiple templates, the same rules as for `this.template.querySelector()` or `this.querySelector()` apply. Once the new template is rendered, `this.refs` refers to elements defined in the new template:
```
kevinv11n, line 115: Related to https://github.com/salesforce/lwc-rfcs/pull/59/files#r770203733, if one captures `this.refs` and then changes templates in the next render, what do captured `this.refs` refer to? What if the templates are identical (same content)?

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+  }
+  
+  increment () {
+    this.count++;
+  }
+}
+```
+
+```js
+// Logs `this.refs` for a.html
```
kevinv11n, line 139: Logging happens after the render so it'd be more clear if these comments came after the lines that trigger the render

text/0000-refs.md
pmdartus: The only observable change would be if the component access `this.refs` without defining it first. But I am not much worried about this.

```js
class Example extends LightningElement {
  renderedCallback() {
    if (this.refs) {  // <-- Will always return true
        this.refs.div.textContent += '[updated]!';
    } else {
        this.refs = { 
          div: this.template.querySelector('div') 
        };
    }
}
```

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+### Duplicate `lwc:ref`s
+
+If duplicate `lwc:ref`s are defined in the same template:
+
+```html
+<template>
+    <div lwc:ref="foo"></div>
+    <div lwc:ref="foo"></div>
```
pmdartus, line 156: This example is obviously wrong. The template compiler can warn whenever the same `ref` is used twice is the same scope (same if and for block).

text/0000-refs.md
pmdartus: Regarding @caridy 's 4th point, I am wondering how does the `refs` object evolve throughout the lifecycle of the component:
- Shall all the know refs in the template be set to `undefined` in advance?
- When an element with a ref is removed, is the ref set to `undefined`, or is the property removed from the `refs` object?
- When the template is swapped is the previous `refs` object reused for the new template?

What does the `console.log(this.refs)` print the 3 times it is invoked in the following example?

```js
class Example extends LightningElement {
  visible = false;

  connectedCallback() {
    setTimeout(() => {
      this.visible = true;
      
      setTimeout(() => {
        this.visible = false;
      });
    })
  }

  renderedCallback() {
    console.log(this.refs); // Invoked 3 times
  }
}
```

```html
<template>
  <template if:true={visible}>
    <div lwc:ref="my-div"></div>
  </template>
</template>
```

text/0000-refs.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+In this case, it's convenient to be able to call `this.refs.foo` and to not care whether the element is inside the first `<template>` or the second.
+
+Because the template compiler cannot know which conditions will cause an element to exist or not, we cannot support this use case while also disallowing duplicates. Therefore, the simplest solution for duplicates is to define whichever element is last (in terms of depth-first tree order) to be the one that "wins."
+
+Here is how other frameworks do it:
+
+- React: [duplicates allowed, last one wins](https://codepen.io/nolanlawson-the-selector/pen/dyVvZwZ)
+- Vue: [duplicates allowed, last one wins](https://codepen.io/nolanlawson-the-selector/pen/jOGBadg)
```
pmdartus, line 184: For what it's worth, Vue also accepts a function with `:ref` directive to collect references in an array ([v-for array refs](https://v3.vuejs.org/guide/migration/array-refs.html#frontmatter-title) / [example](https://codepen.io/pmdartus/pen/ZEXeZyB)).

text/0000-refs.md
nolanlawson: Ah okay, this is a great point. I updated the document.

text/0000-refs.md
nolanlawson: These are interesting questions. However we decide to represent the object will have an impact on perf and usability.

For instance: if we recreate the `this.refs` object on every render, then that's potentially bad for perf because we're creating and destroying lots of objects. Plus, the user may cache `this.refs` somewhere, leading to a usability footgun (stale data, memory leaks).

Another option is to implement `this.refs` as a `Proxy` with dynamic getters that lazily look up the given ref on-demand (maybe with some internal caching). This might be the best perf-wise, but it's complex to implement.

Whatever we do, I would prefer to optimize for the 99.9% usecase. E.g. `Object.keys()` or `Reflect.ownKeys()` should probably return `[]` or throw an error, because I can't imagine any use case that wouldn't be better served by `querySelectorAll`.

Because I'm a perf guy, I'm currently leaning towards `this.refs === this.refs`, and `refs` is always the same Object (a Proxy) for the entire lifespan of the component.

> Shall all the know refs in the template be set to `undefined` in advance?

Yes. A common pattern developers will use is something like `if (this.refs.foo !== undefined) { /* ... */ }`, to check if the node exists yet or not. We already see this pattern in the wild with `querySelector` and `firstChild`.

> When an element with a ref is removed, is the ref set to `undefined`, or is the property removed from the refs object?

Depends on whether we go with the Proxy approach or not.  If `Object.keys()` or `Reflect.ownKeys()` returns empty / throws an error, then it won't matter.

text/0000-refs.md
nolanlawson: How about handling it in the template compiler? The template compiler can detect duplicate keys, and add some metadata to say which one "wins" in case of conflicts.

text/0000-refs.md
nolanlawson: Good points, fixed.

text/0000-refs.md
nolanlawson: Discussed this a bit here: https://github.com/salesforce/lwc-rfcs/pull/59#discussion_r770847293

text/0000-refs.md
nolanlawson: Fixed!

text/0000-refs.md
nolanlawson: My point, though, is: why should the template compiler bother? If a template has duplicates, it's not a big problem, as long as we have consistent ordering for which one "wins." Trying to hunt down every "incorrect" duplicate may be more trouble than it's worth.

I also noted that, in this same scenario, none of Svelte/Vue/React threw an error; they just had the last one win.

text/0000-refs.md
nolanlawson: I think we need some answer to this, yes. There is also the question of what happens with a `ref` inside of a loop. I'll add it to the unanswered questions.

text/0000-refs.md
caridy: I'm ok with the Proxy, but then we have to talk about compat mode, bummer!

I don't think returning empty keys is going to work well... what if the user does `Object.freeze(this.ref)`? what kind of problems do you see here?

text/0000-refs.md
caridy: Hmm, I'm curious about this, can you elaborate more?

nolanlawson: I'm suggesting we do something like the following:

1. In the compiler, as we walk the tree, give each `lwc:ref` an incrementing integer
2. Output this integer along with the ref name
3. At runtime, if two refs have the same name, the one with the higher integer wins

Since the tree-walking is done at compile time, it should be consistent and we can make it depth-first or breadth-first or whatever we want. (I don't think it really matters FWIW.)

text/0000-refs.md
nolanlawson: By the time this feature ships, will we still be supporting IE11? Can we just make the feature not work in IE11?

> what if the user does `Object.freeze(this.ref)` ?

If we can't throw an error on `Object.freeze()` (I don't know if this is possible), then I would just call it undefined behavior. Users should not be doing this.

text/0000-refs.md
caridy: Ok, well, good news is that this integer already exists, it is the key. It will require some book keeping but I think it will work fine. Thanks for the clarification.

text/0000-refs.md
caridy: I think we should be careful about this, maybe we should chat offline.

pmdartus: I am also sharing the same concerns regarding introducing a new Proxy object.

An alternative approach would create a ref object with getters. Here is a small example of what I have in mind: [link](https://jsfiddle.net/emx1r9v7/1/).

nolanlawson: Conceptually I'm fine with this approach, I'm just a bit worried about the potential overhead of creating a brand-new `refs` object every time anything changes.

Maybe we can be smart and key off of the list of ref names (array of strings)? This should mitigate the over-creation. Or we can be even smarter and generate the object in the compiler. In either case I would be fine with it.

nolanlawson:
@caridy @pmdartus Just realized this PR was never approved. Can I get a :+1:?

text/0000-refs.md
```diff
@@ -0,0 +1,231 @@
+---
```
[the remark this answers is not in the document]
nolanlawson: Done.

text/0000-refs.md
```diff
@@ -0,0 +1,231 @@
...
+
+# How we teach this
+
+This feature is very similar to the same concept in other frameworks (React, Vue, etc.). So we can leverage that familiarity when teaching this.
+
+One benefit of teaching refs over `querySelector` is that there is no difference between light DOM and shadow DOM. I.e. we don't have to explain "Use `this.template.querySelector` for shadow DOM, but `this.querySelector` for light DOM." Considering that `this.template` is kind of a misnomer (it refers to the shadow root, not the template), it can only improve teachability to avoid it, and to have one system for both shadow and light components.
+
+# Unresolved questions
+
+How should `lwc:ref` work inside of a loop? Should there be an equivalent to Vue's [`v-for` array refs](https://v3.vuejs.org/guide/migration/array-refs.html#frontmatter-title) to collect multiple elements in an array?
```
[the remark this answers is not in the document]
nolanlawson: Fixed.

PR #60, opened 2022-01-13
RFC: Priming hints

cseymourSF:
### DEPRECATED, PLEASE SEE https://github.com/salesforce/lwc-rfcs/pull/61 FOR MORE. 

Proposal for adding comment-based hints for priming mechanisms to evaluate properties with specific values.

### DIFF
--- a/text/0118-priming-hints.md
+++ b/text/0118-priming-hints.md
@@ -1,0 +1,87 @@
+---
+title: Priming Hints
+status: DRAFTED
+created_at: 2022-01-04
+updated_at: 2022-01-04
+pr: https://github.com/salesforce/lwc/pull/????
+---
+
+# Priming Hints
+
+## Summary
+
+Priming is an out-of-band mechanism for ensuring that context-specific LWC modules and dependencies are made available in client caches prior to rendering. Priming is useful for both speeding up rendering of components when online, and for ensuring the ability to render unvisited components when offline.
+
+One priming mechanism available for LWC is [Komaci] (https://rfcs.lwc.dev/rfcs/komaci)
+
+This document suggests an enhancement to LWC that would allow components to provide hints to help drive more complete priming.
+
+## Problem description
+
+### Motivation
+
+For maximum flexibility, some LWC priming mechanisms are made available to run out-of-band in a minimal Javascript environment without a DOM. This would be helpful for use cases such as priming LWCs in a client before a user has viewed them, and evaluating LWC dependencies in a headless server context.
+
+Out-of-band priming mechanisms rely on static analysis to traverse component trees and identify dependencies. However, expanding the full LWC component tree may rely upon conditions that are only properly evaluatable when rendered in a DOM. For example, private members may be used to control the presence of specific children, and these members may be initialized only within the DOM lifecycle. These conditions pose barriers to priming mechanisms that run without rendering LWCs.
+
+In order to increase the scope of priming possible, we suggest allowing component developers to provide priming hints. Priming hints can guide priming mechanisms into artificially substituting in property values that introduce more component sub-trees into the component's markup. This would allow the priming mechanism to override property value calculations and mimick happy path tree expansion, introducing a deeper component sub-tree to prime.
+
+## Example
+
+// TODO: example based on lightning-tab/lightning-tabset
+
+## Design
+
+// TODO: exact format
+
+## Other Possible Strategies
+
+Here is a brief survey of other strategies we considered for priming hints.
+
+### Contextual Import
+
+```js
+import {primingContext} from @salesforce/priming;
+...
+get loadContent {
+    return primingContext?.isPriming || _loadContent;
+}
+```
+
+Here we have the property itself explicitly import an indicator of priming status and use it to drive the value to return for the property. This feels like it would be too much work for developers and they could easily calculate the property incorrectly.
+
+### Annotation
+
+```js
+import {forcePriming} from @salesforce/priming;
+...
+
+@forcePriming
+get loadContent {
+    return _loadContent;
+}
+```
+
+Here the component opts in to the priming-specific override by using a decorator. Using decorators like this seems out-of-step with other parts of the LWC specification.
+
+### Metadata XML
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="xmlns=http://soap.sforce.com/2006/04/metadata">
+    <isExposed>true</isExposed>
+    <primingHints>
+        <primingHint value="true">loadContent</primingHint>
+    </primingHints>
+</LightningComponentBundle>
+```
+
+We could express the priming overrides within metadata XML. This feels a little awkward because it's located separately from the code. A developer may miss making the corresponding change here when refactoring their component logic.
+
+## How we teach this
+
+// TODO
+
+# Survey of prior art
+
+// TODO
--- a/text/0118-priming-hints.md
+++ b/text/0118-priming-hints.md
@@ -28,11 +28,98 @@ In order to increase the scope of priming possible, we suggest allowing componen
 
 ## Example
 
-// TODO: example based on lightning-tab/lightning-tabset
+Consider a "bar" container that acts like a tab set. It holds a collection of items, and allows the user to interactively choose an item to be displayed in its default slot:
+
+```html
+<!-- bar.html -->
+<template>
+    <div class={computedClass} title={title}>
+        <bar-header></bar-header>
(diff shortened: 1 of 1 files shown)


text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+
+These comments would be used on property getters, and their format would be:
+
+```js
+/**
+ * priming-hint: <scope> <value>
+ */
+```
+
+With the following values for "scope":
```
AndrewHuffman, line 85: We should call out that this scope would be required to be provided as part of some option into komaci (e.g., the resolver or prefetch).

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Design
+
+We propose adding new meaningful comments to LWC which allow components to opt-in to marking properties that a priming mechanisms could assume to be true. This will allow priming mechanisms to prime a greater scope of component contents.
+
+These comments would be used on property getters, and their format would be:
+
+```js
+/**
+ * priming-hint: <scope> <value>
```
AndrewHuffman, line 81: You left out some details for `value`; my initial thought was that `value` should just default to a boolean `true`. We could expand upon priming hints to support other values if necessary, but for the initial draft it could just be a boolean `true`. Thoughts?

If you disagree, we should discuss what valid values could be provided; I would suggest only simple literals; e.g., strings, numbers, and booleans to minimize having to parse the comment more than necessary.

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+
+```js
+import {primingContext} from @salesforce/priming;
+...
+get loadContent {
+    return primingContext?.isPriming || _loadContent;
+}
+```
+
+Here we have the property itself explicitly import an indicator of priming status and use it to drive the value to return for the property. This feels like it would be too much work for developers and they could easily calculate the property incorrectly.
```
AndrewHuffman, line 138: +1, this would be too much work and it feels awkward to expose this directly via code. Also; how does this get consumed on the komaci side?

text/0118-priming-hints.md
cseymourSF: I left out mention of Komaci consumption as this RFC is just targeted at open source LWC

text/0118-priming-hints.md
cseymourSF: I agree. this is just leaving the door for non-boolean values in the future.

AndrewHuffman:
@shesmer @kevinv11n @pmdartus @gonzalocordero @caridy — Can we get some feedback on this RFC?

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
+---
+title: Priming Hints
+status: DRAFTED
+created_at: 2022-01-04
+updated_at: 2022-01-04
+pr: https://github.com/salesforce/lwc/pull/????
```
shesmer, line 6: Please add a champion and fix PR number

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+
+For a prefetch mechanism based on static analysis, this structure poses a barrier. The conditional loading of barItem is governed by a member _loadContent that is triggered by user interaction. If the prefetch mechanism extracts composition information from markup, and executes resolution based on that analysis without a DOM or user interaction, the best it could hope for it priming the contents of the barItem that is first displayed by default. For some use cases of priming that is reasonable (such as for speeding up render while online). However, for priming for an offline use case one may desire a greater scope of priming - priming the contents of all the barItems. In this case, it would be desirable to provide a hint to the priming mechanism to say "please assume this if:true condition is true for priming purposes".
+
+## Design
+
+We propose adding new meaningful comments to LWC which allow components to opt-in to marking properties that a priming mechanisms could assume to be true. This will allow priming mechanisms to prime a greater scope of component contents.
```
shesmer, line 75: Are you suggesting those to be publically exposed or is this an internal only feature?

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+ */
+```
+
+With the following values for "scope":
+- always: priming mechanisms should always evaluate the given property with the given value for the purposes of priming through the component tree
+- offline: priming mechanisms should evaluate the given property with the given value for the purposes of priming through the component tree only if the priming is targeted at an offline use case
+
+We differentiate between these 2 scopes because online and offline priming use cases can be qualitatively different. Online priming is aimed at enhancing performance of rendering, with the assumption that further resources and data can be pulled from the server later on as necessary. Offline priming aims to prime a complete useful set of resources and data from the server up-front with no additional requests later. In general, online priming is more sensitive to performance and offline priming is more sensitive to missing scope.
+
+For our bar example, we could modify the barItem component by replacing the use of a @track member with a new @api property that uses could make use of this priming hint to guide a priming mechanism to always prime its contents:
```
shesmer, line 91:
```suggestion
For our bar example, we could modify the barItem component by replacing the use of a `@track` member with a new `@api` property that uses could make use of this priming hint to guide a priming mechanism to always prime its contents:
```

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+## Other Possible Strategies
+
+Here is a brief survey of other strategies we considered for priming hints.
+
+### Contextual Import
+
+```js
+import {primingContext} from @salesforce/priming;
```
shesmer, line 131:
```suggestion
import {primingContext} from `@salesforce/priming`;
```

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+import {forcePriming} from @salesforce/priming;
+...
+
+@forcePriming
+get loadContent {
+    return _loadContent;
+}
+```
+
+Here the component opts in to the priming-specific override by using a decorator. Using decorators like this seems out-of-step with other parts of the LWC specification.
```
shesmer, line 152: Why out of step? it could also be a decorator that allows more information like
`@primingHint('always true')`

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+    return primingContext?.isPriming || _loadContent;
+}
+```
+
+Here we have the property itself explicitly import an indicator of priming status and use it to drive the value to return for the property. This feels like it would be too much work for developers and they could easily calculate the property incorrectly.
+
+### Annotation
+
+```js
+import {forcePriming} from @salesforce/priming;
```
shesmer, line 143:
```suggestion
import {forcePriming} from `@salesforce/priming`;
```

text/0118-priming-hints.md
```diff
@@ -0,0 +1,174 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="xmlns=http://soap.sforce.com/2006/04/metadata">
+    <isExposed>true</isExposed>
+    <primingHints>
+        <primingHint value="true">loadContent</primingHint>
+    </primingHints>
+</LightningComponentBundle>
+```
+
+We could express the priming overrides within metadata XML. This feels a little awkward because it's located separately from the code. A developer may miss making the corresponding change here when refactoring their component logic.
```
shesmer, line 166: The question for me here is whether we ever want to elevate these priming hints in the LWC programming model or whether it says an internal core only technology...
Using this xml would work well for internal core only but not allow us to expose this any other way.

AndrewHuffman:
Recreating this RFC in #61.

gonzalocordero (commented):
Here are some general thoughts around this proposal, the first is a general observation not related specifically to this particular RFC, but I thought it might be a provocative idea to share: 

- We are seeing an increased number of features/proposals that want to "scan" code and perform actions: Komaci, UTAM, The new proposal for Targeting Lower Level Components, etc. I'm not saying all of these are related or can be re-used, but they all desired to introduce a new nomenclature(syntax) to target contents, nodes and do something with it. 

So thinking out loud here, are they apples and oranges? or is there an opportunity here for consolidation? 

Another aspect these proposals seem to fail to mention is: "are there any standard tooling that we could re-use?" Even if that's just recorded on these RFCs, so that we know re-usability and standard alignment has been considered. 


- ok, on this proposal, I don't have a strong opinion on the comment based vs decorator. One thing to keep in mind might be tooling and documentation(do we mark this private vs not, how do we make sure it doesn't show up on a component documentation or even mess with our current automated documentation generation. 

My 2 cents!

gonzalocordero:
This PR is now continued on https://github.com/salesforce/lwc-rfcs/pull/61

PR #61, opened 2022-02-03
RFC: Priming hints

AndrewHuffman:
Proposal for adding comment-based hints for priming mechanisms to evaluate properties with specific values.

### DIFF
--- a/text/0118-priming-hints.md
+++ b/text/0118-priming-hints.md
@@ -4,6 +4,7 @@ status: DRAFTED
 created_at: 2022-01-04
 updated_at: 2022-01-04
 pr: https://github.com/salesforce/lwc/pull/????
+champion: Andrew Huffman (andrew.huffman@salesforce.com)
 ---
 
 # Priming Hints
@@ -12,7 +13,7 @@ pr: https://github.com/salesforce/lwc/pull/????
 
 Priming is an out-of-band mechanism for ensuring that context-specific LWC modules and dependencies are made available in client caches prior to rendering. Priming is useful for both speeding up rendering of components when online, and for ensuring the ability to render unvisited components when offline.
 
-One priming mechanism available for LWC is [Komaci] (https://rfcs.lwc.dev/rfcs/komaci)
+One priming mechanism available for LWC is [Komaci](https://rfcs.lwc.dev/rfcs/komaci)
 
 This document suggests an enhancement to LWC that would allow components to provide hints to help drive more complete priming.
 
@@ -88,7 +89,7 @@ With the following values for "scope":
 
 We differentiate between these 2 scopes because online and offline priming use cases can be qualitatively different. Online priming is aimed at enhancing performance of rendering, with the assumption that further resources and data can be pulled from the server later on as necessary. Offline priming aims to prime a complete useful set of resources and data from the server up-front with no additional requests later. In general, online priming is more sensitive to performance and offline priming is more sensitive to missing scope.
 
-For our bar example, we could modify the barItem component by replacing the use of a @track member with a new @api property that uses could make use of this priming hint to guide a priming mechanism to always prime its contents:
+For our bar example, we could modify the barItem component by replacing the use of a `@track` member with a new `@api` property that uses could make use of this priming hint to guide a priming mechanism to always prime its contents:
 
 ```js
 export default class BarItem extends LightningElement {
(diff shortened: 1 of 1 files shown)


AndrewHuffman:
Continuation of #60

text/0118-priming-hints.md
```diff
@@ -0,0 +1,175 @@
...
+
+For a prefetch mechanism based on static analysis, this structure poses a barrier. The conditional loading of barItem is governed by a member _loadContent that is triggered by user interaction. If the prefetch mechanism extracts composition information from markup, and executes resolution based on that analysis without a DOM or user interaction, the best it could hope for it priming the contents of the barItem that is first displayed by default. For some use cases of priming that is reasonable (such as for speeding up render while online). However, for priming for an offline use case one may desire a greater scope of priming - priming the contents of all the barItems. In this case, it would be desirable to provide a hint to the priming mechanism to say "please assume this if:true condition is true for priming purposes".
+
+## Design
+
+We propose adding new meaningful comments to LWC which allow components to opt-in to marking properties that a priming mechanisms could assume to be true. This will allow priming mechanisms to prime a greater scope of component contents.
```
AndrewHuffman, line 76: > @shesmer Are you suggesting those to be publically exposed or is this an internal only feature?

We're suggesting that this be publicly exposed.

text/0118-priming-hints.md
```diff
@@ -0,0 +1,175 @@
...
+import {forcePriming} from `@salesforce/priming`;
+...
+
+@forcePriming
+get loadContent {
+    return _loadContent;
+}
+```
+
+Here the component opts in to the priming-specific override by using a decorator. Using decorators like this seems out-of-step with other parts of the LWC specification.
```
AndrewHuffman, line 153: > @shesmer Why out of step? it could also be a decorator that allows more information like
@primingHint('always true')

I believe there were some concerns using decorators raised by either @kevinv11n or @pmdartus; unfortunately I don't have context on that conversation.

text/0118-priming-hints.md
```diff
@@ -0,0 +1,175 @@
...
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="xmlns=http://soap.sforce.com/2006/04/metadata">
+    <isExposed>true</isExposed>
+    <primingHints>
+        <primingHint value="true">loadContent</primingHint>
+    </primingHints>
+</LightningComponentBundle>
+```
+
```
AndrewHuffman, line 166: > @shesmer The question for me here is whether we ever want to elevate these priming hints in the LWC programming model or whether it says an internal core only technology...
> Using this xml would work well for internal core only but not allow us to expose this any other way.

Another drawback of this programming model is that it separates out the primingHint metadata away from the code. These makes it more likely the developer would, say, forget to update the XML when they rename its corresponding function.

We're advocating that this be a public feature, but we can discuss if that's really the best course. @pmdartus I know had some reservations on making this public.

text/0118-priming-hints.md
pmdartus: The decorator proposal changed a lot since LWC adopted the syntax. The proposal was at stage 3 back in 2018, and then got retrograded to stage 2 because of implementors' concerns. It has been stuck at stage 2 for the last 3 years. Before adding new decorators, we want to ensure this proposal will actually ship. 

The good news is that the revisited proposal will be presented at the next TC39 meeting (in March) requesting stage 3.

text/0118-priming-hints.md
pmdartus: > @pmdartus I know had some reservations on making this public.

Here are my 2 cents on where we should have this discussion. 

The LWC team uses this repo the `lwc-rfcs` repo track proposals related to the open-source LWC framework. Salesforce-only / internal proposals are usually tracked internally using Quip documents. The repository is public, and I don’t think it is the correct place to discuss internal details. 

Regardless of the approach, we take (a comment, field in meta XML, or a decorator), this information will not be consumed by the LWC framework at runtime. The LWC engine isn't, and shouldn't be, aware of how data is retrieved. It's an implementation detail from its standpoint. 

I would recommend moving this proposal somewhere else.

AndrewHuffman: @pmdartus @shesmer @gonzalocordero I have imported this RFC into an internal quip doc, please look for your mentions.

PR #62, opened 2022-02-15
RFC: Else and Else-If Directives

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/else/text/0000-else.md)

### DIFF
--- a/text/0000-else.md
+++ b/text/0000-else.md
@@ -91,7 +91,7 @@ Both `lwc:elseif` and `lwc:else` must be immediately preceded by a sibling `lwc:
 <template lwc:else></template>
 ```
 
-Whitespace is ignored when considering siblinghood.
+Whitespace is ignored when considering siblinghood. (This is consistent with LWC's current behavior, which is to remove whitespace between elements.)
 
 Furthermore, these directives (along with `if:true` and `if:false`) are mutually exclusive, and multiple cannot be applied to the same element:
 
@@ -150,6 +150,38 @@ Similar to `if:true` and `if:false`, the expression passed in to `lwc:if` and `l
 
 
 More complex expressions (e.g. `!foo`, `foo?.bar?.baz`, or `foo % 2 === 1`) are currently not supported.
+
+#### Slots
+
+A developer may want to use if/else-if/else chains to render a `<slot>`:
+
+```html
+<template>
+    <template lwc:if={foo}>
+        <slot></slot>
+    </template>
+    <template lwc:else={foo}>
+        <slot></slot>
+    </template>
+</template>
+```
+
+In this case, the template compiler should _not_ warn about duplicate slots with the same name. This is a perfectly valid use case, and the template compiler knows for certain that the `<slot>` will not be rendered twice.
+
+You might contrast this with the equivalent code in `if:true` / `if:false`:
+
+```html
+<template>
+    <template if:true={foo}>
+        <slot></slot>
+    </template>
+    <template if:false={foo}>
+        <slot></slot>
+    </template>
+</template>
+```
+
+Today, the template compiler _does_ warn about duplicate slots for this case. But this actually makes sense, because the compiler cannot be certain that the `<slot>` will only render once. (For instance, the `foo` getter may have an inconsistent return value, returning `true` the first time and `false` the second time.)
 
 #### Attribute values
 
@@ -252,8 +284,8 @@ This feature is very similar to the same concept in other frameworks (Vue, Svelt
 As for the existing `if:true` and `if:else` statements, we will need to mark them as "legacy," since they are rendered redundant by the new `lwc:if`. To ease the transition, we can provide:
 
 1. ESLint rules to warn about using `if:true` and `if:else`.
-2. Codemods to transform code that uses `if:true` to the equivalent `lwc:if` usages instead. It should be possible for this codemod to be consistent and non-destructive. (`if:false` is not so easy to replace, but it may be possible with a future, more complex expression syntax.)
+2. Codemods to transform code that uses `if:true` to the equivalent `lwc:if` usages instead. It should be possible for this codemod to be consistent and non-destructive. (`if:false` is not so easy to replace, but it may be possible with a future, more complex expression syntax.) It may also be possible to replace an `if:true={foo}`/`if:false={foo}` chain with the equivalent `lwc:if`/`lwc:else` chain, but this is not 100% semantically equivalent (due to the number of accessor calls), and would be trickier to implement.
 
 # Unresolved questions
 
-None at this time.
+The existing `if:true` and `if:false` directives work on arbitrary elements, not just `<template>`s. Should `lwc:if`/`lwc:elseif`/`lwc:else` work the same way?
--- a/text/0000-else.md
+++ b/text/0000-else.md
@@ -221,9 +221,8 @@ Similarly, `lwc:if` and `lwc:elseif` _must_ have an expression as their value:
 
 ### Semantics
 
-`lwc:if` behaves the same as `if:true`, except that it also supports an initial `!` in the expression.
+`lwc:if` behaves the same as `if:true`. `lwc:elseif` behaves the same as `lwc:if`, except that its behavior depends on any preceding `lwc:if`/`lwc:elseif` statements.
 
-`lwc:elseif` behaves the same as `lwc:if`, except that its behavior depends on any preceding `lwc:if`/`lwc:elseif` statements.
 
 Similarly, the behavior of `lwc:else` depends on its previous siblings.
 
--- a/text/0000-else.md
+++ b/text/0000-else.md
@@ -289,3 +289,5 @@ As for the existing `if:true` and `if:else` statements, we will need to mark the
 # Unresolved questions
 
 The existing `if:true` and `if:false` directives work on arbitrary elements, not just `<template>`s. Should `lwc:if`/`lwc:elseif`/`lwc:else` work the same way?
+
+How does this affect the [template AST](https://github.com/salesforce/lwc/blob/7d22e628375c28f529f09bf1353b0efb672bc47c/packages/%40lwc/template-compiler/src/parser/index.ts#L225-L245)?
--- a/text/0000-else.md
+++ b/text/0000-else.md
@@ -8,7 +8,7 @@ created_at: 2022
 
 ## Summary
 
-This proposal replaces and enhances the existing `if:true` and `if:else` directives to add directives that express "else" and "else if" logic.
+This proposal supersedes the existing `if:true` and `if:else` directives to add directives that express "else" and "else if" logic.
 
 ## Basic example
 
@@ -72,10 +72,7 @@ The basic design is to add three new directives:
 - `lwc:elseif`
 - `lwc:else`
 
-The reason for not reusing the existing `if:true` or `if:else` is:
-
-1. It is more consistent to use `lwc:*` for all three.
-2. There is no need for `*:false`, assuming we allow `!` in template expressions, e.g. `lwc:if={!foo}`.
+The reason for not reusing the existing `if:true` or `if:else` is that it is more consistent to use `lwc:*` for all three. (If there is a pressing need for negation, then we may consider a `{!foo}` or similar expression syntax in the future.)
 
 ### Syntax
 
@@ -151,15 +148,8 @@ Whereas when `lwc:preserve-comments` is enabled, comments become syntactically m
 
 Similar to `if:true` and `if:false`, the expression passed in to `lwc:if` and `lwc:elseif` must use [simple dot notation](https://lwc.dev/guide/reference#if%3Atrue%7Cfalse%3D%7Bexpression%7D).
 
-The exception is that "dot notation" is now expanded to include expressions preceded by a single `!` character, indicating negation:
 
-```html
-<!-- Valid -->
-<template lwc:if={!foo}>not foo</template>
-<template lwc:elseif={!bar}>not bar</template>
-```
-
-More complex expressions (e.g. `foo?.bar?.baz` or `foo % 2 === 1`) are currently not supported.
+More complex expressions (e.g. `!foo`, `foo?.bar?.baz`, or `foo % 2 === 1`) are currently not supported.
 
 #### Attribute values
 
@@ -214,8 +204,8 @@ For example, in a series of sibling `lwc:if` / `lwc:elseif` / `lwc:else` directi
 <template lwc:elseif={kadabra}>
     Kadabra!
 </template>
-<template lwc:elseif={!hocus}>
-    Pocus!
+<template lwc:elseif={hocus}>
+    Hocus pocus!
 </template>
 <template lwc:else>
     Kadabra!
@@ -229,8 +219,8 @@ if (component.abra) {
   // Abra!
 } else if (component.kadabra) {
   // Kadabra!
-} else if (!component.hocus) {
-  // Pocus!
+} else if (component.hocus) {
+  // Hocus pocus!
 } else {
   // Kadabra!
 }
@@ -245,7 +235,7 @@ Note that this means:
 
 This implementation would require the template parser to have knowledge of sibling nodes, rather than treating each sibling node independently.
 
-This implementation would also completely replace the existing `if:true` and `if:false` directives, which may lead to some confusion for developers.
+This implementation would also supersede the existing `if:true` and `if:false` directives, which may lead to some confusion for developers.
 
 ## Alternatives
 
@@ -262,7 +252,7 @@ This feature is very similar to the same concept in other frameworks (Vue, Svelt
 As for the existing `if:true` and `if:else` statements, we will need to mark them as "legacy," since they are rendered redundant by the new `lwc:if`. To ease the transition, we can provide:
 
 1. ESLint rules to warn about using `if:true` and `if:else`.
-2. Codemods to transform code that uses `if:true` and `if:false` to the equivalent `lwc:if` usages instead. It should be possible for this codemod to be consistent and non-destructive.
+2. Codemods to transform code that uses `if:true` to the equivalent `lwc:if` usages instead. It should be possible for this codemod to be consistent and non-destructive. (`if:false` is not so easy to replace, but it may be possible with a future, more complex expression syntax.)
 
 # Unresolved questions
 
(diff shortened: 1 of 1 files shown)


caridy:
It all sounds good, but I have one big concern, the new directives are confusing, or add more confusion than the already confusing directives, e.g.: `lwc:else` clause, which is very different from the rest of the directives.

Have we considered other directives like switch? Is there a prior art on that? e.g.:

```html
<!-- component.html -->
<template>
  <template lwc:switch={abra}>
    <template lwc:case={true}>
        Abra!
    </template>
    <template lwc:case={false}>
        Kadabra!
    </template>
    <template lwc:default>
        Alakazam!
    </template>
</template>
```

Something like this seems a lot more simpler to teach, and might not require the exact order or structure.

Another alternative is to change the directives all together:

```html
<!-- component.html -->
<template>
    <template lwc:if={abra}>
        Abra!
    </template>
    <template lwc:elseif={kadabra}>
        Kadabra!
    </template>
    <template lwc:else>
        Alakazam!
    </template>
</template>
```

The example above reads more clear, but it is a departure from previous syntax, at least cosmetically. The `true` and `false` of the equation is removed, and it is the responsibility of the author to provide truthfully values instead.

nolanlawson:
@caridy

> Have we considered other directives like switch? Is there a prior art on that?

Some frameworks have a switch, yes:

- [Angular: `ngSwitch`](https://angular.io/api/common/NgSwitch)
- [Solid: `<Switch>`](https://www.solidjs.com/docs/latest/api#%3Cswitch%3E%2F%3Cmatch%3E)

What's interesting, though, is that they have _both_ an "if/else" and a "switch." I can see a case for supporting both. The "if/elseif/else" is more flexible, but the "switch" is terser for some cases.

> Another alternative is to change the directives all together

I actually really like `lwc:if`, `lwc:elseif`, and `lwc:else` in terms of the harmoniousness. As you say, the only downside is that there's no equivalent to `if:false` and `elseif:false`, which is awkward given that we require a strict dot notation (no `!` allowed).

However, if we allow `!` (e.g. `lwc:if={!foo}`), then I don't see a problem. In that case, we could deprecate `if:true` and `if:false` entirely. Arguably, this would also be better if we eventually support arbitrary JS syntax in expressions (in which case, `if:false` is unnecessary).

mgoldspink-salesforce:
The `switch` statement version is something that would've helped us a lot at Vlocity in the absence of dynamic loading. I'd say the use cases for `if`, `elseif` and `else` we didn't feel like we missed at all, but `switch` statement we definitely would've liked to simplify a lot of our code.

pmdartus:
Great proposal @nolanlawson!

> Have we considered other directives like switch? Is there a prior art on that? e.g.:

First of all, I really like the switch approach. That said I think it's quite a big leap from the existing template syntax. In your example @caridy, the `lwc:case` expression resolves to the `true` and `false` boolean values. The compiler would throw because those aren't valid template expressions.

> Another alternative is to change the directives all together:

I also really like this proposal as well and would echo @nolanlawson's comment on the missing equivalent of `if:false`. 

----

My main takeaway from this discussion is that the lack of complex expressions in the template is clearly holding us back here. It's something that was raised last year, but we were not able to reach a consensus. Where do you stand on this topic @caridy in 2022?

text/0000-else.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+```html
+<!-- Invalid -->
+<template if:true={foo}></template>
+<div>Yolo!</div>
+<template lwc:else></template>
+```
+
+Whitespace is ignored when considering siblinghood.
```
pmdartus, line 94: I would add a comment here, stating that `if:*`, `esleif:*` and `lwc:else` are mutually exclusive on the same element.

text/0000-else.md
```diff
@@ -0,0 +1,219 @@
...
+</template>
+```
+
+#### Accessors
+
+Similar to `if:true` and `if:false`, the expression passed in to `elseif:true` and `elseif:false` must use [simple dot notation](https://lwc.dev/guide/reference#if%3Atrue%7Cfalse%3D%7Bexpression%7D).
+
+#### Attribute values
+
+To avoid typos (e.g. using `lwc:else` when you mean `lwc:elseif`), `lwc:else` must _not_ have an attribute value. Any attribute value is treated as a compile-time error:
```
pmdartus, line 126:
```suggestion
To avoid typos (e.g. using `lwc:else` when you mean `elseif:true`), `lwc:else` must _not_ have an attribute value. Any attribute value is treated as a compile-time error:

```

text/0000-else.md
```diff
@@ -0,0 +1,219 @@
...
+} else if (!component.hocus) {
+  // Pocus!
+} else {
+  // Kadabra!
+}
+```
+
+Note that this means:
+
+1. The property getters are only accessed once per definition in an `if:` or `ifelse:`.
```
pmdartus, line 197: > per definition

I understand the overall meaning of this sentence, but per definition doesn't look like the correct term here.

text/0000-else.md
nolanlawson: Done.

text/0000-else.md
nolanlawson: Changed to "per instance of."

caridy:
@pmdartus 

> Where do you stand on this topic @caridy in 2022?

I went from "open for the conversation" to "lets have the conversation" with this new year jajajaja!

> I agree, but I worry a bit about "boiling the ocean." I would personally prefer a phased approach, where we add support for ! and maybe other simple stuff (string/number literals?), before adding full support for arbitrary expressions.

I agree with this sentiment... if we introduce `!obj` and `obj.val?.prop`, the existing if:true, etc... can use those new value syntax since it is just that, operators for value syntax.

They are going to help a lot, but can we draw the line on what is going to be supported?

caridy:
Btw, I'm more in favor of the `switch` because that's new syntax vs reforming what we have since we haven't get a lot of chores about what we have right now.

nolanlawson:
I think `switch` is great, but I see it as an addition, not a replacement, for if/elseif/else. There are still cases where a component has multiple booleanish properties, not one property with multiple values.

pmdartus:
> I agree with this sentiment... if we introduce `!obj` and `obj.val?.prop`, the existing if:true, etc... can use those new value syntax since it is just that, operators for value syntax.

Agreed, allowing any kind of expressions in the template might be too much of a leap here. I don't have any objection to adding support for the `!` unary operator (other unary operators like `delete` or `++` are more controversial) and optional chaining. 

Regarding literals, I don't have concerns with null, numeric, and boolean values. String literals are a little more complex. Attribute expressions values aren't quoted today. We would have to change this if we want to support strings containing whitespace characters.

```html
<template>
  <div title={'Some title'}></div> <!-- Produces 2 attributes: `title="{'Some"` and `title'}=""` -->
</template>
```

caridy:
Let's not hijack this pull request with that, we can have a separate one to discuss the grammar.

kevinv11n:
> if we introduce !obj and obj.val?.prop, the existing if:true, etc... can use those new value syntax since it is just that, operators for value syntax.
> They are going to help a lot, but can we draw the line on what is going to be supported?

Drawing the line is key. We initially kept operators out of the template for **testability** reasons: force nearly all [1] logic into the backing JS class so it can be tested without rendering. 

[1] "nearly all" because the template supports iterations and if. This is where we historically drew the line.

nolanlawson:
It occurred to me that, if we allow arbitrary JS expressions in templates, then that would reduce the value-add of `switch`. You can just use chained `if`/`if-else` to achieve the same effect. (Not that we still shouldn't do `switch`; it's a nice shorthand.)

@kevinv11n My impression is that most component test code isn't really taking advantage of the fact that the logic is isolated to the component rather than the template. I see a lot of Jest test cases that render to JSDOM and check the rendered DOM output (e.g. snapshots). Not that component API testing doesn't have its place, but it just seems to me that it hasn't proved to be a super-compelling part of the testability story. (I could be wrong though.)

nolanlawson:
Since @pmdartus and I both like @caridy's proposal, I've updated the RFC to use `lwc:if`/`lwc:elseif`/`lwc:else` rather than `if:true`/`elseif:true`/`lwc:else`.

The new proposal is more internally consistent, although it does require us to deprecate `if:true` and `if:false`, as well as add support for `!` in expressions. I've updated the RFC to cover these issues.

text/0000-else.md
```diff
@@ -0,0 +1,269 @@
...
+    <!-- Comment! -->
+    <template lwc:else></template>
+</template>
+```
+
+#### Accessors
+
+Similar to `if:true` and `if:false`, the expression passed in to `lwc:if` and `lwc:elseif` must use [simple dot notation](https://lwc.dev/guide/reference#if%3Atrue%7Cfalse%3D%7Bexpression%7D).
+
+The exception is that "dot notation" is now expanded to include expressions preceded by a single `!` character, indicating negation:
```
kevinv11n, line 154: `{!` is the delimeter for expressions in Salesforce: [Aura](https://developer.salesforce.com/docs/atlas.en-us.234.0.lightning.meta/lightning/expr_overview.htm), [Visualforce](https://trailhead.salesforce.com/content/learn/modules/visualforce_fundamentals/visualforce_variables_expressions#:~:text=The%20expression%20syntax%20in%20Visualforce%20is:). 

This may cause confusion. I think some user research is needed.

text/0000-else.md
```diff
@@ -0,0 +1,269 @@
...
+title: Else and Else-If Directives
+status: DRAFTED
+created_at: 2022
+---
+
+# Else and Else-If Directives
+
+## Summary
+
+This proposal replaces and enhances the existing `if:true` and `if:else` directives to add directives that express "else" and "else if" logic.
```
kevinv11n, line 11: > This proposal replaces ...

I'm surprised this is a replacement rather than deprecate and offer a better alternative. Offline I'd like to understand the rollout plan for Lightning Platform. cc @pmdartus

kevinv11n (commented):
Two big concerns per comments.

> My impression is that most component test code isn't really taking advantage of the fact that the logic is isolated to the component rather than the template.

To me this means we need improved enablement for testing, not that we should embrace more logic in the template and DOM-centric testing.

text/0000-else.md
pmdartus: I don't think this proposal replaces the existing `if:true` and `if:false` directives. As [stated](https://github.com/salesforce/lwc-rfcs/pull/62/files#diff-2556c5c6cd88d674348a2e918cb09e69e6a8f133b8930926d945660d81f31140R262) in the _How we teach this_ section, we would mark the existing directives are legacy. I don't see us dropping support for `if:*` directives in the foreseeable future.

This sentence can certainly be rephrased to alleviate this confusion.

```suggestion
This proposal superseed the existing `if:true` and `if:else` directives to add directives that express "else" and "else if" logic.

```

----

Also on the topic of Salesforce platform enablement, I am wondering if this feature should or not be enabled by default. Should this be gated behind an API version to incentivize developers to bump their component API version's to beneficiate from this feature.

text/0000-else.md
pmdartus: This is a good callout. On this topic, I think it is also worth adding that in all the modern JS frameworks, using the `!` token is used as the negation operator in template expressions.

It appears to me that we don't have a consensus on the template expression syntax extension proposed in this RFC. Taking a step back the introduction of the `!` operator is orthogonal to if/else proposal. I would suggest removing it from this proposal and moving the discussion to another RFC. The current lack of capability to express negation in the template makes the use of the if/else new directive less friendly since it would require developers to express the negation in JavaScript, but I think we can work on this in parallel.

Thoughts @nolanlawson @caridy?

pmdartus:
> Drawing the line is key. We initially kept operators out of the template for **testability** reasons: force nearly all [1] logic into the backing JS class so it can be tested without rendering.
> 
> [1] "nearly all" because the template supports iterations and if. This is where we historically drew the line.

IMO it's time to revisit this original motivation.

I think that we were originally afraid to repeat the same mistake we made with Aura components where both the markup and the component had really complex logic. To course-correct this we designed the LWC template syntax to force developers to write slimmer templates and move all the logic to JavaScript.

We can say with confidence today that most of the LWC components have not only slimmer templates but also slimmer JavaScript code compared to Aura components. I would attribute this to fact that developers use to their advantage the ES module system in LWC to extract the complex logic and data massaging outside their component code. Decoupling this logic from the UI enables them to effectively test their logic without having to mount components to the DOM.

Today, component logic can't be tested without rendering the component, since most of the logic is done in accessors and methods that aren't public (not marked with the `@api` decorator). The means by which you can test LWC components are via public props/methods, events, and DOM state. 
And I strongly believe this approach is the correct approach to test UI components (regardless of the framework used). How components arrive at some DOM state is really an implementation detail. If this state computation requires complex logic, the recommendation would be to extract the logic into another module and decouple the logic from the component's internal state.

text/0000-else.md
nolanlawson: If we get rid of `!`, then we should revert https://github.com/salesforce/lwc-rfcs/pull/62/commits/0639cea2cb6f87db8241613e5f215f25ab129c67 and go back to the `elseif:true` / `elseif:false` / `lwc:if` system.

**Pros:** No need for the `!` in expressions. We can reuse the existing `if:true` / `if:false` rather than deprecating them.

**Cons:** The new system is less internally consistent compared to everything being prefixed with `lwc:`.

I am fine either way. Internal consistency is not everything. 🙂

text/0000-else.md
nolanlawson: Correct, it doesn't "replace" it, it supersedes it.

nolanlawson:
I don't think it's an either/or thing. e2e tests are useful in some ways (you render to an actual browser/DOM, you can verify style, layout, and accessibility state) and unit tests are useful in other ways (faster, more consistent, doesn't rely on the browser implementation). I'd be curious to know what % of tests out there in the wild fall into each category, and what value developers perceive that they're getting from each one.

text/0000-else.md
nolanlawson: We could also rename `lwc:else` to `else:render`. Maybe this would look a bit more consistent:

```html
<template if:true={foo}></template>
<template elseif:true={bar}></template>
<template else:render></template>
```

We can also bikeshed on the name. (`else:do`? `else:show`?)

nolanlawson: Just chatted with @pmdartus ; we have a new proposal:

1. Get rid of the `!`
2. Keep `lwc:if` / `lwc:elseif` / `lwc:else`
3. Move the discussion about complex expressions (including `!`) to a future proposal

The benefit is that we can unblock this proposal for the time being, and consider complex expression syntax later.

kevinv11n (approved):

text/0000-else.md
```diff
@@ -0,0 +1,293 @@
...
+<!-- Invalid -->
+<template lwc:if={foo}></template>
+<template lwc:elseif></template>
+```
+
+### Semantics
+
+`lwc:if` behaves the same as `if:true`, except that it also supports an initial `!` in the expression.
+
+`lwc:elseif` behaves the same as `lwc:if`, except that its behavior depends on any preceding `lwc:if`/`lwc:elseif` statements.
```
pmdartus, line 226:
```suggestion
`lwc:if` behaves the same as `if:true`. `lwc:elseif` behaves the same as `lwc:if`, except that its behavior depends on any preceding `lwc:if`/`lwc:elseif` statements.

```

text/0000-else.md
```diff
@@ -0,0 +1,293 @@
...
+
+This is a non-breaking change, so users can opt-in as they see fit.
+
+# How we teach this
+
+This feature is very similar to the same concept in other frameworks (Vue, Svelte, Angular, etc.). It's also similar to `if` and `else` in JavaScript. So we can leverage that familiarity when teaching this.
+
+As for the existing `if:true` and `if:else` statements, we will need to mark them as "legacy," since they are rendered redundant by the new `lwc:if`. To ease the transition, we can provide:
+
+1. ESLint rules to warn about using `if:true` and `if:else`.
```
pmdartus, line 286:
```suggestion
1. Make the template compiler warn about using `if:true` and `if:else`.
```

ESLint doesn't run against template code. I would suggest adding making the template compiler warn. We can create an HTML parser for ESLint but it would require a lot more work.

text/0000-else.md
```diff
@@ -0,0 +1,293 @@
...
+2. Codemods to transform code that uses `if:true` to the equivalent `lwc:if` usages instead. It should be possible for this codemod to be consistent and non-destructive. (`if:false` is not so easy to replace, but it may be possible with a future, more complex expression syntax.) It may also be possible to replace an `if:true={foo}`/`if:false={foo}` chain with the equivalent `lwc:if`/`lwc:else` chain, but this is not 100% semantically equivalent (due to the number of accessor calls), and would be trickier to implement.
+
+# Unresolved questions
+
+The existing `if:true` and `if:false` directives work on arbitrary elements, not just `<template>`s. Should `lwc:if`/`lwc:elseif`/`lwc:else` work the same way?
```
pmdartus, line 291: > The existing `if:true` and `if:false` directives work on arbitrary elements, not just `<template>`s. Should `lwc:if`/`lwc:elseif`/`lwc:else` work the same way?

I would say yes. This avoids adding a `<template>` tag especially if the condition only applies to a single element on each branch.

pmdartus (approved):

text/0000-else.md
```diff
@@ -0,0 +1,292 @@
...
+#### Slots
+
+A developer may want to use if/else-if/else chains to render a `<slot>`:
+
+```html
+<template>
+    <template lwc:if={foo}>
+        <slot></slot>
+    </template>
+    <template lwc:else={foo}>
```
jye-sf, line 163: Should remove the attribute value here

text/0000-else.md
```diff
@@ -0,0 +1,292 @@
...
+In this case, the template compiler should _not_ warn about duplicate slots with the same name. This is a perfectly valid use case, and the template compiler knows for certain that the `<slot>` will not be rendered twice.
+
+You might contrast this with the equivalent code in `if:true` / `if:false`:
+
+```html
+<template>
+    <template if:true={foo}>
+        <slot></slot>
+    </template>
+    <template if:false={foo}>
```
jye-sf, line 178: Should remove the attribute value here

text/0000-else.md
nolanlawson:
```suggestion
    <template lwc:else>
```

gregwhitworth:
> IMO it's time to revisit this original motivation.

I would like to weigh in here as we're actively working on a demo for this. While it's nice to delete 1 additional `if:true block`; it's not quite as compelling to have a declarative solution for more more complex expressions which are currently not supported. We want to be able to do something like `lwc:if={rows > 0 && show}`.

I understand the reasoning above but it was somewhat interesting to be met with an `if` directive but only be able to place a single property within it.

cc: @nolanlawson since I pinged him regarding this & @caridy @kevinv11n @pmdartus since they had relatively strong opinions regarding it

PR #63, opened 2022-06-13
feat: add Scoped Slots for Light DOM

abdulsattar:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/abdulsattar/scoped-slots/text/0000-scoped-slots-light-dom.md)

PoC: https://github.com/salesforce/lwc/compare/pmdartus/scoped-slot (doesn't implement named slots)

### DIFF
--- a/text/0000-scoped-slots-light-dom.md
+++ b/text/0000-scoped-slots-light-dom.md
@@ -1,0 +1,227 @@
+---
+title: Scoped Slots in Light DOM
+status: DRAFTED
+created_at: 2022-06-10
+updated_at: 2022-06-22
+pr: 
+---
+
+# Scoped Slots in Light DOM
+
+## Summary
+
+[Scoped slots](https://v3.vuejs.org/guide/component-slots.html#scoped-slots) is a feature from the Vue
+framework, that allows passing data from a child component to the content slotted from a parent component
+in a declarative fashion.
+
+One of the most common use-cases for this is to use a generic component to render an array of items and
+let the parent component customize how each item is rendered. For example a generic infinite list
+component, a virtualized list component or, a data table.
+
+*This proposal is for adding Scoped Slots only to Light DOM.*
+
+## Basic example
+
+```html
+// x/parent.html
+<template> <!-- parent need not be light DOM -->
+    <x-list items={items} lwc:slot-data="item">
+        <span>{item.name}</span>
+    </x-list>
+</template>
+
+// x/list.html
+<template lwc:render-mode="light">
+    <ul>
+        <template for:each={items} for:item="item">
+            <li key={item.id}>
+                <slot lwc:slot-bind={item}></slot> <!-- delegating rendering of children to the parent -->
+            </li>
+        </template>
+    </ul>
+</template>
+```
+The `<x-parent>` component is slotting `<span>{item.name}</span>` into `<x-list>` but the `item` is supplied
+by `<x-list>`, the child component. It's essentially like `<x-parent>` is slotting a function that actually
+returns the slotted content.
+
+This allows `<x-parent>` to control how each item renders while leaving the looping/repeating logic to `<x-list>`.
+`<x-list>` is now free to create as many slots as it wants while using the same "slotting function" passed from the parent.
+This allows `<x-list>` to be a virtual list or an infinite list.
+
+It's important to note that for this to happen, slots need to be lazily rendered.
+
+* Native shadow DOM slots: the slotted content is eagerly rendered regardless of there being a slot to “accept” the content. The component slotting the content is in charge of appending and removing the content from the DOM.
+* Scoped slots: the slotted content is lazily rendered; the slotted content will only be rendered if there is a slot to “accept” the content. The component with the <slot> elements is the one in charge of rendering the slotted content passed by the parent component.
+
+It is for this reason that Scoped Slots are allowed only in Light DOM components.
+
+## Detailed design
+
+Two new directives, `lwc:slot-bind` and `lwc:slot-data` are introduced that are used to bind data to the slotted content
+and to specify, in the parent, a placeholder to accept the bound data respectively.
+
+* ### `lwc:slot-bind`
+It has to be on a `<slot>` tag and should be an expression that is bound to a value within the child's scope. It can only
+be present on a Light DOM template. The compiler will throw an error if `lwc:slot-bind` is used in a shadow DOM template.
+* ### `lwc:slot-data`
+It has to be present on a custom element and should be a string literal that represents the variable name that will be
+used in the slotted content to reference the passed data. The custom element has to be a Light DOM LWC element. This check,
+though, will be performed at runtime because the LWC compiler lacks the necessary information to throw at compile-time.
+
+### Named Slots
+Accepting scoped slots for named slots presents a few logistical challenges illustrated below.
+```html
+<template>
+    <x-child lwc:slot-data="defaultSlotData">
+        <h1>Content that goes into default Slot {defaultSlotData.title}</h1>
+        <h2 slot="namedSlot">Named slot content {namedSlotContent.title}</h2> // where should namedSlotContent be declared?
+        <h3 slot="otherNamedSlot">{defaultSlotData.title}</h3> // `defaultSlotData` is actually out of scope, but *looks* in scope for a template author
+    </x-child>
+</template>
+```
+
+To address the above concerns, content slotted into named slots needs to be grouped together using a `<template>` tag
+and `lwc:slot-data` should be used on the `<template>`. 
+```html
+// x/parent.html
+<template>
+    <x-child> <!-- no `lwc:slot-data` on the component tag -->
+        <template slot="slotname1" lwc:slot-data="slot1data">
+            <p>{slot1data.name}</p>
+        </template>
+
+        <template slot="slotname2" lwc:slot-data="slot2data">
+            <p>{slot2data.title}</p>
+        </template>
+
+        <template lwc:slot-data="defaultdata"> <!--// for default slot-->
+            <p>{defaultdata.title}</p>
+        </template>
+    </x-child>
+</template>
+
+// x/child.html
+<template lwc:render-mode="light">
+    <slot name="slotname1" lwc:slot-bind={slot1data}></slot>
+    <slot name="slotname2" lwc:slot-bind={slot2data}></slot>
+    <slot lwc:slot-bind={defaultdata}></slot>
+</template>
+```
+
+When using named slots, the child custom element does not accept `lwc:slot-data`. There cannot be any elements other than
+`<template>` as direct children of the custom element. The check for these conditions will happen at compile time.
+The template author can either use the named slot syntax or regular default slot syntax, but cannot combine both.
+
+### Nested Scoped Slots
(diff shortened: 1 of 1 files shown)


text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,227 @@
...
+        </template>
+    </ul>
+</template>
+```
+The `<x-parent>` component is slotting `<span>{item.name}</span>` into `<x-list>` but the `item` is supplied
+by `<x-list>`, the child component. It's essentially like `<x-parent>` is slotting a function that actually
+returns the slotted content.
+
+This allows `<x-parent>` to control how each item renders while leaving the looping/repeating logic to `<x-list>`.
+`<x-list>` is now free to create as many slots as it wants while using the same "slotting function" passed from the parent.
```
pmdartus, line 49:
```suggestion
`<x-list>` is now free to create as many slots as it wants while using the same template fragment passed from the parent.

```

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,227 @@
...
+This allows `<x-list>` to be a virtual list or an infinite list.
+
+It's important to note that for this to happen, slots need to be lazily rendered.
+
+* Native shadow DOM slots: the slotted content is eagerly rendered regardless of there being a slot to “accept” the content. The component slotting the content is in charge of appending and removing the content from the DOM.
+* Scoped slots: the slotted content is lazily rendered; the slotted content will only be rendered if there is a slot to “accept” the content. The component with the <slot> elements is the one in charge of rendering the slotted content passed by the parent component.
```
pmdartus, line 55: Something super important to call out here is that for scoped slot the fragment is still "owned" by the parent component. 

This had direct implications in terms of styling. If the parent component has a scoped stylesheet associated with it, the styles also apply to the scoped slot content (like standard slotted content).

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+To address the above concerns, content slotted into named slots needs to be grouped together using a `<template>` tag
+and `lwc:slot-data` should be used on the `<template>`. 
+```html
+// x/parent.html
+<template>
+    <x-child> <!-- no `lwc:slot-data` on the component tag -->
+        <template slot="slotname1" lwc:slot-data="slot1data">
+            <p>{slot1data.name}</p>
```
pmdartus, line 91: Native shadow DOM slots offer the capability to slot text nodes in named slots. Text and comment nodes are automatically slotted into the default slot.

If we ever want to support a similar syntax for shadow DOM scoped slots, we will need to ensure that the `<template>` only contains elements.

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,227 @@
...
+</template>
+```
+
+To address the above concerns, content slotted into named slots needs to be grouped together using a `<template>` tag
+and `lwc:slot-data` should be used on the `<template>`. 
+```html
+// x/parent.html
+<template>
+    <x-child> <!-- no `lwc:slot-data` on the component tag -->
+        <template slot="slotname1" lwc:slot-data="slot1data">
```
pmdartus, line 90: This syntax makes sense for scoped slots. Would it makes sense to also allow this with standard slots?

pmdartus (commented):
Great work @abdulsattar!

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,227 @@
...
+            <p>{slot1data.name}</p>
+        </template>
+
+        <template slot="slotname2" lwc:slot-data="slot2data">
+            <p>{slot2data.title}</p>
+        </template>
+
+        <template lwc:slot-data="defaultdata"> <!--// for default slot-->
+            <p>{defaultdata.title}</p>
+        </template>
```
nolanlawson, line 100: I like that this has some symmetry with `for:each` and `iterator:*`, since the `<template>` is optional here if you only have a default slot.

```html
<x-child>
  <template lwc:slot-data="foo">
    <div>{foo}</div>
  </template>
</x-child>
```

```html
<!-- equivalent -->
<x-child lwc:slot-data="foo">
  <div>{foo}</div>
</x-child>
```


This format also looks a lot like Vue's [named scoped slots](https://vuejs.org/guide/components/slots.html#scoped-slots). Might be worth mentioning as a "prior art." 🙂

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,227 @@
...
+created_at: 2022-06-10
+updated_at: 2022-06-22
+pr: 
+---
+
+# Scoped Slots in Light DOM
+
+## Summary
+
+[Scoped slots](https://v3.vuejs.org/guide/component-slots.html#scoped-slots) is a feature from the Vue
```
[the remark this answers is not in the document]
nolanlawson: Also worth mentioning: [WCCG proposed protocol for "render props"](https://github.com/webcomponents-cg/community-protocols/issues/8).

nolanlawson (commented):
Overall LGTM, I actually like the idea of restricting this to light DOM slots. Keeps things simple and clear – no need to align with shadow DOM, no need to support `:slotted` or `slotchange`, etc.

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,410 @@
...
+```html
+<x-parent>
+    #shadow-root
+    |    <x-list>
+    |        <ul>
+    |            <li>
+    |                <span>1 - One</span>
+    |            </li>
+    |        </ul>
+    |        <ul>
```
pmdartus, line 192: There should only be a single `<ul>`.

```suggestion
```

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,509 @@
...
+<template lwc:render-mode="light">
+    <slot lwc:slot-bind={slotdata}></slot>
+    <slot name="slotname1" lwc:slot-bind={slotdata}></slot>
+    <slot name="slotname2" lwc:slot-bind={slotdata}></slot>
+</template>
+```
+_The last one is not ambigious, but for simplifying the restriction and conveying that "slot to binding combination should be unique", it will be restricted._
+
+### Mixing Standard Slots and Scoped Slots
+A parent component can set slot content into a child's standard slots and scoped slots. However, it cannot mix the two slot types for a given slot. In other words, the child cannot have two default slots, one for standard slot and another for scoped slot. Similarly, the child annot have two named slots of mixed slot type.
```
pmdartus, line 288:
```suggestion
A parent component can set slot content into a child's standard slots and scoped slots. However, it cannot mix the two slot types for a given slot. In other words, the child cannot have two default slots, one for standard slot and another for scoped slot. Similarly, the child can't have two named slots of mixed slot type.

```

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,509 @@
...
+</template>
+
+// x/child.html
+<template lwc:render-mode="light">
+    <slot lwc:slot-bind={slotdata}>Scoped Slot</slot>
+    <slot>Regular Slot</slot>
+</template>
+```
+
+- Mixed named slots, one for regular slot and another for scoped slot
```
pmdartus, line 329: Is it invalid because they share the same name? Or because the 2 of them are named slots?

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,509 @@
...
+    </x-table> 
+</template>
+```
+
+### Ownership and styling implication
+Standard slots and scoped slots are identical with respect to ownership. The slotted content is owned by the parent component. This has direct implications in terms of styling. If the parent component has a scoped stylesheet associated with it, the styles also apply to the scoped slot content.
+
+### Validations
+Validation of scoped slot usage between parent and component is handled at runtime.
+* **Missing default slot**: If a parent component slots content in to the child component's default slot and the child does not have a default slot, the content will be ignored.(TODO: Why don't we log an error in this case today for standard slots?)
```
pmdartus, line 388:
```suggestion
* **Missing default slot**: If a parent component slots content into the child component's default slot and the child does not have a default slot, the content will be ignored.(TODO: Why don't we log an error in this case today for standard slots?)

```

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,509 @@
...
+* **Missing named slot**: If a parent component tries to set an unknown named slot of child component, the content will be ignored and an error will be logged in dev mode.
+
+_The above behaviors are consistent with how the engine handles regular slot content._
+
+* **Parent uses scoped slot but child has regular slot**: In this case the content will be ignored and an error will be logged in dev mode.
+* **Parent uses regular slot but child has scoped slot**: This is valid usage. The content provided by the parent will be slotted into the child's scoped slots.
```
pmdartus, line 394: Just to confirm it is the following use case where `<x-list>` has a default scoped slot?

```html
<template>
    <x-child>
        <p>Chocolatier</p>
    </x-child>
</template>
```

How would this work if the child iterates over the default slot, does it means that the `<p>` will be copied multiple time?

text/0000-scoped-slots-light-dom.md
pmdartus: It's unclear whether the `Allowing slot attribute on <template> tag` section answers this question or not.

pmdartus (approved):
No major concerns, let's start with this approach and iterate.

text/0000-scoped-slots-light-dom.md
ravijayaramappa: It's because they share the same name. The ambiguity here is:
1. In native shadow, if a child component has two slots with same name, all the slotted content appear at the first `<slot></slot>`. Should we follow this standard and enforce the same behavior on scoped slots?
2. Or should the engine be intelligent and match the slot content from the parent to the child's slot based on the slot type?
#2 definitely seems wrong IMO. #1 might be good. But to err on the side of caution, let us restrict it and may be open it up later if we find valid usecases.

text/0000-scoped-slots-light-dom.md
ravijayaramappa: Create a [section](https://github.com/salesforce/lwc-rfcs/blob/abdulsattar/scoped-slots/text/0000-scoped-slots-light-dom.md#implications-on-standard-slots) to explicitly call out that there will be no changes to standard slot authoring.

text/0000-scoped-slots-light-dom.md
```diff
@@ -0,0 +1,509 @@
...
+
+However, the same binding can be associated with multiple slots.
+```html
+<template lwc:render-mode="light">
+    <slot lwc:slot-bind={slotdata}></slot>
+    <slot name="slotname1" lwc:slot-bind={slotdata}></slot>
+    <slot name="slotname2" lwc:slot-bind={slotdata}></slot>
+</template>
+```
+_The last one is not ambigious, but for simplifying the restriction and conveying that "slot to binding combination should be unique", it will be restricted._
```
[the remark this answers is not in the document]
ravijayaramappa: yes, it is valid. Made that explicit.

PR #65, opened 2022-07-05
RFC: Light DOM components default to scoped styles

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/light-dom-default-scoped/text/0000-default-scoped-light.md)

### DIFF
--- a/text/0000-default-scoped-light.md
+++ b/text/0000-default-scoped-light.md
@@ -1,0 +1,132 @@
+---
+title: Light DOM components default to scoped styles
+status: DRAFTED
+created_at: 2022-07-05
+updated_at: 2022-07-05
+pr: https://github.com/salesforce/lwc-rfcs/issues/65
+---
+
+# Light DOM components default to scoped styles
+
+## Summary
+
+In the case of [light DOM components](https://rfcs.lwc.dev/rfcs/lwc/0115-light-dom), the default for `*.css` files will be [scoped](https://rfcs.lwc.dev/rfcs/lwc/0116-light-dom-scoped-styles) rather than unscoped. In addition, an explicit `*.unscoped.css` naming convention is added, for cases where unscoped styles are explicitly desired.
+
+## Basic example
+
+The current behavior for scoped styles is as follows:
+
+| Filename       | Result   |
+|----------------|----------|
+| `*.css`        | Unscoped |
+| `*.scoped.css` | Scoped   |
+
+In this system, `*.css` means "unscoped" and `*.scoped.css` means "scoped."
+
+(Note that this scoping mechanism is _not_ the same thing as shadow DOM style scoping, either in synthetic or native mode. What we're talking about here is [class-based style scoping](https://lwc.dev/guide/light_dom#implement-scoped-styles), which works in both light DOM and shadow DOM.)
+
+The proposal is to change this to:
+
+| Filename         | Shadow DOM | Light DOM |
+|------------------|------------|-----------|
+| `*.css`          | Unscoped   | Scoped    |
+| `*.scoped.css`   | Scoped     | Scoped    |
+| `*.unscoped.css` | Unscoped   | Unscoped  |
+
+In this system, `*.css` defaults to "unscoped" for shadow DOM (i.e. the existing behavior, i.e. relying on shadow DOM for scoping), whereas the same filename in light DOM defaults to "scoped."
+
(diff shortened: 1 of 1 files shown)


text/0000-default-scoped-light.md
```diff
@@ -0,0 +1,132 @@
...
+We could _not_ do this proposal, and teach developers how to use scoped styles in the existing system.
+
+Arguably, this approach is not terrible because:
+
+1) Developers already have a lot of new concepts to learn with light DOM (`lwc:render-mode`, `renderMode`, what `:host` means, security, etc.) – this is just one extra concept to learn.
+2) 99% of the time, developers can just remember "Always use `*.scoped.css`," and it will match their intuition. For instance, even in the case of native shadow DOM, using scoped styles doesn't add a huge performance tax, and it avoids the aforementioned edge cases with shadow-parent-light-child and multi-template components.
+
+## Adoption strategy
+
+It will be a breaking change in some cases. We will have to communicate it very clearly.
```
pmdartus, line 124: Since the API is already beta on the Salesforce platform, all the changes have to be backward compatible. This change will need to be behind a compiler flag to avoid potentially breaking existing light DOM components on the platform.

text/0000-default-scoped-light.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Drawbacks
+
+### Breaking existing behavior
+
+Light DOM is considered "beta" and has not officially shipped in LEX yet. However, some folks may be using it off-core. In those cases, the behavior will change, and they will have to fix their existing components, if they were relying on `*.css` being unscoped for light DOM components.
+
+### More complexity in the design
+
+The old system (`*.css` is unscoped, `*.scoped.css` is scoped) was simple to describe. The new system has a 3x3 table to explain how it works.
```
pmdartus, line 99: Super duper nitpick: It's either a 2X3 or a 4X3 matrix if you include the labels. 🤣

```suggestion
The old system (`*.css` is unscoped, `*.scoped.css` is scoped) was simple to describe. The new system has a 2x3 table to explain how it works.

```

gonzalocordero:
Thanks @nolanlawson, I'm totally on board with this proposal. I was wondering, would it make sense to add a section on "other alternatives explored" or similar, by the time I was done reading this I had already forgotten why we couldn't "force" or "default" users to `scoped.css` vs bringing the scoped functionality to the `.css` file.

nolanlawson:
@gonzalocordero Are you talking about the "Make scoped styles the default everywhere" alternative I mention? Or another alternative that we didn't explore? I'm not sure I follow you.

gonzalocordero:
> @gonzalocordero Are you talking about the "Make scoped styles the default everywhere" alternative I mention? Or another alternative that we didn't explore? I'm not sure I follow you.

@nolanlawson yeah, making the  `.scoped` extension to be the default one.

gonzalocordero (approved):

text/0000-default-scoped-light.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Breaking changes for bundlers/adapters
+
+We will need to change the existing behavior in the LWC adapters for Rollup, Jest, Webpack, LWR, lwc-platform, and anywhere else that implements the LWC resolution algorithm. We may break existing unit tests and apps while getting the feature out the door. 
+
+## Alternatives
+
+### Make scoped styles the default everywhere
+
+We cannot do this, because it would potentially break existing shadow DOM components. Even forgetting that light DOM exists, the addition of an extra class selector could change the specificity of the CSS rules, and thus the runtime styling behavior (in both synthetic and native).
```
nolanlawson, line 112: @gonzalocordero Does this section address [your concern](https://github.com/salesforce/lwc-rfcs/pull/65#issuecomment-1176592916)?

nolanlawson:
@brandonferrua @gonzalocordero We just chatted on the LWC team about this (me, @divmain, @jmsjtu, @abdulsattar). Because [light DOM is already GA in experience cloud](https://github.com/salesforce/lwc-rfcs/pull/65#discussion_r914769804), we cannot reasonably implement this RFC without breakages. Furthermore, the new design is more complex than the old design (in terms of both implementation and usage).

A much less invasive change would be to log a warning if unscoped CSS is used in a light DOM component. So for instance, if you use `foo.css`, you'd get a warning saying something like:

    `foo.css` is global when used with a light DOM component. Please use `foo.scoped.css` instead.

This avoids breakages while also avoiding the footgun of accidentally using global CSS in light DOM. What do you think of this solution?

PR #66, opened 2022-07-06
rfc: native web components

ekashida:
W-9110835

### DIFF
--- a/text/0000-third-party-web-components-support.md
+++ b/text/0000-third-party-web-components-support.md
@@ -1,0 +1,103 @@
+---
+title: Native Web Components Support
+status: DRAFTED
+created_at: 2022-07-05
+updated_at: 2022-07-05
+pr: https://github.com/salesforce/lwc-rfcs/pull/66
+---
+
+# Native Web Components Support
+
+## Summary
+
+This proposal is about enabling developers to integrate third party web components by introducing
+LWC support for native web components.
+
+## Motivation
+
+It is common for developers to want to integrate third party web components in an LWC application.
+While there are several ways to do this today (e.g., through the use of iframes or the
+`lwc:dom="manual"` directive), the preferred way would be to enable this use case through the
+rendering of native web components. Browser support for native web components was still in the early
+stages when LWC was first implemented, but it is quite good today. In addition, many frameworks now
+support native web components as a compilation target so it makes sense to add native web component
+support as a first class citizen.
+
+## Detailed design
+
+This section covers the identified work to add native web component support.
+
+### Constructor resolution
+
+By convention, LWC implicitly resolves component constructors for every custom element in its
+template. This behavior needs to be modified to allow authors the flexibility to resolve their own
+components.
+
+#### Proposed solution
+
+LWC would provide a directive (e.g., `lwc:external`) which signals that component resolution should
+be delegated to the browser. LWC will not attempt to resolve components for custom elements tagged
+with this directive. If it happens that the component was not registered, the custom element will
+simply render as an HTMLUnknownElement, which is the default browser behavior.
+
+### Passing data
+
+By convention, custom elements are assumed to be backed by LWC components and values are set through
+their properties. Given that a native web component might expect data passed as an attribute or as a
+property, we would need to support both mechanisms. There are multiple approaches we could take
+here, and the following options are how some popular frameworks do it today. We could default to
+either attributes or properties and introduce a directive that would act as a signal to the engine
+to use one or the other, we could implement a runtime heuristic to choose one over the other, or we
+could default to attributes and set properties when setting a data structure.
+
+#### Proposed solution
+
+When passing data to a native web component, we can take a heuristical approach where if the
+property exists, the property is set. Otherwise the attribute is set. While it may be advantageous
+to require an explicit signal in the form of a directive, this approach seems to strike a good
+balance between developer experience and API semantics.
+
+### Events
+
+EventTarget polyfilled methods and properties such as `add/removeEventListener` and `target` go
+through the synthetic shadow polyfill due to those methods and properties being globally patched.
+This is the current status quo for native shadow roots and should not be an issue for supporting
+third party web components.
+
+LWC's declarative event bindings only support lowercase events. Events using anything other than
+lowercase must be listened to imperatively. This would not need to change in order to support native
+web components.
+
+### Serverside rendering (SSR)
+
+LWC's SSR implementation is tailored specifically for LWC and may not be compatible with a native
+web component implementation. We would probably need to implement the entire synchronous DOM API if
+we wanted to support SSR for native web components.
+
+## Drawbacks
+
+As with any new feature, adding support for native web components introduces additional complexity.
+
+From the platform perspective, there are a few things that Locker vNext needs to implement in order
+to GA this feature. 1) Lift the restriction on `customElements.define()` and 2) provide the
+virtualization mechanism to allow different namespaces to use the same component name.
+
+## Alternatives
+
+No alternatives were considered for the constructor resolution portion of this proposal. An initial
+uninformed idea to introduce error handling for implicit constructor imports was quickly eliminated
+due to the fact that dynamic imports are promise-based.
+
+Alternatives for passing data have already been discussed.
+
+## Adoption strategy
+
+Not much would be required to "sell" this feature as it is something that developers have been
+asking for. Documentation of the new `lwc:external` directive would be required.
+
+# Unresolved questions
+
+- More of a platform-related question, but do we need to make metadata adjustments to exclude native
+  web components from referential integrity?
+- Investigate whether the synthetic shadow slotting implementation is compatible with native web
+  components.
--- a/text/0000-third-party-web-components-support.md
+++ b/text/0000-third-party-web-components-support.md
@@ -112,13 +112,6 @@ If an external element is marked with a directive that identifies it as such, it
 should be rendered and its attributes should be set. Properties will not be set and no attempt will
 be made to render the subtree.
 
-### Custom element conflicts
-
-In environments like the Salesforce platform which employ a multitenant architecture, developers
-will certainly run into custom element conflicts when given the freedom to register their own
-components. This can be managed in some environments (i.e., internal developers or applications
-consuming OSS LWC) but will require virtualization per sandbox before we can GA the feature.
-
 #### Proposed solution
 
 From the platform perspective, there are a few things that Locker vNext needs to implement in order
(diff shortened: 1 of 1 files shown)


text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,103 @@
...
+lowercase must be listened to imperatively. This would not need to change in order to support native
+web components.
+
+### Serverside rendering (SSR)
+
+LWC's SSR implementation is tailored specifically for LWC and may not be compatible with a native
+web component implementation. We would probably need to implement the entire synchronous DOM API if
+we wanted to support SSR for native web components.
+
+## Drawbacks
```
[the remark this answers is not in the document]
caridy: additionally, any blob of HTML that can be cached (fragment) by the static hoisting mechanism will have to be updated, or at least validated, that when a custom element marked as external SHOULD prevent the fragment to be considered hoistable.

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,103 @@
...
+## Adoption strategy
+
+Not much would be required to "sell" this feature as it is something that developers have been
+asking for. Documentation of the new `lwc:external` directive would be required.
+
+# Unresolved questions
+
+- More of a platform-related question, but do we need to make metadata adjustments to exclude native
+  web components from referential integrity?
+- Investigate whether the synthetic shadow slotting implementation is compatible with native web
```
caridy, line 102: I beleive we can assume that because an external custom element is NOT supposed to use synthetic shadow (which I think we can probably enforce somehow... or at least error out), we should be fine here.

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,103 @@
...
+### Events
+
+EventTarget polyfilled methods and properties such as `add/removeEventListener` and `target` go
+through the synthetic shadow polyfill due to those methods and properties being globally patched.
+This is the current status quo for native shadow roots and should not be an issue for supporting
+third party web components.
+
+LWC's declarative event bindings only support lowercase events. Events using anything other than
+lowercase must be listened to imperatively. This would not need to change in order to support native
+web components.
```
nolanlawson, line 69: The test suite for [Custom Elements Everywhere](https://custom-elements-everywhere.com/), which is basically the industry yardstick for "does a framework support web components," explicitly checks to see if events with `names-like-this` or `namesLikeThis` can be listened for in a declarative way. (They don't consider it good enough that you can always call `addEventListener`.)

I think maybe we should consider relaxing our requirement that every event is lowercase without hyphens, since otherwise we won't score 100% on this test.

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,103 @@
...
+web component implementation. We would probably need to implement the entire synchronous DOM API if
+we wanted to support SSR for native web components.
+
+## Drawbacks
+
+As with any new feature, adding support for native web components introduces additional complexity.
+
+From the platform perspective, there are a few things that Locker vNext needs to implement in order
+to GA this feature. 1) Lift the restriction on `customElements.define()` and 2) provide the
+virtualization mechanism to allow different namespaces to use the same component name.
```
nolanlawson, line 83: Just to be clear, the solution we're aiming for aligns with the [pivots PR](https://github.com/salesforce/lwc/pull/2724), correct? If so, it may be good to spell out exactly how this would work (i.e. LWC can make as many components with the same name as it wants, but third-party custom element frameworks are still beholden to global `customElements.define` semantics).

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Alternatives
+
+No alternatives were considered for the constructor resolution portion of this proposal. An initial
+uninformed idea to introduce error handling for implicit constructor imports was quickly eliminated
+due to the fact that dynamic imports are promise-based.
+
+Alternatives for passing data have already been discussed.
```
nolanlawson, line 91: Did we consider the alternative of not requiring an additional directive, but having some kind of fallback logic? E.g.:

1. The developer uses `<x-foo></x-foo>` in a template
2. First, we check if `x-foo` is an LWC component
3. If not, we check if `x-foo` is defined in the global registry
4. If not, we throw an error

Is something like this feasible?

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,103 @@
...
+This is the current status quo for native shadow roots and should not be an issue for supporting
+third party web components.
+
+LWC's declarative event bindings only support lowercase events. Events using anything other than
+lowercase must be listened to imperatively. This would not need to change in order to support native
+web components.
+
+### Serverside rendering (SSR)
+
+LWC's SSR implementation is tailored specifically for LWC and may not be compatible with a native
```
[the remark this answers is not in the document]
nolanlawson: I'm confused about option # 2. Could you elaborate?

For # 1, if you mean "just render `<x-foo data-bar="baz"></x-foo>` as-is, with nothing inside of the element," then that makes sense to me. It will just "hydrate" when the element is upgraded on the client-side. Not really SSR, but good enough given we don't know how to SSR other frameworks' components.

text/0000-third-party-web-components-support.md
ekashida: I added some more context around event names in this rfc. Will schedule the topic in a dev sync and mark this as resolved.

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,121 @@
...
+could default to attributes and set properties when setting a data structure.
+
+#### Proposed solution
+
+When passing data to a native web component, we can take a heuristical approach where if the
+property exists, the property is set. Otherwise the attribute is set. While it may be advantageous
+to require an explicit signal in the form of a directive (both in terms of static analysis and
+performance), this approach seems to strike a good balance between developer experience and API
+semantics. It should be noted that this heuristical approach will require a check on the element's
+prototype to see whether a property is defined, which will incur a runtime performance cost.
```
pmdartus, line 60: >  check on the element's prototype to see whether a property is defined, which will incur a runtime performance cost.

I don't the runtime performance cost to be high. If it's a real issue we could, for example, cache which value be set as a property vs. attribute for follow-up rendering.

text/0000-third-party-web-components-support.md
pmdartus: > * The developer uses `<x-foo></x-foo>` in a template

If the template contains an `<x-foo>` element, the LWC compiler would automatically add an import for the `x/foo` module. If the `x/foo` module doesn't exist the compilation would fail. While you can always create this module in LWC OSS, platform developers only have access to their own `c` namespace.

We need a mechanism to tell the compiler not to attempt to import certain custom elements to avoid running into those resolution issues.

text/0000-third-party-web-components-support.md
caridy: Yes, this will run on top of the polyfill that we have in LWC, but that virtualization only affects a specific sandbox, we should be good here.

text/0000-third-party-web-components-support.md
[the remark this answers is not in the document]
caridy: @nolanlawson I'm talking about this:

```html
<template>
    <x-foo>
        <p></p>
    </x-foo>
</template>
```

vs 

```html
<template>
    <x-foo lwc:external>
        <p></p>
    </x-foo>
</template>
```

If `x-foo` is lwc then SSR will be able to do some work to produce the content of it... and if it has a shadow, it will produce the proper shadow as well... but if x-foo is external, then what? We can add a proprietary mechanism to infer that a shadow is needed, or they can go with the standard mechanism, which is going to be: `<x-foo lwc:external shadow-root="open">` or something like that, I don't remember.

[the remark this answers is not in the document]
ekashida: Doesn't rehydration take care of slotting the `<p></p>` on the client?

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,155 @@
...
+### Events
+
+Today, LWC's declarative event bindings only support lowercase events. Events using anything other
+than lowercase can be listened for using the imperative `addEventListener()` API. There are several
+historical discussions (e.g., [#1811][1811] and [#1904][1904]) about relaxing this constraint;
+however, the current behavior would not need to change in order to support native web components. It
+should be noted that adding support for events has always been about improved interoperability, as
+the browser's imperative API has no restriction regarding the event type.
+
+EventTarget polyfilled methods and properties such as `add/removeEventListener` and `target` go
```
caridy, line 104: I didn't quite get what you're saying here.

text/0000-third-party-web-components-support.md
```diff
@@ -0,0 +1,155 @@
...
+If an external element is marked with a directive that identifies it as such, its custom element
+should be rendered and its attributes should be set. Properties will not be set and no attempt will
+be made to render the subtree.
+
+### Custom element conflicts
+
+In environments like the Salesforce platform which employ a multitenant architecture, developers
+will certainly run into custom element conflicts when given the freedom to register their own
+components. This can be managed in some environments (i.e., internal developers or applications
+consuming OSS LWC) but will require virtualization per sandbox before we can GA the feature.
```
caridy, line 120: I don't think LWC should care here... this is a platform thing. Also, I don't think virtualization capabilities by locker should gate this because today 2nd party and 3rd party components cannot register componnets anyway, so it should be fine... the only problem is really a platform registered component that can now be created by a namespace, which is fine I believe, we don't expect a bunch of people to register components in system mode anyways, and they can get to them by flashlight if that the case... I don't know.

text/0000-third-party-web-components-support.md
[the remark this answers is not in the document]
ekashida: Sorry, I guess I'm a little dense today. I think I'm missing an implicit assumption in this comment: `in fact, it will just produce the element as HTML, without explicit knowledge about what goes inside it... making it more tricky to determine if the shadow should be there or not`. Isn't this always the case regardless of SSR? Any custom element marked as external is simply rendered without looking for its constructor. Why do we need to know if a shadow root might be involved?

text/0000-third-party-web-components-support.md
ekashida: I'm basically saying that I investigated a little to see if the synthetic shadow polyfill for events might need to be adjusted and found that no work was required there.

leobalter:
@nolanlawson @pmdartus @caridy it seems like @ekashida address most of the feedback here. 

I see a nit pick from @caridy here: https://github.com/salesforce/lwc-rfcs/pull/66#discussion_r921616139 that might need - or not - removing a paragraph.

Do we need anything else?

text/0000-third-party-web-components-support.md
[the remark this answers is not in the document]
ekashida: @caridy marking this as resolved, let me know if you believe if we need to add more detail here.

ekashida:
Merging this for now, will be happy to address any further comments in another PR.

PR #67, opened 2022-08-19
fix: update template refs with better backwards compat

nolanlawson:
This updates the "template refs" proposal with betters backwards compatibility with existing Lightning Web Components.

### DIFF
--- a/text/0000-refs.md
+++ b/text/0000-refs.md
@@ -117,6 +117,10 @@ In the case of shadow DOM components, `this.refs` refers to elements inside of t
 
 To preserve backwards compatibility with existing components that may have defined a `refs` property, `refs` will be configurable and writable, so if a component defines its own `refs`, it will replace the one from `LightningElement.prototype`.
 
+To further improve backwards compatibility, the `refs` property will only be defined if the directive `lwc:ref` is present somewhere in the template. If `lwc:ref` is not defined in the template, then `this.refs` returns undefined.
+
+In the case of a component that renders multiple templates, the presence or absence of `lwc:ref` is checked on-demand in the `this.refs` getter. So whether `this.refs` is undefined or an object will depend on the current template when the `this.refs` getter is invoked.
+
 ### Multiple templates
 
 In the case of multiple templates, the same rules as for `this.template.querySelector()` or `this.querySelector()` apply. Once the new template is rendered, `this.refs` refers to elements defined in the new template:


nolanlawson:
[This commit](https://github.com/salesforce/lwc/pull/2690/commits/ee17841728a0a9b313e5b40f7c445a4752680183) implements the proposed change.

The _only way_ I can imagine that this would still cause an observable change is if someone is doing:

```js
if ('refs' in this) {}
```

...rather than:

```js
if (!this.refs) {}
```

This possibility seems so vanishingly small that I can't imagine it's actually a problem in practice. Developers are much more likely to use the shorter (second) one rather than the longer (first) one.

nolanlawson:
One downside of this change is that we would be introducing a footgun for component authors using refs and multiple templates:

```html
<!-- template A -->
<template>
  <div lwc:ref="foo"></div>
</template>
```

```html
<!-- template B -->
<template>
  <div></div>
</template>
```

```js
if (this.refs.foo) // throws an error when template B is rendered
```

The fix is relatively simple, though:

```js
if (this.refs?.foo)
```

And hopefully this isn't a very big deal in practice.

caridy:
Ok, I still don't understand what the problem is in the first place... yes, the `in` operator or any other way to observe that there is a NEW api in the base element is observable but that doesn't mean it is going to be a problem.

This proposal it more problematic IMO because it only works for default templates if any since the details of the html are not always associated to a class declaration, how can you know if you need the ref or not?

nolanlawson:
@caridy

> Ok, I still don't understand what the problem is in the first place... yes, the `in` operator or any other way to observe that there is a NEW api in the base element is observable but that doesn't mean it is going to be a problem.

Yeah I admit it's a stretch. The reason it might be a problem is something like this:

```js
connectedCallback() {
  if (!('refs' in this)) {
    this.refs = 'foo'
  }
  // user expects this.refs to be a string, or a boolean, or whatever, but it's undefined
  // because LWC has now claimed it
  this.message = this.refs
}
```

Again, it's a stretch.

> This proposal it more problematic IMO because it only works for default templates if any since the details of the html are not always associated to a class declaration, how can you know if you need the ref or not?

This proposal works with dynamic templates as well. When the user calls the `this.refs` getter, we check on-demand for the current state of the refs. And whatever template rendered last will have set this state.

The state is updated [here](https://github.com/salesforce/lwc/blob/ee17841728a0a9b313e5b40f7c445a4752680183/packages/%40lwc/engine-core/src/framework/template.ts#L275-L277) when the template is invoked (i.e. `html.call()`).

And then the `this.refs` getter refers to it [here](https://github.com/salesforce/lwc/blob/ee17841728a0a9b313e5b40f7c445a4752680183/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L503).

caridy:
I don't understand... if I use the `in` operator in the constructor... how can you know that it should be there or not before getting the template to be renderer?

nolanlawson:
@caridy Imagine that someone is using a property called `refs` today. Maybe they expect it to be a string:

```js
class extends LightningElement {
  constructor() {
    if (!('refs' in this)) {
      this.refs = 'foo'
    }
    console.log(this.refs.toUpperCase()) // throws if we ship template refs
  }
}
```

The above code works today, but it throws as soon as we ship template refs, because we're now squatting on the `refs` property. The user expects `refs` to be a string, but with template refs, it's an object (or `undefined`, with this PR).

Again, it's not very likely, but it is possible.

caridy:
I will NOT consider this an issue, for few main reasons:

* Imagine that tomorrow HTML adds a new property to `HTMLElement.prototype`, you will have the exact same issue because we pipe all global props.
* We have explicit documentation discouraging the use of expandos, in fact, the `refs` in the example above is not even reactive.
* Non-declared fields are not reactive, we have documentation for that... so the usage of non-declared fields should be very low... then on top of that to match that name `refs` is even lower.

nolanlawson:
> Imagine that tomorrow HTML adds a new property to `HTMLElement.prototype`, you will have the exact same issue because we pipe all global props.

I don't think this is actually the case. Looking at `base-lightning-element.ts`, it seems we apply [only a subset](https://github.com/salesforce/lwc/blob/b1c2b2089ccc5cb650afd4e257b7841edeab43d7/packages/%40lwc/engine-core/src/framework/attributes.ts#L10-L20) of `HTMLElement` properties to `LightningElement`. It looks like we document this as [`HTMLElementTheGoodParts`](https://github.com/salesforce/lwc/blob/23027c15fe872aba0cd26529018dd5f43d303406/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L132-L168).

For instance, you can test this with a relatively recent addition to `Element.prototype`, [`part`](https://developer.mozilla.org/en-US/docs/Web/API/Element/part):

```js
connectedCallback() {
  console.log(this.part) // logs undefined
}
```

> We have explicit documentation discouraging the use of expandos, in fact, the `refs` in the example above is not even reactive.

This is a valid point, although unfortunately people do use expandos, hence the concern about observerable changes.

caridy:
@nolanlawson if we don't keep those two lists up to date, a dev will have no ways to use those attributes, nor to customize their behavior, so this is a bullet that we will have to bite.

nolanlawson:
@caridy My understanding was that, historically, we made a distinction between the `LightningElement` when accessed internally versus externally. So internally you don't have access to things like `this.shadowRoot`, but externally you do have access to `elm.shadowRoot`. Are you saying we should change that distinction and allow access to all `HTMLElement` props via `this`?

nolanlawson:
@jodarove @pmdartus @caridy Could I get a +1 on this PR please?

jodarove (approved):

PR #68, opened 2022-08-25
feat: Declarative Shadow DOM polyfill

jodarove:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/jodarove/dec-shadow-dom-polyfill/text/0000-declarative-shadow-dom-polyfill.md)


text/0000-declarative-shadow-dom-polyfill.md
```diff
@@ -0,0 +1,346 @@
...
+                    </template><script type="text/javascript">replaceTemplateWithShadowRoot(document.currentScript);</script>
+                </x-btn>
+            </div>
+        </template><script type="text/javascript">replaceTemplateWithShadowRoot(document.currentScript);</script>
+    </x-foo>
+    */
+}
+```
+
+Other alternative solutions were taken into account, but at the moment there’s not a strong use case for them, and other than removing the extra string manipulation operation, there’s no added benefit.
```
nolanlawson, line 276: Based on the benchmarks and the tradeoffs, this sounds reasonable to me. Assuming there is no other way for a `<template>` to sneak into the SSR output, the best approach seems to be to just ask container to do regex replacement. Otherwise we are offering a lot of API surface to maintain, which is not really much better than regex replacement.

text/0000-declarative-shadow-dom-polyfill.md
```diff
@@ -0,0 +1,346 @@
...
+**CHROME**
+Despite Chrome natively supporting declarative shadow dom, we measure the speeds of the polyfills, because in the case of the inline script, the script element must be removed.
+
+<figure>
+  <img src="../assets/declarative-shadow-dom-polyfill/perf-chrome.png" alt="Performance comparison in Google Chrome">
+  <figcaption align = "center">
+    <b>Fig.1: Performance comparison in Google Chrome</b>
+  </figcaption>
+</figure>
+<br></br>
```
nolanlawson, line 207: I don't understand how it's possible for the single loop script to be slower than the inline script in Chrome. With the inline script, the browser needs to halt parsing after every `<template>` just to do a useless action (removing the `<script>`). Whereas with the single loop script, it does one useless action at the end (check `supportsDeclarativeShadowDOM`, then exit).

text/0000-declarative-shadow-dom-polyfill.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Single loop script shadow root attachment ([reference](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md#baseline-2---single-script-based-shadow-root-attachment))
+
+Uses a single script at the end of the HTML that loops over all templates to attach the shadow roots and move the template contents into each root.
+
+This approach is optimized for speed - the entire page is parsed first, and then one script does all of the shadow root attachment and population. The advantage here is that the parser is not blocked for `script` on each shadow root. The major **disadvantages** are:
+
+1. Streaming is no longer possible. The entire page will consist of inert `<template>` elements that do not render, until the final script loops through and "converts" them into shadow roots.
+2. Content shifts will likely occur. The initially-parsed result, if rendered, will not contain the shadow content. Once the final loop runs, and shadow roots are attached, the page layout will likely change, leading to CLS problems.
```
nolanlawson, line 82: It's worth noting that all these downsides only exist for non-Chromium browsers. In Chromium, we will still get streaming and no content shifts.

text/0000-declarative-shadow-dom-polyfill.md
nolanlawson: I'm also in favor of leaving the decision up to containers. It's up to them to decide and weigh the tradeoffs of Chromium vs other browsers, streaming vs end-to-end performance, etc.

nolanlawson (approved):

jodarove:
adding @tedconn @priandsf @nrobertdehault for their input.

text/0000-declarative-shadow-dom-polyfill.md
```diff
@@ -0,0 +1,346 @@
...
+## Polyfills
+
+This section describes different terms and solutions to polyfill the Declarative ShadowDOM spec. It will define invariants for any of the described polyfill implementations, and compare them using the common comparison points when polyfilling the spec.
+
+### Invariants
+
+1. We would only support native shadow DOM when rendering declarative shadow DOM, synthetic shadow DOM will be out of the picture.
+2. The `getInnerHTML` method from the spec will not be polyfilled.
+3. The new `ElementInternals.shadowRoot` will not be polyfilled.
+4. For completeness, each polyfill removes the `<template>` element (and the inline `<script>` node when needed), so that the resulting tree is identical to the declarative output.
```
pmdartus, line 65: Removing the `<template>` element is an implementation detail. I would suggest being a little more general for the invariant here.

```suggestion
4. The declarative shadow DOM polyfill should produce the same final DOM tree as its native counterpart.
```

text/0000-declarative-shadow-dom-polyfill.md
```diff
@@ -0,0 +1,346 @@
...
+Compare the single loop example to the inline script in the [supplementary repository](https://github.com/jodarove/dec-shadow-dom-polyfill-spike#examples).
+
+
+```js
+<body>
+    <div>
+        <template shadowroot="open"></template>
+        <script type="text/javascript">
+            replaceTemplateWithShadowRoot(document.currentScript);
+        </script>
```
pmdartus, line 93: This is not needed for this use case, correct?

```suggestion
```

pmdartus (approved):

text/0000-declarative-shadow-dom-polyfill.md
jodarove: correct!

PR #71, opened 2023-01-03
RFC: Dynamic Component Creation

jmsjtu:
[Rendered](https://github.com/jmsjtu/lwc-rfcs/blob/jtu/dynamic-component-creation/text/0119-dynamic-component-creation.md)

### DIFF
--- a/text/0119-dynamic-component-creation.md
+++ b/text/0119-dynamic-component-creation.md
@@ -219,14 +219,16 @@ _Note in the case of lazy loading, the component author is responsible for resol
 
 #### Selecting the dynamic component
 
-Dynamic components have certain properties that will not work with `this.template.querySelector`.  First, the `<lwc:component>` tag will not be rendered to the DOM which means it is not available to be selected using `this.template.querySelector`.  Second, the LWC engine does not provide a way to retrieve the dynamic component's custom element name for use as a selector with `this.template.querySelector` (we recommend using `lwc:ref` instead).
+Dynamic components have certain properties that will not work with `this.template.querySelector`:
+1. The `<lwc:component>` tag will not be rendered to the DOM which means it is not available to be selected using `this.template.querySelector`. 
+2. The LWC engine does not provide a way to retrieve the dynamic component's custom element name for use as a selector with `this.template.querySelector` (we recommend using `lwc:ref` instead).
 
 To select the dynamic component, the actual custom element must be selected once it has been rendered to the DOM by using the [`lwc:ref`](https://rfcs.lwc.dev/rfcs/lwc/0000-refs) directive or through another attribute assigned to the component.
 
 Here are some ways component authors can detect when a dynamic component has mounted:
-1. Use `connectedCallback` on the dynamic component to signal when it has mounted.
-2. A dynamic component constructor is guaranteed to be mounted in the next rendering cycle once it has been set.  When it is set, the parent component can wait until the `renderedCallback` lifecycle method is invoked to detect when the dynamic component is mounted.
-3. Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect when the dynamic component has mounted.
+- Use `connectedCallback` on the dynamic component to signal when it has mounted.
+- A dynamic component constructor is guaranteed to be mounted in the next rendering cycle once it has been set.  When it is set, the parent component can wait until the `renderedCallback` lifecycle method is invoked to detect when the dynamic component is mounted.
+- Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect when the dynamic component has mounted.
 
 ```html
 <template>
--- a/text/0119-dynamic-component-creation.md
+++ b/text/0119-dynamic-component-creation.md
@@ -190,7 +190,7 @@ _Note in the case of lazy loading, the component author is responsible for resol
 
 #### Selecting the dynamic component
 
-It is important to note that `<lwc.dynamic>` will not be rendered to the DOM.  As such, it is not available to be selected using `this.template.querySelector`.  The actual custom element must be selected once it has been rendered to the DOM.  The dynamic component can be selected either through the `lightningElementName` property on the constructor or through another attribute assigned to the component.
+It is important to note that the `<lwc.dynamic>` element itself will not be rendered to the DOM.  As such, it is not available to be selected using `this.template.querySelector`.  The actual custom element must be selected once it has been rendered to the DOM.  The dynamic component can be selected either through the `lightningElementName` property on the constructor or through another attribute assigned to the component.
 
 Component authors can use the `connectedCallback` of the dynamic component to signal when the component has been rendered to the DOM.
 
--- a/text/0119-dynamic-component-creation.md
+++ b/text/0119-dynamic-component-creation.md
@@ -28,13 +28,13 @@ Dynamic components refer to custom elements where the constructor is not known a
 
 ```javascript
 import { LightningElement } from 'lwc';
+import ConcreteComponent from 'lightning/concreteComponent';
 
 export default class extends LightningElement {
     lazyConstructor;
 
-    async connectedCallback() {
-        const { default: ctor } = await import('lightning/concreteComponent');
-        this.lazyConstructor = ctor;
+    connectedCallback() {
+        this.lazyConstructor = ConcreteComponent;
     }
 }
 ```
@@ -65,33 +65,37 @@ For example:
 
 ```html
 <template>
-  <x-lazy lwc:dynamic={customCtor}></x-lazy>
+    <button onclick={loadCtor}>Click to change constructor</button>
+    <x-lazy lwc:dynamic={customCtor}></x-lazy>
 </template>
 ```
 
 ```javascript
 import { LightningElement } from "lwc";
+import About from 'c/about';
+import Home from 'c/home';
 
-export default class DynamicCtor extends LightningElement {
-  customCtor;
+export default class extends LightningElement {
+    customCtor;
+    ctors = [About, Home]
 
-  connectedCallback() {
-    this.loadCtor();
-  }
+    loadCtor() {
+        if (this.customCtor) {
+            this.ctors.push(this.customCtor);
+        }
 
-  async loadCtor() {
-    const { default: ctor } = await import("c/customConstructor");
-    this.customCtor = ctor;
-  }
+        this.customCtor = this.ctors.shift();
+    }
 }
 ```
 
 Will always render the same tag regardless of the constructor.
 
 ```html
-  <x-lazy>
+<button onclick={loadCtor}>Click to change constructor</button>
+<x-lazy>
     <!-- c-custom-constructor content -->
-  </x-lazy>
+</x-lazy>
 ```
 
 This proposal aims to overcome the issues with the current design by exposing the custom element name at compile time and using it as the name of the dynamic component that is created at runtime.
--- a/text/0119-dynamic-component-creation.md
+++ b/text/0119-dynamic-component-creation.md
@@ -142,7 +142,7 @@ namespace/
 
 The LWC module resolver also allows for namespace and name mappings that do not adhere to the default folder structure, such as the case with [alias module records](https://rfcs.lwc.dev/rfcs/lwc/0020-module-resolution#aliasmodulerecord).
 
-Leveraging this, the namespace and name can be resolved at compile time and used to create the custom element name, which will be in the form `namespace-name`.  
+Leveraging this, the namespace and name can be resolved at compile time and used to construct the custom element name, which will be in the form `namespace-name`.  
 
 ##### `componentRegisteredNameMap`
 
@@ -213,25 +213,21 @@ This means that when the value provided to `lwc:is` changes the component and al
 
 - **The constructor is falsy**
     - When the constructor is falsy the `<lwc:component>` tag along with all of its children are not rendered to the DOM.
-- **The constructor is defined and is not a `LightningElementConstructor`**
+- **The constructor is defined and is not a `LightningElement` constructor**
     - When the value provided to `lwc:is` is not a valid `LightningElement` constructor then an error is thrown.
-- **The constructor is defined and is a `LightningElementConstructor`**
+- **The constructor is defined and is a `LightningElement` constructor**
     - When the constructor is a valid `LightningElement` constructor, the component will be rendered in place of `<lwc:component>`. The tag name used for the dynamic component is the value returned from `getComponentRegisteredName` for the given constructor.
 
-_Note in the case of lazy loading, the component author is responsible for resolving the promise.  The only value that can be given to `lwc:is` is a `LightningElementConstructor`._
+_Note in the case of lazy loading, the component author is responsible for resolving the promise.  The only value that can be given to `lwc:is` is a `LightningElement` constructor._
 
 #### Selecting the dynamic component
 
-Dynamic components have certain properties that will not work with `this.template.querySelector`:
-1. The `<lwc:component>` tag will not be rendered to the DOM which means it is not available to be selected using `this.template.querySelector`. 
-2. The LWC engine does not provide a way to retrieve the dynamic component's custom element name for use as a selector with `this.template.querySelector` (we recommend using `lwc:ref` instead).
-
-To select the dynamic component, the actual custom element must be selected once it has been rendered to the DOM by using the [`lwc:ref`](https://rfcs.lwc.dev/rfcs/lwc/0000-refs) directive or through another attribute assigned to the component.
+To select the dynamic component, the actual custom element must be selected once it has been rendered to the DOM by using the [`lwc:ref`](https://rfcs.lwc.dev/rfcs/lwc/0000-refs) directive or through another attribute assigned to the component such as a class name.
 
 Here are some ways component authors can detect when a dynamic component has mounted:
 - Use `connectedCallback` on the dynamic component to signal when it has mounted.
 - A dynamic component constructor is guaranteed to be mounted in the next rendering cycle once it has been set.  When it is set, the parent component can wait until the `renderedCallback` lifecycle method is invoked to detect when the dynamic component is mounted.
-- Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect when the dynamic component has mounted.
+
 
 ```html
 <template>
@@ -305,9 +301,13 @@ export default class extends LightningElement {
 ```
 #### Usage with other LWC directives
 
-All [LWC custom element directives](https://lwc.dev/guide/reference#directives-for-nested-templates) with the exception of `lwc:external` are available to use on `<lwc:component>` and their functionality will be passed through to the custom element once it has been created.
+All directives for [nested templates](https://lwc.dev/guide/reference#directives-for-nested-templates) are available to use on `<lwc:component>` and their functionality will be passed through to the custom element once it has been created.
 
-`lwc:external` will not be available because the constructor provided to the `lwc:is` directive must be a `LightningElementConstructor`.
+Additionally, the following directives will also be supported:
+- [`lwc:spread`](https://lwc.dev/guide/reference#lwc%3Aspread%3D%7Bchildprops%7D)
+- [`lwc:ref`](https://rfcs.lwc.dev/rfcs/lwc/0000-refs)
+
+_Note `lwc:external` will not be available because the constructor provided to the `lwc:is` directive must be a `LightningElement` constructor._
 
 #### Children of the dynamically created element
 
@@ -350,7 +350,7 @@ In both cases the dynamic component will be fully mounted and unmounted.
 
 #### Styles
 
-Something to keep in mind is that [light DOM styles are injected to the closest root node](https://rfcs.lwc.dev/rfcs/lwc/0115-light-dom#styles) and are not removed once the components are unmounted. This means that repeated mounting and mounting of a dynamic component could cause styles to be overwritten.  This can be avoided by using either scoped styles or shadow dom, as the styles will be scoped.
+Something to keep in mind is that [light DOM styles are injected to the closest root node](https://rfcs.lwc.dev/rfcs/lwc/0115-light-dom#styles) and are not removed once the components are unmounted. This means that repeated mounting and unmounting of dynamic component could cause styles to be overwritten.  This can be avoided by using either scoped styles or shadow dom, as the styles will be scoped.
 
 ## Drawbacks
 
--- a/text/0119-dynamic-component-creation.md
+++ b/text/0119-dynamic-component-creation.md
@@ -28,13 +28,14 @@ Dynamic components refer to custom elements where the constructor is not known a
 
 ```javascript
 import { LightningElement } from 'lwc';
-import ConcreteComponent from 'lightning/concreteComponent';
 
 export default class extends LightningElement {
     lazyConstructor;
 
     connectedCallback() {
-        this.lazyConstructor = ConcreteComponent;
+        import('lightning/concreteComponent')
+            .then(({ default: ctor }) => this.lazyConstructor = ctor)
+            .catch(err => console.log('Error importing component'));
     }
 }
 ```
@@ -77,13 +78,15 @@ import Home from 'c/home';
 
 export default class extends LightningElement {
     customCtor;
-    ctors = [About, Home]
+    ctors = [About, Home];
 
     loadCtor() {
         if (this.customCtor) {
+            // Append to the end of the array
             this.ctors.push(this.customCtor);
         }
 
+        // Remove the first element in the array
         this.customCtor = this.ctors.shift();
     }
 }
@@ -98,7 +101,7 @@ Will always render the same tag regardless of the constructor.
 </x-lazy>
 ```
 
-This proposal aims to overcome the issues with the current design by exposing the custom element name at compile time and using it as the name of the dynamic component that is created at runtime.
+This proposal aims to overcome the issues with the current design by resolving the custom element name at compile time and using it as the name of the dynamic component that is created at runtime.
 
 ## Prior art
 
@@ -116,7 +119,7 @@ The goal of managing dynamic components in LWC is to follow the same component l
 
 ## Detailed design
 
-A new property, `lightningElementName` on `LightningElementConstructor` is introduced to store the name of the custom element.
+A new internal `Map` called `componentRegisteredNameMap` is introduced to store the custom element name and an internal API `getComponentRegisteredName` is introduced to retrieve the custom element name.
 
 A special tag, `<lwc:component>` along with a new template directive `lwc:is` are also introduced and serves as the anchor in the DOM where the dynamic component will be rendered.
 
@@ -140,16 +143,42 @@ namespace/
     ├── *.html
 ```
 
-Leveraging this, the namespace and name can be accessed at compile time and ultimately stored on the constructor.  A new read-only property on `LightningElementConstructor` called `lightningElementName` will store the custom element name which will be of the form `namespace-name`.
+The LWC module resolver also allows for namespace and name mappings that do not adhere to the default folder structure, such as the case with [alias module records](https://rfcs.lwc.dev/rfcs/lwc/0020-module-resolution#aliasmodulerecord).
+
+Leveraging this, the namespace and name can be resolved at compile time and used to create the custom element name, which will be in the form `namespace-name`.  The custom element name will be stored on an internal `Map` called `componentRegisteredNameMap` where the key is the `LightningElementConstructor` and the value is the custom element name.
+
+At compile time, LWC will inject a call to `registerComponent` providing the custom element name.  At runtime `registerComponent` will associate the custom element name to `componentRegisteredNameMap` in a similar way to how templates are associated to `LightningElementConstructor`.
+
+If the custom element name is unable to be resolved at compile time and no value is provided to `registerComponent` the compiler will report an error.
+
+In contrast, the custom element name can be retrieved through an internal API called `getComponentRegisteredName`.  The LWC engine will use this API to retrieve the name when the dynamic component is ready to be instantiated.
+
+_Note `getComponentRegisteredName` should be the only way to retrieve the custom element name._
+
+The `componentRegisteredNameMap`, `getComponentRegisteredName`, and custom element name are internal APIs that should only be used by the LWC engine and will not be observable to component authors.  This is to prevent any unintended side effects.
+
+_See the [Selecting the dynamic component](#Selecting-the-dynamic-component) for details on how to access the component once it has been instantiated._
+
+##### Considerations
+
+There is a possibility that a constructor is mapped to more than one alias, such as:
 
 ```javascript
-const { default: ctor } = await import("c/customConstructor");
-// The tag name can be directly accessed on the constructor
-ctor.lightningElementName
+{
+    "modules": [
+       {
+            "name": "ui/button",
+            "path": "src/modules/ui/button/button.js"
+        },
+        {
+            "name": "ui/button2",
+            "path": "src/modules/ui/button/button.js"
+        }
+    ]
+}
 ```
-At compile time LWC will inject a call to `registerComponent` providing the custom element name.  At runtime `registerComponent` will associate the custom element name to `lightningElementName` in a similar way to how templates are associated to `LightningElementConstructor`.  The value of `lightningElementName` should only be set by LWC to prevent component authors from modifying it and producing unintended side effects.
 
-Alternatively, the custom element name can be stored in a `weakMap` and a public API can be exposed to retrieve the constructor's custom element name. However, because the custom element name can be used as a selector with `this.template.querySelector` placing it on the constructor seems more ergonomic.
+Since there is no way to know which alias to use in this case, the custom element name will be the first alias that is resolved.  This should be an edge case that does not appear often.
 
 #### Defining an anchor in the DOM
 
@@ -180,22 +209,52 @@ Dynamic component instantiation follows the same creation path as any other LWC 
 This means that when the value provided to `lwc:is` changes the component and all of its children will be destroyed and recreated using the new constructor.  In addition, when the value provided to `lwc:is` is evaluated, the LWC engine will render depending on the following scenarios:
 
 - **The constructor is falsy**
-    - When the constructor is undefined the `<lwc:component>` tag along with all of its children are not rendered to the DOM.
+    - When the constructor is falsy the `<lwc:component>` tag along with all of its children are not rendered to the DOM.
 - **The constructor is defined and is not a `LightningElementConstructor`**
     - When the value provided to `lwc:is` is not a `LightningElementConstructor` then an error is thrown indicating the constructor value must be `LightningElementConstructor`.
 - **The constructor is defined and is a `LightningElementConstructor`**
-    - When the constructor is a `LightningElementConstructor`, the component will be rendered in place of `<lwc:component>`. The tag name used for the dynamic component is the value of `lightningElementName` on `LightningElementConstructor`.
+    - When the constructor is a `LightningElementConstructor`, the component will be rendered in place of `<lwc:component>`. The tag name used for the dynamic component is the value returned from `getComponentRegisteredName`.
 
 _Note in the case of lazy loading, the component author is responsible for resolving the promise.  The only value that can be given to `lwc:is` is a `LightningElementConstructor`._
 
 #### Selecting the dynamic component
 
-It is important to note that the `<lwc:component>` element itself will not be rendered to the DOM.  As such, it is not available to be selected using `this.template.querySelector`.  The actual custom element must be selected once it has been rendered to the DOM.  The dynamic component can be selected either through the `lightningElementName` property on the constructor or through another attribute assigned to the component.
+Dynamic components have certain properties that will not work with `this.template.querySelector`.  First, the `<lwc:component>` tag will not be rendered to the DOM which means it is not available to be selected using `this.template.querySelector`.  Second, the LWC engine does not provide a way to retrieve the dynamic component's custom element name for use as a selector with `this.template.querySelector` (we recommend using `lwc:ref` instead).
+
+To select the dynamic component, the actual custom element must be selected once it has been rendered to the DOM by using the [`lwc:ref`](https://rfcs.lwc.dev/rfcs/lwc/0000-refs) directive or through another attribute assigned to the component.
 
 Here are some ways component authors can detect when a dynamic component has mounted:
 1. Use `connectedCallback` on the dynamic component to signal when it has mounted.
 2. A dynamic component constructor is guaranteed to be mounted in the next rendering cycle once it has been set.  When it is set, the parent component can wait until the `renderedCallback` lifecycle method is invoked to detect when the dynamic component is mounted.
 3. Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect when the dynamic component has mounted.
+
+```html
+<template>
+    <lwc:component lwc:is={lazyConstructor} lwc:ref="foo"></lwc:component>
+</template>
+```
+
+```javascript
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    lazyConstructor;
+
+    connectedCallback() {
+        import('lightning/concreteComponent')
+            .then(({ default: ctor }) => this.lazyConstructor = ctor)
+            .catch(err => console.log('Error importing component'));
+    }
+
+    renderedCallback() {
+        // this.refs.foo will be available on the next rendering cycle after the constructor has been set.
+        if (this.refs.foo) {
+            // this.refs.foo will contain a reference to the DOM node
+            console.log(this.refs.foo);
+        }
+    }
+}
+```
 
 #### Assigning attributes and event listeners
 
@@ -213,22 +272,30 @@ Declaratively:
 
 Imperatively:
 
+```html
+<template>
+    <lwc:component lwc:is={lazyConstructor} lwc:ref="foo"></lwc:component>
+</template>
+```
+
 ```javascript
 import { LightningElement } from "lwc";
 
-export default class DynamicCtor extends LightningElement {
-  customCtor;
+export default class extends LightningElement {
+    lazyConstructor;
 
-  connectedCallback() {
-    this.loadCtor();
-  }
+    connectedCallback() {
+        import('lightning/concreteComponent')
+            .then(({ default: ctor }) => this.lazyConstructor = ctor)
+            .catch(err => console.log('Error importing component'));
+    }
 
-  async loadCtor() {
-    const { default: ctor } = await import("c/customConstructor");
-    this.customCtor = ctor;
-    this.template.querySelector(ctor.lightningElementName).setAttribute('title', 'hawaii');
-    this.template.querySelector(ctor.lightningElementName).addEventListener('click', () => console.log('hello world'));
-  }
+    renderedCallback() {
+        if (this.refs.foo) {
+            this.refs.foo.setAttribute('title', 'hawaii');
+            this.refs.foo.addEventListener('click', () => console.log('hello world!'));
+        }
+    }
 }
 ```
 #### Usage with other LWC directives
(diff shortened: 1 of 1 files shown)


text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+Broken down, there are three main parts to this design:
+
+1. Storing and retrieving a custom element's name.
+2. Defining an anchor in the DOM.
+3. Instantiating the dynamic custom element.
+
+#### Storing and retrieving a custom element's name
+
+In today's world, an LWC module's namespace and name are known at compile time by [examining the directory structure](https://rfcs.lwc.dev/rfcs/lwc/0101-module-unification) and through [module resolution](https://rfcs.lwc.dev/rfcs/lwc/0020-module-resolution).
```
pmdartus, line 127: Side note: I am suspecting that we might have to do some work in the `@lwc/rollup-plugin`. I know that in the past the `<namespace>/<name>/<name>.js` folder structure was not always respected in off-core projects. It's something that we might have fixed over the years.

https://github.com/salesforce/lwc/blob/3fafcac1ae4b1887c1b5213d076d02d559dd7549/packages/%40lwc/rollup-plugin/src/index.ts#L214-L215

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+```
+namespace/
+└── name/
+    ├── *.css
+    ├── *.js
+    ├── *.html
+```
+
+Leveraging this, the namespace and name can be accessed at compile time and ultimately stored on the constructor.  A new property on `LightningElementConstructor` called `lightningElementName` will store the custom element name which will be of the form `namespace-name`.
```
pmdartus, line 139: Let's make sure that the component author doesn't attempt to override by making it read-only if possible.

```suggestion
Leveraging this, the namespace and name can be accessed at compile time and ultimately stored on the constructor.  A new read-only property on `LightningElementConstructor` called `lightningElementName` will store the custom element name which will be of the form `namespace-name`.
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+```javascript
+const { default: ctor } = await import("c/customConstructor");
+// The tag name can be directly accessed on the constructor
+ctor.lightningElementName
+```
+
+The value of `lightningElementName` should only be set by the LWC engine to prevent component authors from modifying it and producing unintended side effects.
```
pmdartus, line 147: This section should also discuss where this property is set. I wasn't able to attend the last dev sync, but I was under the impression that the compiler would attach the tag name to the component constructor via `registerComponent`.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+#### Defining an anchor in the DOM
+
+##### `<lwc.dynamic>` tag
+
+`<lwc.dynamic>` is a placeholder tag for the dynamic component and will not be rendered to the DOM.  It serves as a special signal to the compiler that the dynamic component will be rendered at the specific location in the DOM. (Similar to how `template` is not rendered)
+
+Note the separation of the namespace and name using a '.' is intentional because of the following reasons:
+
+1. It is neither a valid [tag name](https://html.spec.whatwg.org/multipage/syntax.html#syntax-tag-name) nor a valid [custom element name](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name) which should clearly signal that this is a special element (which will not be rendered to the DOM).
+2. Initially considered using a ':' instead of a '.' but ':' is the delimiter Aura uses to separate namespace and name.  To prevent confusion on whether this is an Aura tag, decided to use '.' instead.
```
pmdartus, line 160: My 2 cents: While I understand the point about confusion with Aura components, in my opinion, it would be more idiomatic to use a `:` instead of a `.` here as it maps well to the concept of XML namespace. Since we would only permit `<lwc:dynamic>` for now, and maybe other `<lwc:*>` in the future, the compiler could log errors and warning messages when this syntax is misused.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+Additionally, `<lwc.dynamic>` is intended to be used in tandem with `lwc:is`, otherwise `<lwc:dynamic>` serves no purpose. A warning should be issued when `<lwc.dynamic>` is used without the `lwc:is` template directive.
```
pmdartus, line 162:
```suggestion
Additionally, `<lwc.dynamic>` is intended to be used in tandem with `lwc:is`, otherwise `<lwc:dynamic>` serves no purpose. The compiler would report and error when `<lwc.dynamic>` is used without the `lwc:is` template directive.
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+A new template directive called `lwc:is` will be used to signal to compiler that the constructor is unknown at compile time. 
+
+The directive has the following properties:
+
+1. It can only be used with `<lwc.dynamic>` tag.
+    - An error will be issued when `lwc:is` is used on any other element.
+2. At compile time the value supplied to `lwc:is` can only be an expression.
+3. At runtime the expression supplied to `lwc:is` must be a `LightningElementConstructor`.
+    - An error will be thrown if anything other than a `LightningElementConstructor` is passed to `lwc:is`.
```
pmdartus, line 174:
```suggestion
    - An error will be thrown at runtime if anything other than a `LightningElementConstructor` is passed to `lwc:is`.
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+#### Instantiation of the custom element
+
+Dynamic component instantiation follows the same creation path as any other LWC custom element. When the dynamic component is ready to be rendered, the element will be created and mounted to the DOM.  If there is an existing element supplied to `lwc:is`, it will be unmounted as well.
```
pmdartus, line 178:
```suggestion
Dynamic component instantiation follows the same creation path as any other LWC custom element. When the dynamic component is ready to be rendered, the element will be created and mounted to the DOM.  If the component constructor supplied to `lwc:is` changes, the existing element will be unmounted.
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+    - When the value provided to `lwc:is` is not a `LightningElementConstructor` then an error is thrown indicating the constructor value must be `LightningElementConstructor`.
+- **The constructor is defined and is a `LightningElementConstructor`**
+    - When the constructor is a `LightningElementConstructor`, the component will be rendered in place of `<lwc.dynamic>`. The tag name used for the dynamic component is the value of `lightningElementName` on `LightningElementConstructor`.
+
+_Note in the case of lazy loading, the component author is responsible for resolving the promise.  The only value that can be given to `lwc:is` is a `LightningElementConstructor`._
+
+#### Selecting the dynamic component
+
+It is important to note that `<lwc.dynamic>` will not be rendered to the DOM.  As such, it is not available to be selected using `this.template.querySelector`.  The actual custom element must be selected once it has been rendered to the DOM.  The dynamic component can be selected either through the `lightningElementName` property on the constructor or through another attribute assigned to the component.
```
pmdartus, line 193:
```suggestion
It is important to note that the `<lwc.dynamic>` element itself will not be rendered to the DOM.  As such, it is not available to be selected using `this.template.querySelector`.  The actual custom element must be selected once it has been rendered to the DOM.  The dynamic component can be selected either through the `lightningElementName` property on the constructor or through another attribute assigned to the component.
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+Component authors can use the `connectedCallback` of the dynamic component to signal when the component has been rendered to the DOM.
```
pmdartus, line 195: Here are a couple of other options to detect whenever a dynamic component is mounted:
- Once a dynamic component constructor is set, it is guaranteed that will be mounted in the next rendering cycle. The parent component can wait until `renderedCallback` in the parent component after the component constructor is set to detect when it gets mounted.
- Use a `MutationObserver`.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+  connectedCallback() {
+    this.loadCtor();
+  }
+
+  async loadCtor() {
+    const { default: ctor } = await import("c/customConstructor");
+    this.customCtor = ctor;
+    this.template.querySelector(ctor.lightningElementName).setAttribute('title', 'hawaii');
+    this.template.querySelector(ctor.lightningElementName).addEventListener('click', () => console.log('hello world'));
```
pmdartus, line 227: This will fail, the component will only be mounted in the DOM in the next rendering cycle.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+`lwc:external` will not be available because the constructor provided to the `lwc:is` directive must be a `LightningElementConstructor`.
+
+#### Children of the dynamically created element
+
+The `<lwc.dynamic>` tag accepts child elements, first creating the dynamic component and subsequently the children afterwards.  Each time the dynamic component changes, the existing dynamic component will be unmounted along with all of its children.  The new dynamic component will be created along with the children.
+
+```html
+<template>
+    <lwc.dynamic lwc:is={ctor}>
```
pmdartus, line 243: Side note: Let's make sure to add tests for the different kinds of slots: shadow DOM, light DOM, scoped slot, slotting the `<lwc.dynamic>` element itself in the default slot and in a named slot, etc.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+_This is similar to the [Angular approach to dynamic components](https://angular.io/guide/dynamic-component-loader)._
+
+#### Comparison with `<lwc.dynamic lwc:is={expression}>`
+
+With `LightningDynamicElement` the framework keeps track of references to the dynamic components and also manages its lifecycle.  In contrast with `<lwc.dynamic>` the component author is responsible for keeping track of the custom element.  This means that any dynamic component must be retrieved each time the component changes.  Additionally, any imperatively assigned attributes or event listeners will need to be reapplied once the component changes as well.
+
+## Adoption strategy
+
+This is a breaking change for existing user of `lwc:dynamic` as it supersedes the directive.  However, `lwc:dynamic` directive has not reached GA and all consumers of the directive are internal.  They will need to switch their components to change the tag names to `lwc.dynamic` and the directives from `lwc:dynamic` to `lwc:is`.
```
pmdartus, line 322: Let's call out that we will support both `lwc:dynamic` and `lwc:is` directives at the same time to help with the transition. We will also need to add a compilation warning when the compiler detects the usage of the `lwc:dynamic` directive.

pmdartus (commented):
Great RFC!

text/0119-dynamic-component-creation.md
jmsjtu: Thanks for calling this out!  I was wondering if there may be some cases when the folder structure mapping would be off. 

I'll make a note to monitor this and do some research to see what the current state of affairs is.

text/0119-dynamic-component-creation.md
jmsjtu: Yes, I intend to set the tag name through `registerComponent`.  That was what I meant by the LWC engine setting the value, but I will make it more clear in this section what will happen.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+
+```javascript
+const { default: ctor } = await import("c/customConstructor");
+// The tag name can be directly accessed on the constructor
+ctor.lightningElementName
+```
+
+The value of `lightningElementName` should only be set by the LWC engine to prevent component authors from modifying it and producing unintended side effects.
+
+Alternatively, the custom element name can be stored in a `weakMap` and a public API can be exposed to retrieve the constructor's custom element name. However, because the custom element name can be used as a selector with `this.template.querySelector` placing it on the constructor seems more ergonomic.
```
nolanlawson, line 149: I would strongly suggest we go with the `WeakMap` approach, so that it's not exposed to component authors. `this.template.querySelector` should be unnecessary; component authors can use template refs instead. (Assuming we make template refs work with this proposal. 🙂)

text/0119-dynamic-component-creation.md
nolanlawson: I also find it interesting that GitHub's syntax highlighting breaks when using the `.`:

```html
<lwc.dynamic></lwc.dynamic>
```

Purely for readability, I agree with @pmdartus about using a `:` character:

```html
<lwc:dynamic></lwc:dynamic>
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+    - An error will be thrown if anything other than a `LightningElementConstructor` is passed to `lwc:is`.
+
+#### Instantiation of the custom element
+
+Dynamic component instantiation follows the same creation path as any other LWC custom element. When the dynamic component is ready to be rendered, the element will be created and mounted to the DOM.  If there is an existing element supplied to `lwc:is`, it will be unmounted as well.
+
+This means that when the value provided to `lwc:is` changes the component and all of its children will be destroyed and recreated using the new constructor.  In addition, when the value provided to `lwc:is` is evaluated, the LWC engine will render depending on the following scenarios:
+
+- **The constructor is `undefined` or `null`**
+    - When the constructor is undefined the `<lwc.dynamic>` tag along with all of its children are not rendered to the DOM.
```
nolanlawson, line 183: I would align with Svelte and say "when it is falsy." This handles cases like `valueThatMayBeFalse && MyConstructor`.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+- **The constructor is defined and is not a `LightningElementConstructor`**
+    - When the value provided to `lwc:is` is not a `LightningElementConstructor` then an error is thrown indicating the constructor value must be `LightningElementConstructor`.
+- **The constructor is defined and is a `LightningElementConstructor`**
+    - When the constructor is a `LightningElementConstructor`, the component will be rendered in place of `<lwc.dynamic>`. The tag name used for the dynamic component is the value of `lightningElementName` on `LightningElementConstructor`.
+
+_Note in the case of lazy loading, the component author is responsible for resolving the promise.  The only value that can be given to `lwc:is` is a `LightningElementConstructor`._
```
nolanlawson, line 189: Side comment: undoubtedly some folks will want to render a placeholder while the promise is resolving, but this seems to be trivial with this proposal because you can just set the constructor to the placeholder constructor while waiting for the promise to resolve.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+    this.template.querySelector(ctor.lightningElementName).setAttribute('title', 'hawaii');
+    this.template.querySelector(ctor.lightningElementName).addEventListener('click', () => console.log('hello world'));
+  }
+}
+```
+#### Usage with other LWC directives
+
+All [LWC custom element directives](https://lwc.dev/guide/reference#directives-for-nested-templates) with the exception of `lwc:external` are available to use on `<lwc.dynamic>` and their functionality will be passed through to the custom element once it has been created.
+
+`lwc:external` will not be available because the constructor provided to the `lwc:is` directive must be a `LightningElementConstructor`.
```
nolanlawson, line 235: We should add support for `lwc:ref`, as mentioned above. 🙂

text/0119-dynamic-component-creation.md
nolanlawson: Bonus points if it behaves the same across synthetic and native! Based on my reading, it should. The slot contents should be unmounted and remounted in both cases.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+Is semantically the same as
+
+```html
+<template>
+    <!-- The ctor swaps between light and shadow DOM components -->
+    <lwc.dynamic lwc:is={ctor}></lwc.dynamic>
+</template>
+```
+
+In both cases the dynamic component will be fully mounted and unmounted.
```
nolanlawson, line 274: This same challenge applies to synthetic vs native shadow components, since they have different slot mechanisms.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,335 @@
...
+    const { default: ctor } = await import("c/customConstructor");
+    this.customCtor = ctor;
+    this.template.querySelector(ctor.lightningElementName).setAttribute('title', 'hawaii');
+    this.template.querySelector(ctor.lightningElementName).addEventListener('click', () => console.log('hello world'));
+  }
+}
+```
+#### Usage with other LWC directives
+
+All [LWC custom element directives](https://lwc.dev/guide/reference#directives-for-nested-templates) with the exception of `lwc:external` are available to use on `<lwc.dynamic>` and their functionality will be passed through to the custom element once it has been created.
```
nolanlawson, line 233: I'm a bit unclear what directives are or are not included here. For instance, is `lwc:spread` included?

nolanlawson (commented):
Looks amazing to me so far, just a few comments and suggestions.

text/0119-dynamic-component-creation.md
jmsjtu: I agree, if they need a placeholder they can implement one and render it while the promise is resolving.  I thought about making `<lwc:dynamic>` a custom element to handle this scenario but it felt unnecessary.

text/0119-dynamic-component-creation.md
jmsjtu: I don't see a reference to `lwc:ref` in the lwc.dev site and I wanted to reference the guide to ensure that I covered all the directives.  

I'll check to see if there's another source with a comprehensive list, otherwise, I can list them out here.

text/0119-dynamic-component-creation.md
jmsjtu: I added some more details on how the `lightningElementName` is set [here](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR146)

text/0119-dynamic-component-creation.md
jmsjtu: Sounds good, I've updated to use `lwc:dynamic`.

text/0119-dynamic-component-creation.md
jmsjtu: Thanks for these, added them to the [RFC](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR193-R194)!

text/0119-dynamic-component-creation.md
jmsjtu: Added a [section](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR325) to indicate we will support both directives.

text/0119-dynamic-component-creation.md
```diff
@@ -309,27 +308,31 @@ Full details can be found [here](https://github.com/salesforce/lwc/pull/3204)
...
 Examples of [documentation](https://svelte.dev/docs#template-syntax-svelte-component) and [playground](https://svelte.dev/tutorial/svelte-component) from Svelte.
 
 # Unresolved questions
 
 Will there be issues with collisions?  For example if a dynamic constructor is imported but has the same custom element name as a custom element that's already on the page?
-    - Does the `upgradeableCallback` take care of this?
\ No newline at end of file
+    - Does the `upgradeableCallback` take care of this?
+
+Should we log errors when component authors try to use `<lwc:*>` incorrectly?  For example should we prevent component authors from doing something like `document.createElement('lwc:other');`?
```
pmdartus, line 338: I would recommend staying away from global patching. This opens the door for all sorts of bugs and it is hard to maintain in the long run (our old friend synthetic shadow DOM). 
I would recommend explicitly mentioning in the documentation that `<lwc:dynamic>` is an LWC compiler signal, and doesn't have any bearing at runtime.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,338 @@
...
+
+```javascript
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    lazyConstructor;
+
+    async connectedCallback() {
+        const { default: ctor } = await import('lightning/concreteComponent');
+        this.lazyConstructor = ctor;
```
pmdartus, line 37: This dynamic import and `async` / `await` could confuse the reader (one of my coworkers in France wasn't clear about this). I would recommend either stripping this from this example or adding another example where the constructor is imported sync just to nail the point.

text/0119-dynamic-component-creation.md
nolanlawson: I'll add it to the docs. It should be there.

jmsjtu:
@pmdartus @nolanlawson I'm thinking about changing the tag name from `<lwc:dynamic>` to `<lwc:component>`.

Looking through the [Svelte](https://svelte.dev/docs#template-syntax-svelte-component) and [Vue](https://vuejs.org/guide/essentials/component-basics.html#dynamic-components) implementations, they both use `component` in their tag names.

Should we change the tag name to align with what the other frameworks are doing?  It may help component authors understand this proposal and build their mental model.

Would love to get your thoughts on this.

nolanlawson:
@jmsjtu I like it, yeah! It's a component, not "a dynamic." 😄

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,338 @@
...
+Dynamic components refer to custom elements where the constructor is not known at compile time.
+
+*This proposal only focuses on the dynamic component creation process, lazy loading which was included in the [original proposal for dynamic components](https://rfcs.lwc.dev/rfcs/lwc/0110-dynamic-components) is not included.*
+
+## Basic example
+
+```html
+<template>
+    <div class="container">
+        <lwc:dynamic lwc:is={lazyConstructor}></lwc:dynamic>
```
caridy, line 24: `lwc:is` was defined because at the time we thought the `is` property in web components was going to be implemented everywhere, at this point, we just don't know. The question is: should we use a more descriptive directive? `<lwc:dynamic component:is={}>` or should we stick to it? I think we already agree that all directives should be prefixed with `lwc:` so another option is: `<lwc:dynamic lwc:component={}>` which probably reads a little better.

text/0119-dynamic-component-creation.md
caridy: agreed! also, if their assign a promise to the `lazyConstructor` by mistake, it should throw, It must be a qualifying constructor or undefined, make sure you mentioned that.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,338 @@
...
+- [Vue: `<component :is="...">`](https://vuejs.org/guide/essentials/component-basics.html#dynamic-components)
+- [Angular: viewContainerRef](https://angular.io/guide/dynamic-component-loader)
+- [Lit: static expressions](https://lit.dev/docs/templates/expressions/#static-expressions)
+- [Aura: $A.createComponent](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/js_cb_dynamic_cmp_async.htm)
+
+Frameworks that use JSX such as [React](https://reactjs.org/docs/rendering-elements.html#rendering-an-element-into-the-dom) and [Stencil](https://stenciljs.com/docs/templating-jsx) can dynamically render components directly into the markup using expressions.
+
+The goal of managing dynamic components in LWC is to follow the same component lifecycle as all other LWC custom elements.  Having said that, taking a similar approach as Svelte and Vue seems the most natural and intuitive path forward.
+
+## Detailed design
```
caridy, line 113: Add a note to mark this type of component as "risky"

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,338 @@
...
+
+A new property, `lightningElementName` on `LightningElementConstructor` is introduced to store the name of the custom element.
```
caridy, line 115: In our call, we agreed that this must not be observable to user-land... and registerComponent is a good way to achieve that.

text/0119-dynamic-component-creation.md
caridy: it will definitely work with this proposal... make sure we have an example.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,338 @@
...
+
+In both cases the dynamic component will be fully mounted and unmounted.
+
+#### Styles
+
+Something to keep in mind is that [light DOM styles are injected to the closest root node](https://rfcs.lwc.dev/rfcs/lwc/0115-light-dom#styles) and are not removed once the components are unmounted. This means that repeated mounting and mounting of a dynamic component could cause styles to be overwritten.  This can be avoided by using either scoped styles or shadow dom, as the styles will be scoped.
+
+## Drawbacks
+
+The main drawback of this design is that when the constructor changes, the entire component hierarchy will need to be re-rendered. This means that the children of the custom element will also need to re-rendered.  Careful consideration should be taken by the component author as this may cause performance issues.
```
caridy, line 281: This is only true if synthetic shadow isn't it? Let's chat more about this.

text/0119-dynamic-component-creation.md
nolanlawson: Bikeshedding, but I'm partial to

```html
<lwc:component lwc:this={}>
```

or

```html
<lwc:component lwc:constructor={}>
```

Agreed that `lwc:is` may give the wrong impression. Even though I do like how it reads ("component is ___"):

```html
<lwc:component lwc:is={}>
```

pmdartus: I don't have a strong opinion on the name. I like the `<lwc:component lwc:is={}>` combo since it maps to a concept that has been speced out even if all the browsers do not implement it. On the other hand, the `lwc:this` isn't very intuitive in my opinion.

nolanlawson: In that case, I would vote for `lwc:is` despite the potential confusion. At least it is prefixed by `lwc:`, so it's clearly not the same as the browser one.

jmsjtu: My vote would go to `<lwc:component lwc:is={}>` as well, I think it conveys the meaning the best out of the other options.

Vue also does something similar with the [`:is`](https://vuejs.org/guide/essentials/component-basics.html#dynamic-components) attribute.

text/0119-dynamic-component-creation.md
jmsjtu: @pmdartus just to clarify, if the folder structure does not respect the standard `<namespace>/<name>/<name>.js` format the only way it can be recognized is if it has an [alias module record](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/module-resolver#alias-module-record) correct?

pmdartus: > @pmdartus just to clarify, if the folder structure does not respect the standard `<namespace>/<name>/<name>.js` format the only way it can be recognized is if it has an [alias module record](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/module-resolver#alias-module-record) correct?

Correct.

jmsjtu:
@pmdartus @nolanlawson @caridy - I've updated the RFC based on feedback provided so far, here's a quick summary of the changes that have been added:

1. Updated the examples to remove the `async`/`await` syntax.
2. Updated the _Storing and retrieving a custom element's name_ section to use a `Map` instead of a property on `LightningElementConstructor`.
3. Updated the relevant sections to use `lwc:ref` as the preferred way to reference the dynamic component.
4. Changed the tag name and directive to `<lwc:component lwc:is={}>`

Things still left to do:
1. Mark the components using `<lwc:dynamic lwc:is={}>` as risky.
2. Update the _Usage with other LWC directive_s section to be more descriptive.

text/0119-dynamic-component-creation.md
jmsjtu: Fixed the [example](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR286-R301) with the latest commit.

text/0119-dynamic-component-creation.md
jmsjtu: Updated to use a `Map` since that's what's [currently being used for templates](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/component.ts#L18).

text/0119-dynamic-component-creation.md
jmsjtu: Added an [example](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR233-R258) using `lwc:ref`

text/0119-dynamic-component-creation.md
jmsjtu: Added a [note](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR412) to address this.

text/0119-dynamic-component-creation.md
jmsjtu: All examples have been updated and there is a note about using promises [here](https://github.com/salesforce/lwc-rfcs/pull/71/files#diff-a1de05148a61ea9b00f70bd4ff5feb3895c16dd99707b173ba57c081b859250bR218).

text/0119-dynamic-component-creation.md
jmsjtu: `<lwc:component lwc:is={}>` is the winner!

text/0119-dynamic-component-creation.md
jmsjtu: I fleshed out the _Storing and retrieving a custom element's name_ section to call this out.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,412 @@
...
+- [Lit: static expressions](https://lit.dev/docs/templates/expressions/#static-expressions)
+- [Aura: $A.createComponent](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/js_cb_dynamic_cmp_async.htm)
+
+Frameworks that use JSX such as [React](https://reactjs.org/docs/rendering-elements.html#rendering-an-element-into-the-dom) and [Stencil](https://stenciljs.com/docs/templating-jsx) can dynamically render components directly into the markup using expressions.
+
+The goal of managing dynamic components in LWC is to follow the same component lifecycle as all other LWC custom elements.  Having said that, taking a similar approach as Svelte and Vue seems the most natural and intuitive path forward.
+
+## Detailed design
+
+A new internal `Map` called `componentRegisteredNameMap` is introduced to store the custom element name and an internal API `getComponentRegisteredName` is introduced to retrieve the custom element name.
```
jmsjtu, line 122: I decided to switch to storing the custom element name on a `Map` because we decided that the [custom element's name should not be observable to the component authors](https://github.com/salesforce/lwc-rfcs/pull/71#discussion_r1062982660).

The main appeal to me of storing the name on the `LightningElementConstructor` was to be able to use it with `this.template.querySelector` but since we can use `lwc:ref` instead, it doesn't seem necessary anymore.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,418 @@
...
+    loadCtor() {
+        if (this.customCtor) {
+            // Append to the end of the array
+            this.ctors.push(this.customCtor);
+        }
+
+        // Remove the first element in the array
+        this.customCtor = this.ctors.shift();
+    }
+}
```
pmdartus, line 92: An attempt to make this example more readable.

```suggestion
import { LightningElement } from "lwc";
import About from 'c/about';
import Home from 'c/home';

const ctors = [About, Home]; 

export default class extends LightningElement {
    customCtor;
    index = 0;

    loadCtor() {
        const nextCtor = ctors[this.index++ % ctors.length];
        this.customCtor = nextCtor;
    }
}
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,418 @@
...
+```
+
+Since there is no way to know which alias to use in this case, the custom element name will be the first alias that is resolved.  This should be an edge case that does not appear often.
+
+#### Defining an anchor in the DOM
+
+##### `<lwc:component>`
+
+`<lwc:component>` is a placeholder tag for the dynamic component and will not be rendered to the DOM.  It serves as a special signal to the compiler that the dynamic component will be rendered at the specific location in the DOM. (Similar to how `template` is not rendered)
```
pmdartus, line 193:
```suggestion
`<lwc:component>` is a placeholder tag for the dynamic component and will not be rendered to the DOM.  It serves as a special signal to the compiler that the dynamic component will be rendered at the specific location in the DOM. (Similar to how `<template>` is not rendered)
```

pmdartus (approved):

text/0119-dynamic-component-creation.md
ravijayaramappa: @caridy is the component being marked as "risky" to provide renderer hooks in compiled template?
I ask because we decided not to provide renderer hooks for custom elements because of how we cascade the renderer from owner to its children.

jmsjtu: @caridy it looks like custom renderer hooks currently [do not work with custom elements](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/shared/renderer-hooks.ts#L48-L54).  

Should I mention in this RFC that once custom renderer hooks work with custom elements they will be marked as "risky" or skip that for now?

caridy: ah, right! you're right!

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+    - An error will be thrown at runtime if anything other than a `LightningElementConstructor` is passed to `lwc:is`.
+
+#### Instantiation of the custom element
+
+Dynamic component instantiation follows the same creation path as any other LWC custom element. When the dynamic component is ready to be rendered, the element will be created and mounted to the DOM.  If the component constructor supplied to `lwc:is` changes, the existing element will be unmounted.
+
+This means that when the value provided to `lwc:is` changes the component and all of its children will be destroyed and recreated using the new constructor.  In addition, when the value provided to `lwc:is` is evaluated, the LWC engine will render depending on the following scenarios:
+
+- **The constructor is falsy**
+    - When the constructor is falsy the `<lwc:component>` tag along with all of its children are not rendered to the DOM.
```
caridy, line 215: probably just add the actual condition here, `ctor == undefined` I suspect.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+- **The constructor is defined and is not a `LightningElementConstructor`**
+    - When the value provided to `lwc:is` is not a `LightningElementConstructor` then an error is thrown indicating the constructor value must be `LightningElementConstructor`.
```
caridy, line 217:
```suggestion
    - When the value provided to `lwc:is` is not a valid `LightningElement` constructor then an error is thrown.
```

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+1. The `<lwc:component>` tag will not be rendered to the DOM which means it is not available to be selected using `this.template.querySelector`. 
+2. The LWC engine does not provide a way to retrieve the dynamic component's custom element name for use as a selector with `this.template.querySelector` (we recommend using `lwc:ref` instead).
+
+To select the dynamic component, the actual custom element must be selected once it has been rendered to the DOM by using the [`lwc:ref`](https://rfcs.lwc.dev/rfcs/lwc/0000-refs) directive or through another attribute assigned to the component.
+
+Here are some ways component authors can detect when a dynamic component has mounted:
+- Use `connectedCallback` on the dynamic component to signal when it has mounted.
+- A dynamic component constructor is guaranteed to be mounted in the next rendering cycle once it has been set.  When it is set, the parent component can wait until the `renderedCallback` lifecycle method is invoked to detect when the dynamic component is mounted.
+- Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect when the dynamic component has mounted.
```
caridy, line 234: this is tricky, because you have no way to get the name of the component from the constructor that you are setting. I think the options are, use ref, or use a wrapping element that you can query and check for firstChild on that.

caridy: for the mutation observe to be useful, you need a way to identify the element, which in most cases you don't.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+- **The constructor is defined and is not a `LightningElementConstructor`**
+    - When the value provided to `lwc:is` is not a `LightningElementConstructor` then an error is thrown indicating the constructor value must be `LightningElementConstructor`.
+- **The constructor is defined and is a `LightningElementConstructor`**
+    - When the constructor is a `LightningElementConstructor`, the component will be rendered in place of `<lwc:component>`. The tag name used for the dynamic component is the value returned from `getComponentRegisteredName`.
+
+_Note in the case of lazy loading, the component author is responsible for resolving the promise.  The only value that can be given to `lwc:is` is a `LightningElementConstructor`._
+
+#### Selecting the dynamic component
+
+Dynamic components have certain properties that will not work with `this.template.querySelector`:
```
caridy, line 225: this paragraph and list below are confusing. The idea is probably to explain that you don't have a name to query for... so either you add a class to the `<lwc:component>`, so you can query with that, or you use a ref, which is the recommended way. It is not really about `<lwc:component>` since they will never see that in the dom, so no need to mention it.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+            // this.refs.foo will contain a reference to the DOM node
+            console.log(this.refs.foo);
+        }
+    }
+}
+```
+
+#### Assigning attributes and event listeners
+
+Attributes and event listeners can be passed to the `<lwc:component>` tag declaratively or can be assigned directly to the custom element imperatively.
```
caridy, line 266: I think the whole section is kinda not needed, just mentioned that `<lwc:component>` does accept all , or almost all APIs (I believe it supports all). And it obey the roles of those apis, including attributes, events, etc.

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+    <!-- The ctor swaps between light and shadow DOM components -->
+    <lwc:component lwc:is={ctor}></lwc:component>
+</template>
+```
+
+In both cases the dynamic component will be fully mounted and unmounted.
+
+#### Styles
+
+Something to keep in mind is that [light DOM styles are injected to the closest root node](https://rfcs.lwc.dev/rfcs/lwc/0115-light-dom#styles) and are not removed once the components are unmounted. This means that repeated mounting and mounting of a dynamic component could cause styles to be overwritten.  This can be avoided by using either scoped styles or shadow dom, as the styles will be scoped.
```
caridy, line 353: this poses a new question not necessary related to this RFC, what happen if I have two instances of the same light component on the same template, and one of them is removed?

text/0119-dynamic-component-creation.md
caridy: ping

text/0119-dynamic-component-creation.md
```diff
@@ -0,0 +1,415 @@
...
+
+Examples of [documentation](https://svelte.dev/docs#template-syntax-svelte-component) and [playground](https://svelte.dev/tutorial/svelte-component) from Svelte.
+
+# Unresolved questions
+
+Will there be issues with collisions?  For example if a dynamic constructor is imported but has the same custom element name as a custom element that's already on the page?
+- Pivots in Locker and the `UpgradeableConstructor` should take care of this.
+
+Should we log errors when component authors try to use `<lwc:*>` incorrectly?  For example should we prevent component authors from doing something like `document.createElement('lwc:other');`?
+- In order to do this we would need to globally patch certain APIs which is something that we want to avoid to prevent bugs that are hard to maintain in the long run.  To handle this case, explicitly mention in the documentation that `<lwc:dynamic>` is an LWC compiler signal, and doesn't have any bearing at runtime.
```
caridy, line 415: I'm fine with this... the idea really is to say that doing something like `document.createElement('lwc:other')` give you no powers.

caridy (approved):
LGTM, just a bunch of little suggestions.

text/0119-dynamic-component-creation.md
jmsjtu: I agree, I think this could be a bit tricky.  I think as long as you can get a reference to the DOM element you can identify the component through the `tagName` or `nodeName` properties.

I think it's possible to identify the element using the mutation observer, you could have a wrapping element around the dynamic element and listen for mutations in the children.

Once you have a reference to the dynamic element's DOM node you should be able to get the component's name.

Here's a quick [POC](https://stackblitz.com/edit/salesforce-lwc-lrlwos?file=src/modules/x/app/app.html) using the existing `lwc:dynamic`, let me know if this is what you meant.

I do think though that using `renderedCallback` with `lwc:ref` is more straight forward.

text/0119-dynamic-component-creation.md
jmsjtu: That's what I was going for, I'll reword this portion and make it more clear.

text/0119-dynamic-component-creation.md
jmsjtu: I think in that case the styles will be inserted at the top of the root node.  If either or both of the elements are removed from the template, the styles will still be at the root node.

jmsjtu: I also noticed this reads incorrectly, I mean to say that repeated mounting and unmounting of different dynamic components could cause styles to be overwritten.

I'll update this section too.

text/0119-dynamic-component-creation.md
pmdartus: > I do think though that using `renderedCallback` with `lwc:ref` is more straight forward.

Let's get rid of the note about `MutationOberserver`.

text/0119-dynamic-component-creation.md
jmsjtu: I think when the dynamic component has a `<slot>` then there shouldn't be a difference between synthetic and native shadow.  The dynamic component will be unmounted along with its children and recreated using the new constructor.  Here's an [example](https://stackblitz.com/edit/salesforce-lwc-bsehiw?file=src/modules/x/child2/child2.js) using the current `lwc:dynamic` implementation.

If the dynamic component doesn't have a `<slot>` though there will be a difference as I believe in synthetic shadow the children will not be rendered but in native shadow they will be rendered.

This means that in synthetic shadow there are no children to mount/unmount while in native there will be.  

I'll do some more testing and let you know what I find.

text/0119-dynamic-component-creation.md
jmsjtu: This got me thinking, are there any cases when you would want to identify the component before it has been inserted into the DOM (when you only have the constructor)?

pmdartus: > This got me thinking, are there any cases when you would want to identify the component before it has been inserted into the DOM (when you only have the constructor)?

I don't see how you can get a handle on an LWC component before it gets inserted into the DOM. 

The LWC engine executes those steps synchronously when creating a component: create the component host element, invoke its `constructor`, insert it into the DOM, and invoke `connectedCallback`. There is no way for the parent component to inject some logic in the middle of the creation process.

text/0119-dynamic-component-creation.md
jmsjtu: I did some testing, here's an [example](https://stackblitz.com/edit/salesforce-lwc-bsehiw?file=src/modules/x/app/app.html) comparing both native and synthetic shadow when there is no `<slot>`.

The synthetic shadow does not mount any children while the native shadow does mount the children.

text/0119-dynamic-component-creation.md
jmsjtu: @nolanlawson updated this section, let me know if it's clearer now which directives are supported.

text/0119-dynamic-component-creation.md
caridy: just add an example with `class="foo"` or some `data-` attribute, then use that with `this.template.querySelector`

text/0119-dynamic-component-creation.md
caridy: this is probably a question for @nolanlawson , how can we track down the multiple instances that contribute to a particular css style to be added?

nolanlawson: This section is incorrect:

> This means that repeated mounting and mounting of a dynamic component could cause styles to be overwritten

Styles are never overwritten in LWC. We simply append styles as needed. Two instances of the same light DOM component will result in one stylesheet being added to the nearest shadow root. This stylesheet is never removed (unless the shadow root itself is removed).

> how can we track down the multiple instances that contribute to a particular css style to be added?

@caridy Currently there is no debug information to figure this out; you have to just know which stylesheet came from which component.

PR #72, opened 2023-01-19
RFC: Declarative root element

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/declarative-root/text/0000-declarative-root.md)

### DIFF
--- a/text/0000-declarative-root.md
+++ b/text/0000-declarative-root.md
@@ -1,0 +1,276 @@
+---
+title: Declarative root element
+status: DRAFTED
+created_at: 2022-01-19
+updated_at: 2022-01-19
+champion: Nolan Lawson (nolanlawson)
+pr: https://github.com/salesforce/lwc-rfcs/pull/72
+---
+
+# Declarative root element
+
+## Summary
+
+Provides a declarative way to apply reactive attributes, classes, and event listeners to the root (host) element.
+
+## Basic example
+
+In a component's template HTML file:
+
+```html
+<template>
+    <lwc:root
+      class="static-class"
+      data-foo={dynamicAttribute}
+      onclick={onRootClick}
+      lwc:spread={otherAttributes}
+    ></lwc:root>
+    <h1>Hello world</h1>
+</template>
+```
+
+Result:
+
+```html
+<x-component class="static-class" data-foo="foo" data-other="other">
+    #shadow-root
+      <h1>Hello world</h1>
+</x-component>
+```
+
+Classes, attributes, and event listeners are applied from the `<lwc:root>` element to the
+root `<x-component>` element. (The event listener is not shown above.)
+
+## Motivation
+
+A common pattern in LWC components is something like this:
+
+```js
+export default class extends LightningElement {
+  connectedCallback() {
+    this.template.host.classList.add('my-class');
+  }
+}
+```
+
+Developers may want to add a class, set an attribute, or add an event listener to the root (host) element
+of the component. Today there is no declarative way to do this, so they resort to doing it manually in the
+`connectedCallback`, `constructor`, or `renderedCallback`.
+
+This has several downsides:
+
+1. It does not work well with SSR, where it would require shims for DOM APIs like `classList`, `setAttribute`, etc., where `connectedCallback` timings [may differ](https://github.com/salesforce/lwc/issues/3009) between SSR and DOM, and where `renderedCallback` doesn't fire at all.
+2. It is not easy to make it reactive, e.g. to set a class that updates based on a `@track`ed property.
+
+Prior art in other frameworks:
+
+- Stencil: [`<Host>` functional component](https://stenciljs.com/docs/host-element)
+- Angular: [`@HostBinding`](https://angular.io/api/core/HostBinding)
+- FAST: [Host directive template](https://www.fast.design/docs/fast-element/using-directives/#host-directives)
+- Lit: [Nonexistent, but considered](https://github.com/lit/lit/issues/1825)
+
+## Detailed design
+
+The `<lwc:root>` element is a synthetic element defined in a component's template HTML file. Similar to [dynamic components](https://github.com/salesforce/lwc-rfcs/pull/71), it does not actually render in the DOM.
+
+```html
+<template>
+    <lwc:root class="foo"></lwc:root> <!-- Doesn't actually render -->
+</template>
+```
+
+However, many attributes and directives that can be applied to a normal `HTMLElement` can be applied to `<lwc:root>`. These include:
+
+- Standard [global HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes), such as `class` and `tabindex`
+- Custom HTML attributes, such as `data-*`
+- Event listeners, such as `onclick` or `onfocus`
+- The `lwc:spread` directive
+
+Other directives that apply to normal `HTMLElement`s, such as `lwc:inner-html`, `lwc:ref`, and `key`, are not supported.
+
+A bare `<lwc:root></lwc:root>` with no attributes/listeners/directives is allowed, but effectively does nothing.
+
+### Restrictions
+
+#### Placement
+
+A `<lwc:root>` element must be placed at the top level (root) of a `<template>`, and may not be preceded by other elements:
+
+```html
+<!-- Valid -->
+<template>
+    <lwc:root></lwc:root>
+    <div></div>
+</template>
+```
+
+```html
+<!-- Invalid -->
+<template>
+    <div>
+        <lwc:root></lwc:root>
+    </div>
+</template>
+```
+
+```html
+<!-- Invalid -->
+<template>
+    <div></div>
+    <lwc:root></lwc:root>
+</template>
+```
+
+Comments are allowed to precede `<lwc:root>`:
+
+```html
+<!-- Valid -->
+<template>
+    <!-- Comment -->
+    <lwc:root></lwc:root>
+</template>
+```
+
+However, in the case of `lwc:preserve-comments`, comments may not precede `<lwc:root>`:
+
+```html
+<!-- Invalid -->
+<template lwc:preserve-comments>
+    <!-- Comment -->
+    <lwc:root></lwc:root>
+</template>
+```
+
+#### Contents
+
+The `<lwc:root>` element may not have any contents other than whitespace and comments:
+
+```html
+<!-- Valid -->
+<template>
+    <lwc:root>
+        
+    </lwc:root>
+</template>
+```
+
+```html
+<!-- Valid -->
+<template>
+    <lwc:root>
+        <!-- Comment -->
+    </lwc:root>
+</template>
+```
+
+However, in the case of `lwc:preserve-comments`, comments inside of `<lwc:root>` are not allowed:
+
+```html
+<!-- Invalid -->
+<template lwc:preserve-comments>
+    <lwc:root>
+        <!-- Comment -->
+    </lwc:root>
+</template>
+```
+
+This restriction is why `lwc:inner-html` is not supported on `<lwc:root>`.
+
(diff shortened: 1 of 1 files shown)


text/0000-declarative-root.md
```diff
@@ -0,0 +1,276 @@
...
+
+The reason for this is to allow a component to set defaults for its root's behavior, while still allowing consumers to override those defaults.
+
+## Drawbacks
+
+Implementing this feature does require additional complexity, and moves us further away from standard custom element syntax.
+
+## Alternatives
+
+### Calling it "host" instead of "root"
```
nolanlawson, line 239: BTW I am a bit agnostic on "root" versus "host" versus some other name we can think of.  "Root" is also kind of confusing because it might make you think of [shadow root](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot).

text/0000-declarative-root.md
```diff
@@ -0,0 +1,276 @@
...
+### Placing attributes on the root `<template>` tag.
+
+Historically in LWC, `<template>` is used to designate a reusable tree of HTML, e.g. in `for:each` and [scoped slots](https://github.com/salesforce/lwc-rfcs/pull/63). The root `<template>` does not actually refer to the root/host element.
+
+Also, because `<lwc:root>` supports attributes and directives that generally would apply to normal `HTMLElement`s, it makes
+sense to represent it as a pseudo-normal HTML element.
+
+### Supporting `lwc:ref`
+
+Supporting `lwc:ref` is technically feasible, but doesn't make much sense, as there are already alternatives. In shadow DOM, developers can use `this.template.host`, and in light DOM, they can use `this`.
```
nolanlawson, line 255: Technically `this` in light DOM is not _exactly_ the same as what you'd get with `lwc:ref`, since the former is the internal view of the component and the latter is the external view. But it's ~99% the same. You can do `this.classList`, `this.children`, `this.setAttribute`, etc., and it works as expected.

text/0000-declarative-root.md
```diff
@@ -0,0 +1,276 @@
...
+    <div class={clazz}></div>
+</template>
+```
+
+In other words, in the above example, if `clazz` changes, then the root element's `class` attribute should be changed
+in the same tick as when the `<div>`'s `class` changes.
+
+The above also applies to the timing of when event listeners are attached using `on*`.
+
+No guarantees are made about the ordering of changes made to `<lwc:root>` versus elements inside of the template.
```
jmsjtu, line 196: > ordering of changes made to `<lwc:root>`

Is this referring to the ordering of how classes, event handlers, etc. are applied from `<lwc:root>` to the host element?

> versus elements inside of the template

Does this mean that there could be a difference in the order the classes, event handlers, etc are applied on the host vs to elements within the host?

text/0000-declarative-root.md
```diff
@@ -0,0 +1,276 @@
...
+Prior art in other frameworks:
+
+- Stencil: [`<Host>` functional component](https://stenciljs.com/docs/host-element)
+- Angular: [`@HostBinding`](https://angular.io/api/core/HostBinding)
+- FAST: [Host directive template](https://www.fast.design/docs/fast-element/using-directives/#host-directives)
+- Lit: [Nonexistent, but considered](https://github.com/lit/lit/issues/1825)
+
+## Detailed design
+
+The `<lwc:root>` element is a synthetic element defined in a component's template HTML file. Similar to [dynamic components](https://github.com/salesforce/lwc-rfcs/pull/71), it does not actually render in the DOM.
```
jmsjtu, line 74: Will this element be available for light DOM as well?

text/0000-declarative-root.md
nolanlawson: No, I didn't think of that. Updated! bbbc288

text/0000-declarative-root.md
nolanlawson: Yes. Clarified: 815466f

text/0000-declarative-root.md
```diff
@@ -0,0 +1,280 @@
...
+              lwc:spread={others}>
+    </lwc:root>
+</template>
+```
+
+In cases of conflict, the following rules apply:
+
+- `lwc:spread` precedence applies [as normal](https://github.com/salesforce/lwc-rfcs/pull/52) within the `<lwc:root>` element itself.
+- For event listeners such as `onclick`, both event listeners are attached.
+- For the `class` attribute, strings are concatenated with a single whitespace (`" "`) character (e.g. `"foo quux"` in the above example). No attempt is made to deduplicate duplicate strings.
```
pmdartus, line 232: How do you envision reconciliating the `class` attribute if it is changed programmatically?

For example, what would be the generated `class` attribute value if the parent runs this in its `renderedCallback` and if the child updates the class on the `<lwc:root class={computed}>`?

```
this.template.querySelector('x-child').classList.add('prog')
```

IMO, it's reasonable to discard any programmatic change if the `class` property is set declaratively.

text/0000-declarative-root.md
pmdartus: > "Root" is also kind of confusing because it might make you think of [shadow root](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot).

It's the first thing that came to my mind. My 2 cents is that "host" makes more sense. It also aligns with the terminology used by other frameworks.

text/0000-declarative-root.md
```diff
@@ -0,0 +1,280 @@
...
+Also, because `<lwc:root>` supports attributes and directives that generally would apply to normal `HTMLElement`s, it makes
+sense to represent it as a pseudo-normal HTML element.
+
+### Supporting `lwc:ref`
+
+Supporting `lwc:ref` is technically feasible, but doesn't make much sense, as there are already alternatives. In shadow DOM, developers can use `this.template.host`, and in light DOM, they can use `this`.
+
+Also, the whole point of this feature is to avoid needing programmatic access to the root/host element. So adding `lwc:ref` would be counter-productive.
+
+## Adoption strategy
```
pmdartus, line 263: I would go a step further and deprecate all APIs exposed on the `LightningElement` that are covered by this proposal.

The main motivation would be to remove those programmatic APIs on the engine server. Removing programmatic host element mutation would certainly greatly simplify the SSR logic and would enable us to get rid of some of the DOM APIs polyfills we have to maintain for SSR.

text/0000-declarative-root.md
```diff
@@ -0,0 +1,280 @@
...
+
+This proposal can be adopted as a net-new feature and does not have backwards-compatibility implications. `lwc:root` is
+not a pre-existing built-in HTML element, and it's extremely unlikely that anyone is creating runtime elements with this name.
+
+It may be possible to write codemods that look for simple usages of e.g. `this.template.host.classList.add('foo')`
```
pmdartus, line 268: A linting rule is certainly welcome to raise awareness of this new capability.

text/0000-declarative-root.md
```diff
@@ -0,0 +1,280 @@
...
+## Alternatives
+
+### Calling it "host" instead of "root"
+
+Other frameworks call this feature "host." And there is some precedent in LWC, in that [light DOM scoped styles](https://github.com/salesforce/lwc-rfcs/pull/50) allow for styling the root of a light DOM component using `:host` in `*.scoped.css`.
+
+However, this proposal prefers "root" to "host," because "root" is a more generic term that does not imply shadow DOM. This
+makes it clear that, for example, a child light DOM component cannot set attributes on a parent shadow DOM component by using this technique. (Whereas a child light DOM component can indeed style its parent's root node using `:host` in unscoped CSS.)
+
+### Placing attributes on the root `<template>` tag.
```
abdulsattar, line 250: What about replacing the `template` tag with `<lwc:root>`?

e.g.
```html
<lwc:root class="klass">
  <h1>regular elements</h1>
</lwc:root>
```

Of course, the regular `<template>` will still be supported.

text/0000-declarative-root.md
```diff
@@ -0,0 +1,280 @@
...
+<template>
+    <lwc:root class="foo"></lwc:root> <!-- Doesn't actually render -->
+</template>
+```
+
+Instead, `<lwc:root>` refers to either the shadow host element (in the case of shadow DOM components) or the root element (in the case of light DOM components). (In other words, if you are writing a component whose JavaScript file is `x/foo.js`, it refers to `<x-foo>`.)
+
+Many attributes and directives that can be applied to a normal `HTMLElement` can also be applied to `<lwc:root>`. These include:
+
+- Standard [global HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes), such as `class` and `tabindex`
```
jhefferman-sfdc, line 86: This sounds very good, ones that are important to us include `aria-xyz`, `role`, `size`, `id` and also things like "`variant`" on our primitive-icon (looking into that one)

text/0000-declarative-root.md
nolanlawson: Hmmm, OK. I am fine with changing it to "host." At least we are in good company with other frameworks.

nolanlawson: b6df56f

text/0000-declarative-root.md
nolanlawson: We have this same problem today already with scoped styles. Classes are added by the framework to every element.

I agree with you that we should not make any attempt to reconcile programmatic changes. My philosophy is "you broke it, you bought it." Developers should be discouraged from programmatically modifying things that the framework is also modifying.

FWIW, Svelte takes this same approach with their scoped styles – there is no attempt to merge programmatic styles with framework-added styles:

Screen Shot 2023-01-20 at 11 12 04 AM

nolanlawson: 2d4109e

text/0000-declarative-root.md
nolanlawson: To clarify, are you proposing that we remove all DOM-like APIs on the `this` exposed to component authors? E.g.:

```js
renderedCallback() {
  console.log(this.children)
  console.log(this.getAttribute('class'))
  console.log(this.querySelector)
}
```

?

text/0000-declarative-root.md
nolanlawson: 702d98f

text/0000-declarative-root.md
nolanlawson: To me, this seems like too much of a drastic change to the existing system. Developers are very accustomed to putting `<template>` at the top.

It also raises questions of how we would need to support `lwc:render-mode`, `lwc:preserve-comments`, etc. on `<lwc:root>`. Whereas if the two are separated, then it's clear which directives apply to which tag.

text/0000-declarative-root.md
pmdartus: Yes, I was wondering if we should deprecate all those DOM-like APIs. That said there are some APIs like `dispatchEvent` that would certainly have to stay on the `LightningElement` prototype.

nolanlawson: My understanding was that, someday, we might converge the "bridge element" and "lightning element" classes. It would be a nice simplification from my POV if `this` just referenced the element (as it does in other web component frameworks like Lit). In that world, all of these APIs would "just work."

pmdartus: I was always under the impression that the long-term goal should be to merge the `LightningElement` instance with the host element. 

But I am now questioning this, because of the tension it creates with SSR. In a world where we can achieve this merge. Does it mean that we would have to polyfill all the DOM APIs available on the host element to make LWC components compatible with SSR? 

Something that I really about this proposal is that we provide a declarative alternative to those DOM APIs. Those APIs enable developers to write isomorphic components without having to worry about the environment they are running in. Another added benefit of this declarative approach is that it enables the LWC compiler to optimize the usage of those APIs.

This is just food for thought, I don't have yet a formal opinion on this.

PR #73, opened 2023-01-19
chore: update numbering of approved and implemented RFCs.

ravijayaramappa:
this will ensure that perma links to rfcs can be shared without worrying about them changing in the future


nolanlawson (approved):
Thank you!! I was just thinking that we let these get out-of-date.

PR #75, opened 2023-01-25
RFC: LWC Localization Mechanism (Intl V2)

caridy:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/caridy-intl-v2/text/0000-declarative-root.md)

### DIFF
--- a/text/0000-intl-v2.md
+++ b/text/0000-intl-v2.md
@@ -1,0 +1,218 @@
+---
+title: LWC Localization Mechanism (Intl)
+status: DRAFT
+created_at: 2023-01-23
+updated_at: 2023-01-23
+champion: Caridy Patiño (caridy)
+rfc: 
+---
+
+# LWC Localization Mechanism (Intl)
+
+## Motivation
+
+As of today, LWC lacks a mechanism to facilitate the interpolation of content in the LWC template with the proper internalization mechanics. This proposal introduces a novel mechanism to integrate LWC Components with Fluent Messages. This is not the first time we attempt to solve this, in 2017, we put together the first RFC, which was rejected due to the complexity of the grammar on the template.
+
+As a result, we have a huge amount of components today that are doing a poor job producing complex messages by doing string replacement and concatenations. As a result, these components will simply not work on other locales, or will produce sentences that makes no sense to the user.
+
+## Use Cases
+
+* As a developer, I can opt-in to localize my template.
+* As a developer, I can interpolate text, variables and HTML tags.
+* As a developer, I can format number and dates in my template.
+* As a developer, I can select from different statements in my template.
+* As a developer, I can use grammatically to format nouns based on app configurations.
+* As a developer, I can pluralize sentences based on numeric values.
+
+## Detailed design
+
+### The Breakthrough
+
+Often when thinking about localization of components, we tend to think that in order to produce the right message, which often requires interpolation of text and data, we must provide either syntax or helper functions into the templating engine to describe what message to use in the UI and what data is needed for that message to be rendered accordingly. Essentially, looking at the message as the unit of operation. It turns out that this is the wrong pattern. Instead, if we think about the UI as the interpolation of two distinct templates that are defined in tandem to produce the final UI, then you start thinking about how to produce a snapshot of the two templates based on the current state of the component, and then how to merge them together into the final production for the UI. We already have a good understanding about how to render a template, and how to rehydrate it, how to update it over time, etc. The precedent of how we process and manipulate the HTML template is sufficient. And the problem then gets reduced to how to interpolate the two productions together rather than how to produce individual messages that are accessible to the template production.
+
+#### From Abstract To Concrete
+
+If we have a LWC Template like this:
+
+```html
+<template>
+    <p>
+        {i18n.paragraph.content}
+    </p>
+    <p><button onclick={handleBuy}>{i18n.action.buy}</button></p>
+</template>
+```
+
+And a message file associated to the same template as follow:
+
+```
+paragraph 
+    .content = { $counter ->
+        [one] {"A new"}
+        *[other] {"New"}
+    } { $accounts ->
+      [one] book was
+      *[other] books were
+   } released today.
+}.
+action 
+    .buy = Buy { $counter ->
+      [one] Book
+      *[other] Books
+   }
+}.
+```
+
+For the following component:
+
+```js
+export default class Foo extends LightningElement {
+    @api counter = 0;
+}
+```
+
+We can process both, the template and the fluent files using the same technique, you have the component state (the component instance itself) as the context to render each of these two templates, the HTML file and the fluent file. In this case, both of them can have access to all the data available in the context provided, which in this case is `counter` field, which is a numeric value.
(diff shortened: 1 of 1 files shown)


text/0000-intl-v2.md
```diff
@@ -0,0 +1,218 @@
...
+  bar.fluent
+  x.html
+  y.html
+```
+
+The way LWC can register these association is the same used for the default template, using `registerComponent`. Additionally, we must provide a way for a developer to import fluent dependencies declaratively using import declaration, the same way we allow importing templates. In this case, we must provide a new hook for the developer to return a custom fluent reference, in the same fashion that `render()` method works.
+
+#### Connecting The Dots
+
+When a component is about to be rendered, the engine must use the default associated fluent, if exist. If the hook is defined in user-land, it must be invoked to resolve the custom fluent reference. If the fluent reference is new, an internal fluent instance must be created, binding that instance to the component instance in question. The messages exported by the fluent instance must be made available via `i18n` getter on `LightningElement.prototype`. Each message exported from the fluent instance must be a getter, and must be constructed on demand when the message is accessed. This must be done prior to invocation of the template.
```
nolanlawson, line 108: > If the hook is defined in user-land, it must be invoked to resolve the custom fluent reference.

What does this mean? Are you proposing something like [programmatic style sheets](https://lwc.dev/guide/css#associate-css-stylesheets-to-a-component-programmatically), but for fluent files?

> The messages exported by the fluent instance must be made available via `i18n` getter on `LightningElement.prototype`.

By this, I assume you mean that `this.i18n.foo` works, but `this.template.querySelector('x-component').i18n.foo` does not work, correct? In other words, `i18n` is only exposed to code inside the component, not outside the component?

text/0000-intl-v2.md
```diff
@@ -0,0 +1,218 @@
...
+
+Once they understand there 4 things, they should be able to create very compelling, fully localizable components.
+
+## Open questions
+
+1. Q: How to use messages in a for-loops using the context of the loop as data?
+
+### 2. Q: Do we need a reverse parser? from Fluent to Grammaticus? if someone were going to write a label as a Fluent message, how do we make sure this can be put through the grammaticus engine for translations?
+
+Caridy: I don't think that should be the case for 1P, certain, we can lint 1P to make sure that everything that they use in the fluent file is either interpolation of existing entities or labels, or direct usage of entities and labels. For 2P/3P, the question remains.
```
nolanlawson, line 218: What is the downside of dropping Grammaticus support entirely? In other words, what do we gain by adding Grammaticus support?

Based on my read of [this post](https://engineering.salesforce.com/open-sourcing-grammaticus-308d4c9f696e/), Grammaticus is not directly exposed to component authors today; it is just used internally.

text/0000-intl-v2.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Adoption strategy
+
+### Backward Compatibility & Risks
+
+One of the compelling things about this proposal is that it does not affect in any way the existing HTML Parser, or syntax. It does not introduce any new directive either. In previous attempts to solve this problem, that was one of the major concerns. This new feature should be backward compatibility, and users of the framework can continue using the existing label resolution in platform alongside the new feature.
+
+### Prior Art
+
+FormatJS, which relies of helpers and template literals to define the messages, and format the data. Our own implementation of the label system today, which promotes the import of a platform label, and the manual transformation of such string value to produce the final result. As a result, that component is just not localizable anymore.
```
nolanlawson, line 199: It would be great to expand this section with examples from other popular frameworks (React, Vue, Svelte, Angular, etc.) and meta frameworks (Next, Nuxt, SvelteKit, Remix, etc.). How do they handle i18n?

nolanlawson (commented):
This looks great to me overall and it seems like a reasonable direction to go in. Adding a new parser and compiler output to support is painful, but it would address a real developer need. 

Since Fluent is not a web standard though, I think we need to be super-cautious about this and ensure we're hitching our wagon to the right horse. Have we evaluated all alternatives? Are there any extensions to the syntax that are promising?

text/0000-intl-v2.md
```diff
@@ -0,0 +1,218 @@
...
+
+```
+foo/
+  bar.js
+  bar.fluent
+  x.html
+  y.html
+```
+
+The way LWC can register these association is the same used for the default template, using `registerComponent`. Additionally, we must provide a way for a developer to import fluent dependencies declaratively using import declaration, the same way we allow importing templates. In this case, we must provide a new hook for the developer to return a custom fluent reference, in the same fashion that `render()` method works.
```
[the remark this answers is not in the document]
caridy: exactly! if the js exports a component (meaning it calls registerComponent), the fluent file can be associated at that point.

text/0000-intl-v2.md
caridy: > > If the hook is defined in user-land, it must be invoked to resolve the custom fluent reference.
> 
> What does this mean? Are you proposing something like [programmatic style sheets](https://lwc.dev/guide/css#associate-css-stylesheets-to-a-component-programmatically), but for fluent files?

I'm not sure anymore... if we are saying that this implicit association is sufficiently, then no. Let's think about it.
 
> > The messages exported by the fluent instance must be made available via `i18n` getter on `LightningElement.prototype`.
> 
> By this, I assume you mean that `this.i18n.foo` works, but `this.template.querySelector('x-component').i18n.foo` does not work, correct? In other words, `i18n` is only exposed to code inside the component, not outside the component?

Correct. Just like `template`.

text/0000-intl-v2.md
caridy: That is the thing... for Admins, the UI to localize their app is backed up by the Grammaticus system in Java land. I will love to drop it entirely, but that's not going to happen. Keep in mind that neither Aura, LWR nor LWC has been able to integrate it yet.

text/0000-intl-v2.md
caridy: I can certainly add more details, formatjs website does have a good set of examples for some of those frameworks.

caridy:
> This looks great to me overall and it seems like a reasonable direction to go in. Adding a new parser and compiler output to support is painful, but it would address a real developer need.

Thanks for the promptly review @nolanlawson 

> Since Fluent is not a web standard though, I think we need to be super-cautious about this and ensure we're hitching our wagon to the right horse. Have we evaluated all alternatives? Are there any extensions to the syntax that are promising?

At the moment, there is only two major formats for complex messages, one from ICU (Message Format, colloquially called MF1), and one from Mozilla (Fluent). Mozilla is proposing, or at least it was, that fluent (which was designed to fulfill the short-comings of MF1) to become MF2, and become part of ECMA402. At least that was the plan half a decade ago. jejejeje.

Although I must say that Fluent is NOT the core of this proposal, we can have the exact same proposal using MF1. The core here is the state machinery for LWC to provision and compute the messages.

nolanlawson:
@caridy Makes sense! The pros and cons of ICU vs Fluent would be interested to discuss in the RFC, though. At some point, we need to pick one. (Unless we support both? 🤷)

PR #76, opened 2023-01-27
RFC: Dynamic stylesheet replacement

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/dynamic-styles/text/0000-dynamic-stylesheet-replacement.md)

### DIFF
--- a/text/0000-dynamic-stylesheet-replacement.md
+++ b/text/0000-dynamic-stylesheet-replacement.md
@@ -120,7 +120,7 @@ Another goal of this RFC is to fully remove the [hacks that component authors ar
 ```js
 import template from './template.html';
 import darkStylesheets from './dark.css';
-import lightStylesheets from './dark.css';
+import lightStylesheets from './light.css';
 
 export default class extends LightningElement {
   @api darkMode;
--- a/text/0000-dynamic-stylesheet-replacement.md
+++ b/text/0000-dynamic-stylesheet-replacement.md
@@ -183,7 +183,7 @@ The stylesheets are injected in the following order:
 
 This allows component authors to override the stylesheets that are implicitly associated with the `template`. For instance, they could have some shared default styles defined in `template.css`, whereas `a.css` and `b.css` would selectively override those defaults.
 
-LWC still makes no attempt to remove stylesheets, or to ensure a particular ordering _between_ `render()` calls. (This is in line with existing behavior.)
+LWC still makes no attempt to remove stylesheets, or to ensure a particular ordering of injected stylesheets in the DOM when `render()` is called multiple times with multiple return values. (This is in line with existing behavior.)
 
 ### Restrictions
 
--- a/text/0000-dynamic-stylesheet-replacement.md
+++ b/text/0000-dynamic-stylesheet-replacement.md
@@ -127,8 +127,8 @@ export default class extends LightningElement {
   
   render() {
     // DEPRECATED - DO NOT DO THIS!
-    template.stylesheets = darkMode ? darkStylesheets : lightStylesheets;
-    template.stylesheetToken = darkMode ? 'dark' : 'light';
+    template.stylesheets = this.darkMode ? darkStylesheets : lightStylesheets;
+    template.stylesheetToken = this.darkMode ? 'dark' : 'light';
     return template;
   }
 }
--- a/text/0000-dynamic-stylesheet-replacement.md
+++ b/text/0000-dynamic-stylesheet-replacement.md
@@ -187,14 +187,27 @@ LWC still makes no attempt to remove stylesheets, or to ensure a particular orde
 
 ### Restrictions
 
-The returned object _must_ contain a `template` property pointing to a valid HTML template function.
+The returned object _may_ contain a `template` property pointing to a valid HTML template function.
 
 The returned object _may_ contain a `stylesheets` property pointing to a valid stylesheet function or recursively deep array of stylesheet functions.
 
-For compatibility with the `static stylesheets` property, the validation is exactly the same. The validation is skipped only if the `stylesheets` property does not exist on the object.
+The returned object _must_ contain either the `template` or `stylesheets` key. Any other object is invalid and throws a runtime error:
 
 ```js
-return { template };              // No validation required
+return {}; // Invalid
+return { foo }; // Invalid
+```
+
+However, extraneous keys on an otherwise valid object are permitted:
+
+```js
+return { template, stylesheets, foo }; // Valid - `foo` is ignored
+```
+
+For compatibility with the `static stylesheets` property, stylesheet validation is exactly the same. The validation is skipped only if the `stylesheets` property does not exist on the object.
+
+```js
+return { template };              // No stylesheet validation required
 return { template, stylesheets }; // Stylesheets must be validated
 ```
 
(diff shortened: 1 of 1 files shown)


AllanOricil:
I believe it can be simplified a little bit more. What if you do it as shown below. There could exist a prop in lightningElement that have references to style sheets. When changed, style sheets are recomputed and passed to template. This way we won't need to import template just to pass it along the resolved style sheets inside the render method. It is like a sintatic sugar. 

````js
 import dark from './dark.css'; 
 import light from './light.css'; 
  
 export default class extends LightningElement { 
   @api darkMode = false; 
    
    this.styleSheets = this.darkMode? dark : light;
 } 
 ```

`this.styleSheets` can receive an object, array or function that returns either an object or array.


When working with objects style sheets are the object keys, and their values can be true or false. If true, the style sheet is applied. For example:

````js
this.styleSheets = {
   styleSheet1 : true
   styleSheet2 : false,
   ...
}
````

When working with an array,  all style sheets in a given array are applied:

````js
this.styleSheets = [ styleSheet1, styleSheet2 ];
````

It could also accept functions

````js
this.styleSheets = this.getStyleSheets(); // components function

Or
//inline function
this.styleSheets = () => { 
   let computedStyleSheets;
   ...
   return computedStyleSheets;
}


Or
import { getStyleSheets } from 'c/utility';

//this function is used in so many other components that it was exported as a module, and then imported everywhere it is used
this.styleSheets = getStyleSheets(); 
````

**Comment by nolanlawson** [comment]
@AllanOricil Thanks for the feedback! A few comments:

First off, there is an existing [programmatic style injection RFC](https://rfcs.lwc.dev/rfcs/lwc/0000-programmatic-style-injection) which is already being shipped in the latest open-source version of LWC. This defines a `static stylesheets` property. Unfortunately I think the presence of this feature means that having a new non-static `stylesheets` property (or anything similarly-named) would risk confusing developers.

> In my opinion, dynamically choosing templates and style sheets are two separate problems/concerns. Maybe developers need to switch between style sheets but not templates.

This is a strong point. The reason I choice `render` was mostly to re-use the existing infrastructure for rendering multiple templates. But you're right that it's a bit odd to have to pass in a `template` if you only have one.

I think your "`styleSheets` hook" suggestion is a neat idea. Except I would maybe rename it:

```js
   @api darkMode = false; 
   renderStyleSheets() {
      return this.darkMode? dark : light;
   }
 } 
```

The main advantage I see here is that we don't need to have `render` return an object, and to explain what happens if the properties are missing.

Also, it's pretty clear that if you have an implicitly-associated HTML file (and no `render`), then that file will be rendered alongside the stylesheets. (That said, I could also make the `template` property optional in the current RFC, which would have the same effect.)

The main downside I see is that this introduces additional API surface and a new `render*` function for developers to learn. But maybe it's a better approach?

AllanOricil:
@nolanlawson 
> The main downside I see is that this introduces additional API surface and a new `render*` function for developers to learn. But maybe it's a better approach?

Today I changed my mind when I started to think what render means in other contexts, like rendering 2D|3D objects. When using libraries to create these objects, render means draw shapes/vertices, then apply colors, texture, materials, shaders to make them look good. The render method concern is a single one: do everything that is necessary to draw something in the screen. So, If we think that templates are shapes/vertices, and the rest are like styles and style sheets, then It makes sense to handle both in a single render method.
The main reason I thought about a second method was just because I wanted to avoid unnecessary `imports`. Then I started to think if there could exist another way to enable the `render` method handle both features at the same time, and independently of each other, and I found another solution:

What if the `render` method returns 3 possible object types `< Render | Template | StyleSheet | StyleSheet [] >` ?
The `Render` object takes care of use cases where both dynamic style sheets and templates are required. This approach would be compatible with the current API and avoid unnecessary imports.


### dynamic templates only
````js
 import template_1 from './template_1.html'; 
 import template_2 from './template_2.html'; 
  
    render_template_1 = true;
 
    render(){
      return this.render_template_1 ? template_1 : template_2;
    }
 } 
````

###  dynamic style sheets only
````js
 import dark from './dark.css'; 
 import light from './light.css'; 
  
   @api darkMode = false; 
    
    render(){
      return this.darkMode? dark : light;
    }
 } 
````
###  dynamic style sheets and template, single style sheet
````js
 import template_1 from './template_1.html'; 
 import template_2 from './template_2.html'; 
 import dark from './dark.css'; 
 import light from './light.css'; 
  
   render_template_1 = true;
   @api darkMode = false; 
    
    render(){
      return { 
         template: this.render_template_1 ? template_1 : template_2,
         styleSheet:  this.darkMode? dark : light
      }
    }
 } 

````

###  dynamic style sheets with multiple style sheets
````js
 import dark from './dark.css'; 
 import light from './light.css'; 
 import common from './common.css'; 
  
   render_template_1 = true;
   
   @api darkMode = true;
    
    render(){
      return this.darMode ? [dark, common] : [light, common]
    }
 } 

````

###  dynamic style sheets and template, with multiple style sheets
````js
 import template_1 from './template_1.html'; 
 import template_2 from './template_2.html'; 
 import dark from './dark.css'; 
 import light from './light.css'; 
 import common from './common.css'; 
  
   render_template_1 = true;
   
   @api darkMode = true;
    
    render(){
      return { 
         template: this.render_template_1 ? template_1 : template_2,
         styleSheet[]:   this.darMode ? [dark, common] : [light, common]
      }
    }
 } 

````

**Comment by nolanlawson** [comment]
I think we can achieve a similar thing by allowing the `render()` function to return `{ stylesheets }` (no `template`). In this case, the `template` would be implied.

You're right that we could also distinguish between stylesheets and templates being returned directly, but I wonder if the shorthand is worth it – I personally tend to prefer explicit compared to implicit. That said, it could always be added later in a backwards-compatible way.

**Comment by nolanlawson** [comment]
I've modified the RFC to support returning `{ stylesheets }` (no `template`),  or `stylesheets` directly as a shorthand. In this case, the template is just the implicitly-associated `*.html` file.

File: text/0000-dynamic-stylesheet-replacement.md [hunk 1]
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+}
+```
+
+The stylesheets are injected in the following order:
+
+1. Stylesheets implicitly associated with the `template`
+2. Stylesheets from the returned object's `stylesheets`
+3. Static `stylesheets` on the component
```
rui-rayqiu, line 182: I'm wondering what is a use case that developers would use both 2 and 3 here? Is it that technically I can just use the returned object's `stylesheets` so I don't need to modify static `stylesheets` or is there a case that would only work for 3?

text/0000-dynamic-stylesheet-replacement.md
```diff
@@ -0,0 +1,260 @@
...
+
+This allows component authors to override the stylesheets that are implicitly associated with the `template`. For instance, they could have some shared default styles defined in `template.css`, whereas `a.css` and `b.css` would selectively override those defaults.
+
+LWC still makes no attempt to remove stylesheets, or to ensure a particular ordering _between_ `render()` calls. (This is in line with existing behavior.)
```
rui-rayqiu, line 186: Just for my understanding, what is the "ordering between render() calls" here referring to?

text/0000-dynamic-stylesheet-replacement.md
nolanlawson: Static `stylesheets` are the only thing that allows for extending styles from a superclass. If you only use the `render()` API, you would be negating any of the superclass's styles and using only your own.

```js
static stylesheets = [...super.stylesheets, ...myStylesheets]
```

So my thought process is: use `static stylesheets` if you want your subclasses to reuse and extend them. Use `render()` otherwise.

text/0000-dynamic-stylesheet-replacement.md
nolanlawson: I should be more specific. I mean ordering of injected stylesheets w.r.t the order that `render()` was called. I'll update this.

nolanlawson:
We have decided not to move forward with this RFC. Instead, we have [documented several solutions](https://lwc.dev/guide/css#deprecated-template-mutation) to replace the current `template.stylesheets = ...` hacks.

PR #77, opened 2023-02-08
RFC: Dual light/shadow components

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/dual-mode/text/0000-dual-light-shadow.md)

### DIFF
--- a/text/0000-dual-light-shadow.md
+++ b/text/0000-dual-light-shadow.md
@@ -118,7 +118,8 @@ it can also be applied to LWC components referenced inside a template:
 </template>
 ```
 
-When used in this way, the value of `lwc:render-mode` can only be a static string, and it can only be `"light"` or `"shadow"`.
+When used in this way, the value of `lwc:render-mode` can only be a static string. It can be set to `"light"` or `"shadow"`,
+which controls the mode that the child component renders in.
 
 If `"dual"` is used in this context, a compile-time error is thrown:
 
@@ -128,6 +129,54 @@ If `"dual"` is used in this context, a compile-time error is thrown:
     <x-component lwc:render-mode="dual"></x-component>
 </template>
 ```
+
+Only three values are accepted in this context: `"light"`, `"shadow"`, or `"inherit"`. Any other value will throw a compile-time error.
+
+### Inheritance
+
+`"inherit"` is a special value of `lwc:render-mode` that can only be used in the context of a child LWC component:
+
+```html
+<template lwc:render-mode="dual">
+    <x-component lwc:render-mode="inherit"></x-component>
+</template>
+```
+
+This instructs the `<x-component>` to render using whatever mode the parent component is using – either shadow or light. If `<x-component>` is not a dual-mode component, then a runtime error will be thrown.
+
+For the parent component, however, `"inherit"` can be used in non-dual-mode components as well as dual-mode components:
+
+```html
+<template lwc:render-mode="light">
+    <!-- x-component will render as light -->
+    <x-component lwc:render-mode="inherit"></x-component>
+</template>
+```
+
+In this case, the same rules apply – the child component uses the mode inherited from its parent.
+
+The above example is equivalent to:
+
+```html
+<template lwc:render-mode="light">
+    <!-- x-component will render as light -->
+    <x-component lwc:render-mode="light"></x-component>
+</template>
+```
+
+For the purposes of inheritance, `<slot>` contents are considered to be children of the slotting component, not the
+slottable component.
+
+```html
+<template>
+    <x-slottable>
+        <!-- Mode is inherited from this template, not from x-slottable's template -->
+        <x-component lwc:render-mode="inherit"></x-component>
+    </x-slottable>
+</template>
+```
+
+### Coherence
 
 If a template is declared to be `lwc:render-mode="dual"`, then the corresponding component must also have
 `static renderMode = 'dual'`, and vice versa.


text/0000-dual-light-shadow.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+When used in this way, the value of `lwc:render-mode` can only be a static string, and it can only be `"light"` or `"shadow"`.
+
+If `"dual"` is used in this context, a compile-time error is thrown:
+
+```html
+<template>
+    <!-- Invalid -->
+    <x-component lwc:render-mode="dual"></x-component>
```
jmsjtu, line 124: I think this is implied but wanted to point out `lwc:render-mode` should also work with `<lwc:component>` too (renders to an LWC custom element).

nolanlawson: Correct, yes. The two RFCs should be complementary; I don't see why the two directives couldn't work together.

nolanlawson: Added some clarifications about dynamic components.

text/0000-dual-light-shadow.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Non-LWC components
+
+`lwc:render-mode` can only be applied to LWC components. If it's applied to any other element, including an 
+`lwc:external` component, then a compile-time error is thrown:
+
+```html
+<template>
+    <!-- Invalid -->
+    <third-party lwc:render-mode="light"></third-party>
```
caridy, line 146: How do you know at built-time that this is a non-LWC? You don't. Is this example missing the `lwc:external` directive?

caridy (commented):
Overall, I'm supportive of this proposal.

text/0000-dual-light-shadow.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Basic example
+
+A component can declare its `lwc:render-mode` to be `"dual"`:
+
+```html
+<!-- component.html -->
+<template lwc:render-mode="dual">
+    <h1>Hello world</h1>
+</template>
```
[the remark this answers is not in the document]
nolanlawson: @AllanOricil Thanks for the feedback! I appreciate it a lot. 🙂 

> couldn't the render-mode directive be simplified to a boolean

I believe the reason we went with a string enum originally was because we weren't sure if someday we would want to support more exotic cases, like closed shadow mode or ["open-stylable" shadow mode](https://github.com/WICG/webcomponents/issues/909). Although maybe that would argue that we should use a different name for `"dual"` – perhaps `"any"` or something.

> What if dual is the default render mode for every component?

I can see the case for "dual by default," but the main issue I see is that components that were not designed with light DOM in mind (or with shadow DOM in mind, for light DOM components) may be forced into a mode that breaks them. Whereas with my proposal, light/shadow components only render in that one mode unless they explicitly opt-in.

>  I believe that using a variable called `useLightDom` inside the `render()`

This would be very convenient from the component authoring perspective, but unfortunately it would greatly complicate how the engine is implemented for the component to switch between light and shadow mode dynamically during its lifetime. For instance, if light-DOM styles were added to the `<head>`, now those have to be converted into shadow DOM styles in the shadow root (or vice versa). And while it's possible to attach a new shadow root to an existing light DOM component,  it's not possible to remove a shadow root (the browser does not provide a "remove" analog to [`attachShadow`](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow)).

> 3rd party companies would have document which components render in light mode no matter the context, because developers wouldn't be able to determine this information by just reading the parent component's template.

I don't think they would need to document it either way, because we should be able to expose the static `renderMode` in the [metadata](https://www.npmjs.com/package/@lwc/metadata). So it would be available to e.g. a drag-and-drop builder UI.

text/0000-dual-light-shadow.md
nolanlawson: Sorry, yes. I missed the `lwc:external` here.

text/0000-dual-light-shadow.md
```diff
@@ -0,0 +1,297 @@
...
+A non-exhaustive list:
+
+- `::slotted` selectors in CSS are permitted, but will only work when running in native shadow mode.
+- `this.template` is undefined for light DOM components.
+- Slots are lazy in light DOM and eager in native shadow DOM.
+- Scoped and non-scoped stylesheets will both behave differently in light versus shadow mode.
+
+#### Runtime modifications
+
+A component cannot dynamically modify its `static renderMode`; changing it after instantiation has no effect.
```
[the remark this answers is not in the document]
nolanlawson: Correct, yes!

nolanlawson:
Update: I added an `"inherit"` keyword that allows the mode to propagate from parents to children. Seems like a nice convenience feature.

text/0000-dual-light-shadow.md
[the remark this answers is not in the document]
caridy: > What if dual is the default render mode for every component?

That is non-backward compatible for you. Additionally, we could say that you can, as a consumer of the component, always only select the mode that you know the consumer component works... saying that it is a responsibility of the consumer to always specify has two big drawbacks that I can't really sign off on:

1. if the current version of the component does work in both modes (just by pure luck), and the consumer select light, then the owner of the component modifies that, and inadvertently breaks light, he never really sign on for that, now we have breaking changes.
2. now the consumer has to learn that there is to mode, and which components are broken and needs to be forced to a particular mode of operation, that seems very wrong to me.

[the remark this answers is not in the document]
AllanOricil: 1- I think this could be mitigated if 3rd party developers release a MAJOR version for their package. Salesforce could instruct 3rd party developers to do this when forcing a render mode on their components, if they were previously "dual". Consumers would have to the read release notes to understand clearly why a MAJOR bump happened, and what could be the implications before upgrading their package.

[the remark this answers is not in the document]
AllanOricil: I understand it now 👍

text/0000-dual-light-shadow.md
```diff
@@ -129,6 +130,54 @@ If `"dual"` is used in this context, a compile-time error is thrown:
...
+```
+
+This instructs the `<x-component>` to render using whatever mode the parent component is using – either shadow or light. If `<x-component>` is not a dual-mode component, then a runtime error will be thrown.
+
+For the parent component, however, `"inherit"` can be used in non-dual-mode components as well as dual-mode components:
+
+```html
+<template lwc:render-mode="light">
+    <!-- x-component will render as light -->
+    <x-component lwc:render-mode="inherit"></x-component>
```
caridy, line 152: two main comments:

1.
> `<x-component lwc:render-mode="inherit"></x-component>`
it could be `<x-component lwc:render-mode="dual"></x-component>` and it will work the same, it does the same as `inherit` but without introducing yet another token.

2.
it seems to me that this should be an error, if you're choosing the mode in the template, why using inherit down here? you should be specific, because you have chosen the context in which this mode will operate by doing it at the template level. you're not typing less, you're not gaining anything... it is now just more ambiguous for the developer to try to remember what mode the template is running (meaning scroll up).

text/0000-dual-light-shadow.md
```diff
@@ -129,6 +130,54 @@ If `"dual"` is used in this context, a compile-time error is thrown:
...
+```
+
+For the purposes of inheritance, `<slot>` contents are considered to be children of the slotting component, not the
+slottable component.
+
+```html
+<template>
+    <x-slottable>
+        <!-- Mode is inherited from this template, not from x-slottable's template -->
+        <x-component lwc:render-mode="inherit"></x-component>
```
caridy, line 174: in this example, who dictates the mode? how can an author look at this code an know? what is the process of knowing what mode this is going to run on?

nolanlawson: > in this example, who dictates the mode?

The top-level `<template>` on Line 171 dictates the mode. In this case, it will be shadow mode, because there is no `render-mode` on line 171.

> how can an author look at this code an know?

They would need to glance up to the top `<template>` in the template they are currently authoring. 🙂 

> what is the process of knowing what mode this is going to run on?

Again, glance up top. 🙂 And if `<x-component>` does not support dual mode, you will get an error.

text/0000-dual-light-shadow.md
nolanlawson: > it could be `<x-component lwc:render-mode="dual"></x-component>` and it will work the same, it does the same as inherit but without introducing yet another token.

We could do that, yeah. I don't have a strong opinion.

>  you're not typing less, you're not gaining anything

Are you saying that the "inherit" behavior should be implied by omitting the `lwc:render-mode` on the `<x-component>` altogether? This is fine, except that it introduces ambiguity on how a component behaves depending on the template it is rendered in. In a light/shadow template, it will be whatever default the component author (in this case the author of `<x-component>`) chose, whereas in a dual template, it will inherit from that template. To me, this is more confusing than making the omission of `lwc:render-mode` always mean "whatever default the component author chose."

PR #78, opened 2023-03-02
RFC: Element Internals and Form Association

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/element-internals/text/0000-element-internals.md)

### DIFF
--- a/text/0000-element-internals.md
+++ b/text/0000-element-internals.md
@@ -1,0 +1,224 @@
+---
+title: Element Internals and Form Association
+status: DRAFTED
+created_at: 2022-03-02
+updated_at: 2022-03-02
+champion: Nolan Lawson (nolanlawson)
+pr: https://github.com/salesforce/lwc-rfcs/pull/78
+---
+
+# Element Internals and Form Association
+
+## Summary
+
+This proposal provides component authors with a way to use the [`ElementInternals`](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) API, as well as the related [Form-Associated Custom Elements](https://web.dev/more-capable-form-controls/#lifecycle-callbacks) (FACE) lifecycle API.
+
+## Basic example
+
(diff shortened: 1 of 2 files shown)


text/0000-element-internals.md
```diff
@@ -0,0 +1,235 @@
...
+To avoid leaking private component details, accessing internals from outside a component is not allowed:
+
+```js
+const component = document.querySelector('my-component');
+console.log(component.attachInternals); // undefined
+```
+
+`this.attachInternals()` can only be called once, and will throw an error if called twice. (This is how vanilla custom elements work.) So component authors are encouraged to guard against this by avoiding calling `this.attachInternals()` in a lifecycle hook that may be called multiple times, such as `connectedCallback` or `renderedCallback`.
+
+In Server-Side Rendering (SSR), `this.attachInternals` will be undefined.
```
nolanlawson, line 110: Maybe this should throw an explicit error instead in SSR? This might be a better dev experience.

text/0000-element-internals.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Deviations from the standard `ElementInternals` object
+
+The `ElementInternals` instance returned by `attachInternals` is exactly the same as the one provided by the browser in a native custom element, with one exception: `internals.shadowRoot` does not actually return an instance of the `ShadowRoot`, but rather the same proxy object returned by `this.template`:
```
nolanlawson, line 114: I just realized this is wrong – [we actually do return the `shadowRoot` directly it seems](https://github.com/salesforce/lwc/blob/003d37e97ea84019b7f9d54a0c7d5255510d1d10/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L488-L500).

nolanlawson:
This is implemented now!

PR #80, opened 2023-05-19
chore: update rfc statuses

jmsjtu:
This PR updates the statuses of several RFCs to `IMPLEMENTED`.

I've also added an RFC statuses definitions section to the `README` to clarify what each status means.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,13 @@ the time of the next major release; therefore we try to keep each RFC
 document somewhat in sync with the language feature as planned,
 tracking such changes via followup pull requests to the document.
 
+### RFC Statuses
+
+1. DRAFTED - RFC is actively in development and seeking feedback.
+2. ACTIVE - RFC is ready for implementation.
+3. APPROVED - RFC is being implemented.
+4. IMPLEMENTED - Implementation has been completed.
+
 ## Implementing an RFC
 
 The author of an RFC is not obligated to implement it. Of course, the
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Some changes do not require an RFC:
 
 In short, to get a major feature added to Lightning Web Components, one usually first gets
 the RFC merged into the RFC repo as a markdown file. At that point the RFC
-is 'active' and may be implemented with the goal of eventual inclusion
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -112,6 +112,13 @@ the time of the next major release; therefore we try to keep each RFC
 document somewhat in sync with the language feature as planned,
 tracking such changes via followup pull requests to the document.
 
+### RFC Statuses
+
+1. DRAFTED - RFC is actively in development and seeking feedback.
+2. ACTIVE - RFC is ready for implementation.
+3. APPROVED - RFC is being implemented.
+4. IMPLEMENTED - Implementation has been completed.
```
jmsjtu, line 120: I thought it would be helpful to add definitions to each status as I wasn't sure what each status meant/how the status progression worked.  

For example, does ACTIVE come before APPROVED?  

I'm also not sure what `APPROVED` means but based on how it's using in the previous RFCs it is either:

1. The RFC is actively being implemented.
2. The same thing as IMPLEMENTED.

Would love to get some feedback on this section and open to changing the wording on any of these.

text/0129-declarative-shadow-dom-polyfill.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Declarative ShadowDOM polyfill
-status: APPROVED
+status: IMPLEMENTED
```
nolanlawson, line 3: [Not actually implemented yet](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001RkJc4YAF/view)

nolanlawson (changes requested):
See above, otherwise LGTM

README.md
nolanlawson: We based our implementation on [Ember's](https://github.com/emberjs/rfcs#stages), looks like they're not quite the same anymore (if they ever were).

I would remove ACTIVE entirely because I'm not sure there's a distinction between ACTIVE and DRAFTED.

nolanlawson: I'd say APPROVED means "The RFC is actively being implemented"

nolanlawson (approved):

PR #82, opened 2023-11-08
first draft of the rfc for signals in lwc

caridy:


caridy:
@jodarove @divmain I could not find the original work that we did on this years ago.

jodarove:
> @jodarove @divmain I could not find the original work that we did on this years ago.

It is on this branch: https://github.com/salesforce/lwc/tree/jodarove/lwc-stores-v1

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,373 @@
...
+* __Educational Materials__: Comprehensive updates to documentation, tutorials, and example code will be essential to support developers in adopting the new model.
+
+### Potential for Misuse
+
+* __Overuse of Signals__: There's a risk that developers might overuse signals for state management, leading to unnecessarily reactive code, which can be less performant and harder to maintain. This is specially relevant because of the runtime detection of `.value` to maintain backward compatibility with existing components. Any misused on the signals can affect the runtime performance of the component and therefore the overall performance of the application.
+* __Subscription Management__: Incorrect handling of subscriptions could introduce memory leaks or lead to unexpected behaviors if not managed correctly. This is specially important because the store implementation doesn't have access anymore to the component's lifecycle hooks, instead it would just stop receiving updates of the stores piped thru `this.$api`, meaning that we are at the mercy of the garbage collector.
```
mburr-salesforce, line 307: This is the one that concerns me the most. I can see *lots* of component authors messing up subscriptions. `@wire` has issues, but one of the nice things about it is that there's no way for a component author to mess up the subscription.

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,373 @@
...
+    subscribe(callback) { /* subscribe to changes */ }
+    unsubscribe(callback) { /* unsubscribe from changes */ }
+}
+```
+
+__NOTE:__ _We can work the details of this protocol. It could be more simple, or hopefully similar to what is out there._
+
+### Integration with Components
+
+Components should treat signals as internal reactive state mechanisms. While a signal can technically be passed down to child components, it's recommended to pass only the necessary data, typically the `.value` of the signal, to keep the child components decoupled from the parent's state management strategy.
```
mburr-salesforce, line 81: I understand & agree with the sentiment here, but I think injecting a signal into a component is going to be the key to making low-code scenarios work?

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,373 @@
...
+export default class ExampleComponent extends LightningElement {
+    @api recordId;
+    record = new RecordStore(this.$api.recordId);
+
+    updateRecord(record) {
+      // do something with the record
+    }
+
+    connectedCallback() {
+        this.record.subscribe(record => this.updateRecord(record));
```
mburr-salesforce, line 268: Can this happen before `connectedCallback`, eg in `constructor`?

mburr-salesforce:
I don't *think* so, but does anything here prevent signals from provisioning values to the component before its initial render? That is, using a signal does not automatically force my component to be rendered twice?

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,373 @@
...
+The example above shows how to integrate with external stores using a static value. But what if the value is actually a property of the component that might change over time? To facilitate interaction with external stores, LWC will provide access to reactive API properties:
+
+```javascript
+export default class ExampleComponent extends LightningElement {
+    @api recordId;
+    record = new RecordStore(this.$api.recordId);
+}
+```
+
+In this design, `this.$api` is a reactive wrapper around the component's API properties, enabling stores to react to changes in properties like recordId. This reensambles the internal mechanism used today in LWC via `vm.cmpProps` that is not exposed to developers.
```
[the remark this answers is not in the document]
caridy: yes! the argument would be a signal-like object created by LWC.

mburr-salesforce:
I expect it would take someone a couple days to figure out they can do something like this:

```
import type { WireAdapter, WireConfigValue, WireDataCallback } from 'lwc'

type SignalCallback = (value: any) => void
type Signal = {
    value: any
    subscribe: (callback: SignalCallback) => void
    unsubscribe: (callback: SignalCallback) => void
}

export default class SignalWire implements WireAdapter {
    signal?: Signal

    constructor(private callback: WireDataCallback) {
    }

    connect(): void {
        this.handleSubscriptions(this.signal)
    }

    disconnect(): void {
        this.handleSubscriptions()
    }

    update(config: WireConfigValue): void {
        this.handleSubscriptions(config.signal)
    }

    private handleSubscriptions(signal?: Signal): void {
        if (this.signal) {
            this.signal.unsubscribe(this.callback)
        }

        this.signal = signal
        if (signal) {
            this.callback(signal.value)
            signal.subscribe(this.callback)
        }
    }
}
```

Should we just preemptively provide this?

text/0000-lwc-signals.md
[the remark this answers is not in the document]
mburr-salesforce: Would `RecordStore` be expected to do any `.unsubscribe` cleanup when the component is disconnected/disposed?

text/0000-lwc-signals.md
seckardt: @caridy I think I remember a demo when signals were introduced for either Solid or React, where signals were explicitly used to skip levels inside component hierarchies (kind of like a poor-man's state management), but where this was key to some major performance improvements observed. Using the following technique there lead to signals "magically" updating their values inside the HTML elements directly, without relying/executing a full re-render of the involved components. I know that this would scare the hell out of you, but that approach was like what brought "lightning" to the demo back then.

```js
// bundle `acme/signals`
import { signal } from 'signals';

export const count = signal(0);
```

```js
// bundle `acme/parent`
import { count } from 'acme/signals';

class Parent extends LightningElement {
    handleClick() {
         count.value++;
    }
}
```

```js
// bundle `acme/grandGrandChild`
import { count } from 'acme/signals';

class GrandGrandChild extends LightningElement {
    get count() {
        return count;
    }
}
```

```html
<!-- bundle `acme/grandGrandChild` -->
<div>{count.value}</div>
```

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,373 @@
...
+* __Interop with Non-Signal Components__: Components designed without signals in mind may require additional logic to work within the new reactivity model, potentially leading to a fragmented ecosystem during the transition period.
+
+### Performance Implications
+
+* __Granularity of Reactivity__: The fine-grained nature of Signals may introduce performance concerns if developers do not carefully consider the reactivity scope and frequency of updates.
+* __Proxy Overhead Elimination__: While removing the proxy overhead of `@track` can lead to performance gains, it's essential to validate these improvements across a wide range of use cases to ensure that they hold true in practice.
+
+## Alternatives
+
+### 1. Continued Use of `@track` and `@wire`
```
seckardt, line 321: @mburr-salesforce @caridy It would be an interesting option to have wires that pass signals instead of deeply frozen objects. That would open things up for a lot of simplifications when for instance passing wired values into e.g. forms without having to create deep clones. But it would of course also open things up for all the mentioned misuse. Still something to assess further in terms of perf pros and cons maybe. Definitely would have an impact on perf, but also on development models as it would allow e.g. mutations without following established approaches like data down and events up.

caridy:
> I don't think so, but does anything here prevent signals from provisioning values to the component before its initial render? That is, using a signal does not automatically force my component to be rendered twice?

@mburr-salesforce that's correct. Since a signal can be initialized, plus its callback can be called at any given time, a component with access to a signal can gain access to a value for the first render.

caridy:
> I expect it would take someone a couple days to figure out they can do something like this:
> 
> ```
> import type { WireAdapter, WireConfigValue, WireDataCallback } from 'lwc'
> 
> type SignalCallback = (value: any) => void
> type Signal = {
>     value: any
>     subscribe: (callback: SignalCallback) => void
>     unsubscribe: (callback: SignalCallback) => void
> }
> 
> export default class SignalWire implements WireAdapter {
>     signal?: Signal
> 
>     constructor(private callback: WireDataCallback) {
>     }
> 
>     connect(): void {
>         this.handleSubscriptions(this.signal)
>     }
> 
>     disconnect(): void {
>         this.handleSubscriptions()
>     }
> 
>     update(config: WireConfigValue): void {
>         this.handleSubscriptions(config.signal)
>     }
> 
>     private handleSubscriptions(signal?: Signal): void {
>         if (this.signal) {
>             this.signal.unsubscribe(this.callback)
>         }
> 
>         this.signal = signal
>         if (signal) {
>             this.callback(signal.value)
>             signal.subscribe(this.callback)
>         }
>     }
> }
> ```
> 
> Should we just preemptively provide this?

We are providing a mechanism to interact, reactively, with LWC engine, that's all. If people want to do these things... sure. Not sure I understand the point though.

text/0000-lwc-signals.md
caridy: That would be a choice to make for the owner of the adapter, not LWC.

text/0000-lwc-signals.md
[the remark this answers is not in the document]
caridy: I would update the API to remove the unsubscribe all together to match what other frameworks are doing. As for the development experience, they should try to unsubscribe, but LWC would not track any of that. When LWC uses (internally) an store, it would unsubscribe accordingly

text/0000-lwc-signals.md
caridy: The only thing to really try to figure out is whether or not we should support `{count}` vs `{count.value}` in the templates, a la-vue. The hesitation here is the signal down to other components, the ambiguity he being:

```
<template>
   <input value={count} />
   <x-foo count={count}></x-foo>
</template>
```

Should `count` property on foo receive a signal or its value? That's the problem.

mburr-salesforce:
> We are providing a mechanism to interact, reactively, with LWC engine, that's all. If people want to do these things... sure. Not sure I understand the point though.

@caridy The point is that it's possible to use `@wire` to do subscription management for your signals. Your `@wire`d function just becomes your signal subscription callback & everything else is handled for you.

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+__Note:__ _since the record might not be ready, it might require some additional logic to handle the case when the record is not available. But this is the same they do today with `@wire`._
+
+### Reactive API Properties
+
+The example above shows how to integrate with external stores using a static value. But what if the value is actually a property of the component that might change over time? To facilitate interaction with external stores, LWC will provide access to reactive API properties:
+
+```javascript
+export default class ExampleComponent extends LightningElement {
+    @api recordId;
+    record = new RecordStore(this.$api.recordId);
```
nolanlawson, line 156: Why the dollar sign on `$api`? Historically we have not put dollar signs for component-specific props like `this.refs` and `this.template`.

If the concern is backwards compatibility, then we can gate this behind an API version.

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+1. __Performance Constraints__: The current model requires components to re-render fully to reflect state changes, which can be inefficient, especially for large and complex component trees.
+
+By adopting a Signals-based reactivity model, LWC can overcome these challenges:
+
+1. __Framework Agnosticism__: With Signals, LWC components can interact with any external data store that follows the protocol. This change would make LWC more agnostic and flexible, allowing for easier integration with various state management patterns and libraries.
+1. __Simplified State Management__: Signals provide a straightforward way to manage reactive data. By deprecating the @wire and @track decorators, we can reduce complexity, making it easier for developers to write and maintain component logic.
+1. __Performance Optimizations__: The proposed model opens up possibilities for performance optimizations by allowing selective rehydration of the UI. Components could update the DOM directly in response to state changes without a full re-render, leading to faster updates and reduced resource consumption.
```
nolanlawson, line 28: Nit: it's not clear to me that signals are _required_ to achieve fine-grained reactivity. E.g. the design described in https://github.com/salesforce/lwc/issues/3624 calls for adding fine-grained reactivity to template loops, without any user-facing API changes.

Also, it is entirely possible to implement signals while still re-running the entire template. It would be useful to see examples where signals unlock perf optimizations that we couldn't get otherwise.

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+
+    disconnectedCallback() {
+        this.recordUnsubscript();
+    }
+}
+```
+
+It is important to highlight that this is the less common method of using a wire adapter. With the new approach, it becomes a little bit more cumbersome, but it is still possible to do it, and it offers a lot more flexibility since you can now use any store that follows the protocol, not just Salesforce-specific services.
+
+## Backward Compatibility
```
nolanlawson, line 278: The easiest way to achieve backwards compatibility would be to use API versioning. As a bonus, we could disallow `@wire` and `@track` based on an API version. (May need to be loosened for first-party components.)

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+2. Exposing more internal implementation details to developers, like the `$api` property, could lead to misuse and potential security issues as described in 1 if the value is an object. An alternative here is to continue doing the readonly proxy for non-primitive values, but that would be a step back in terms of performance.
+
+## Open Discussion Topics
+
+* The signal/store protocol. Except for the `.value`, which seems to be a requirement, the rest of the protocol is open for discussion.
+* The `this.$api` as a way to access internal signal objects created by LWC, that seems to be a good compromise, but we can discuss alternatives.
```
nolanlawson, line 372: It would be great to have examples from other frameworks in this document. I know nearly every framework is moving to signals these days, but I'm not familiar with the exact API patterns used in each one.

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+Objects can be wrapped in Signals to make them reactive. However, changes to nested properties must be handled by replacing the entire object:
+
+```javascript
+import { signal } from 'signals';
+
+export default class ExampleComponent extends LightningElement {
+    position = signal({ x: 0, y: 0 });
+
+    move(x, y) {
+        this.position.value = { ...this.position.value, x, y };
```
ravijayaramappa, line 124: This does become a challenge for nested objects from these two perspectives:
1. Error prone trying to recreate a new object for every combination of change
2. Requires boiler plate code to recreate the object.

nolanlawson: It's also potentially a perf tax due to cloning the object.

text/0000-lwc-signals.md
caridy: I don't think we need to remove either of those to have a fully functional signal integration. They are just two separate features, and eventually we can start the deprecation strategy.

text/0000-lwc-signals.md
caridy: sure. but I don't think this RFC is about that particular aspect of it, but the enablement of stores.

AllanOricil:
@caridy if I have N child components on the same level (siblings), all subscribed to the same signal, would there exist a way to determine the order components are updated and re-rendered?

text/0000-lwc-signals.md
divmain: Agreed 100%. I'm going to attempt to put together a survey of signals implementations & alternate approaches across web frameworks ahead of our dev sync discussion.

divmain:
@AllanOricil

> if I have N child components on the same level (siblings), all subscribed to the same signal, would there exist a way to determine the order components are updated and re-rendered?

This is a good question. Without deliberately implementing something else, the default behavior would rely on the order that components had subscribed to the signal. Is there a specific scenario you're thinking through where you'd want some other, deterministic behavior?

text/0000-lwc-signals.md
divmain: > This does become a challenge for nested objects from these two perspectives:
> 1. Error prone trying to recreate a new object for every combination of change
> 2. Requires boiler plate code to recreate the object.

Because this RFC deals with the low-level primitives that we'd provide, this isn't a huge concern to me. I'm hoping/expecting that this won't be heavily used by individual components and, instead, abstractions (state manager, data provider, etc) would be built on top of signals that provide better ergonomics.

For example, in a state manager, we could use/replicate [Immer](https://immerjs.github.io/immer/) to provide a friendlier facade.

text/0000-lwc-signals.md
divmain: Although it'd be really attractive to retire `@wire`, I think it shouldn't be particularly emphasized in the signals discussion. Initially, signals (or whatever alternative we may come up with) should be purely additive. Long term, we can look at retiring `@wire` or reimplementing its internals with signals.

AllanOricil:
> @AllanOricil
> 
> > if I have N child components on the same level (siblings), all subscribed to the same signal, would there exist a way to determine the order components are updated and re-rendered?
> 
> This is a good question. Without deliberately implementing something else, the default behavior would rely on the order that components had subscribed to the signal. Is there a specific scenario you're thinking through where you'd want some other, deterministic behavior?
> 

I was thinking about a page with lots of siblings components, most of which are outside of the viewable bounds, all subscribed to the same signal. If a signal is updated, I would want to prioritize the ones that are on the view to provide a better ux. So I was thinking that maybe there could exist a reactive/dynamic index that would allow one to control the order components get updated.

AllanOricil:
I could not find a doc where solid describes its heuristics so I asked Bing gpt hehe

obs: I did not check any info showed below

"When a signal changes, Solid uses a heuristic called **topological sorting** to determine the order in which components should be re-rendered. Topological sorting is a graph algorithm that sorts nodes in a directed acyclic graph (DAG) such that for every directed edge `(u, v)`, node `u` comes before node `v` in the ordering. In the context of Solid, the nodes are components, and the edges are the dependencies between signals and components. 

The topological sorting algorithm works by first finding all nodes with no incoming edges (i.e., no dependencies), adding them to the ordering, and then removing them from the graph. This process is repeated until all nodes have been added to the ordering. If there are cycles in the graph (i.e., circular dependencies), then the algorithm will detect them and raise an error.

In summary, Solid uses topological sorting to determine the order in which components should be re-rendered when a signal changes. This ensures that components are re-rendered in the correct order and that circular dependencies are detected and handled appropriately.

Source: Conversation with Bing, 11/16/2023
(1) Using signal outside of component. · solidjs solid - GitHub. https://github.com/solidjs/solid/discussions/397.
(2) Documentation - SolidJS · Reactive Javascript Library. https://www.solidjs.com/docs/latest.
(3) Sharing Signals and Stores: Context API in SolidJS. https://www.thisdot.co/blog/sharing-signals-and-stores-context-api-in-solidjs.
(4) undefined. https://codesandbox.io/s/derivation-j0nzm?file=/index.js."


----


"When there are multiple components on the same level in the tree, Solid uses the **order in which the components are defined** to determine the order in which they should be re-rendered. 

For example, consider the following code snippet:

```jsx
function App() {
  const [count1, setCount1] = createSignal(0);
  const [count2, setCount2] = createSignal(0);

  return (
    <>
      <Counter count={count1} setCount={setCount1} />
      <Counter count={count2} setCount={setCount2} />
    </>
  );
}

function Counter({ count, setCount }) {
  return (
    <div>
      <p>Count: {count()}</p>
      <button onClick={() => setCount(count() + 1)}>Increment</button>
    </div>
  );
}
```

In this example, `App` has two `Counter` components at the same level in the tree. The order in which the components are defined in the `return` statement of `App` determines the order in which they will be re-rendered. So if `Counter` 1 is defined before `Counter` 2, then `Counter` 1 will be re-rendered before `Counter` 2 when a signal changes.

I hope that helps! Let me know if you have any other questions.

Source: Conversation with Bing, 11/16/2023
(1) Using signal outside of component. · solidjs solid - GitHub. https://github.com/solidjs/solid/discussions/397.
(2) Documentation - SolidJS · Reactive Javascript Library. https://www.solidjs.com/docs/latest.
(3) Signals: Fine-grained Reactivity for JavaScript Frameworks. https://www.sitepoint.com/signals-fine-grained-javascript-framework-reactivity/.
(4) undefined. https://codesandbox.io/s/derivation-j0nzm?file=/index.js."

AllanOricil:
I thought about a different heuristic to determine the order components react to signal changes. What if you think that the signal source is like an antenna broadcasting a change in a 3D space? Then you could order updates with these 2 rules:

- components closer to the signal source are updated first. Depth represents weight/distance of 1.
- components that are inside the viewable box have preference, but the ones with shorter distances are updated first.

You could even create the definition of single signal broadcasting multiple changes in a single event, as if each information was encoded in a different frequency, in order to avoid emitting n separate events in different times.


Is this stupid? I just made an analogy based on how telecommunications systems work

mburr-salesforce:
@AllanOricil The order in which a signal notifies its subscribers and the order in which LWC chooses to re-render components seem like orthogonal concerns? Trying to orchestrate rendering order within the scope of a single signal is going to miss many factors that could cause a component to re-render.

AllanOricil:
> @AllanOricil The order in which a signal notifies its subscribers and the order in which LWC chooses to re-render components seem like orthogonal concerns? Trying to orchestrate rendering order within the scope of a single signal is going to miss many factors that could cause a component to re-render.


I thought the order an event is received would change the order components are re-rendered. Doesn't the render engine re-render all components that have had changes during a time frame?

mburr-salesforce:
> I thought the order an event is received would change the order components are re-rendered. Doesn't the render engine re-render all components that changed in a time frame?

I'm not sure how LWC determines in what order components should be re-rendered. A few things that make me reluctant to overlay rendering concerns on a signal:

- Not all subscribers to a signal will be components. Eg we have use cases where a signal will subscribe to another signal.
- Lots of things that cause a component to re-render do not involve a signal. Eg parent changes an `@api` property, an `@wire` emits a new value, etc.
- Even when a component subscribes to a signal, a new value from the signal might not result in a re-render. Eg when the data that changed is not used in the template.

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+        this.recordUnsubscript = this.record.subscribe(record => this.updateRecord(record));
+    }
+
+    disconnectedCallback() {
+        this.recordUnsubscript();
+    }
+}
+```
+
+It is important to highlight that this is the less common method of using a wire adapter. With the new approach, it becomes a little bit more cumbersome, but it is still possible to do it, and it offers a lot more flexibility since you can now use any store that follows the protocol, not just Salesforce-specific services.
```
hsterlingsfdc, line 276: Most of the internal code I see uses wire methods, not properties.  Do we have numbers on the breakdown?

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+To this:
+
+```javascript
+import { RecordStore } from "c/stores";
+
+export default class ExampleComponent extends LightningElement {
+    @api recordId;
+    record = new RecordStore(this.$api.recordId);
+
+    updateRecord(record) {
```
hsterlingsfdc, line 262: It would be great to flesh out the error handling some more.

- Can creating a store cause an error?
- Do methods have an error object or is there a different mechanism for communicating errors to components?  Error handling frequently breaks static analysis, so if there's a way to handle them separately from the data, that would be nice.

For low-code, we are moving towards completely presentational components that do not make data calls.  The view metadata binds the data to the components, so the framework is responsible for handling errors.  You can envision an admin saying "Show this error component if a data failure occurs".

text/0000-lwc-signals.md
```diff
@@ -0,0 +1,372 @@
...
+### Store Protocol
+
+A store must implement the following protocol to be compatible with the new reactivity model:
+
+```ts
+class Store {
+    get value(): any { /* return current value */ }
+    subscribe((newValue: any) => void): () => void { /* subscribe to changes */ }
+}
+```
```
hsterlingsfdc, line 74: How can ISVs package a store?  In the past, we've asked for headless components for things like headless actions but ended up having to package it as an LWC even though it's very awkward.

We have actual use cases where we need the store to be on the side (vs in the DOM) to solve problems.  Use Case: With Islands architecture, I have a mostly static page (no app layer JS) and two small interactive regions that rely on the same state.

text/0000-lwc-signals.md
nolanlawson: > it's recommended to pass only the necessary data

Let's assume for a second that our users won't always do the "recommended" thing. (A very good assumption!)

- What kinds of guardrails can we put in place? Warnings/errors/etc.?
- What happens if they mess it up? What are the consequences?
- Knowing all this, _should_ we make it possible to do the wrong thing?

> Using the following technique there lead to signals "magically" updating their values inside the HTML elements directly, without relying/executing a full re-render of the involved components.

This is called fine-grained reactivity. It is definitely a perf boost, but I'm not convinced that signals are required to achieve it: https://github.com/salesforce/lwc/issues/3624

text/0000-lwc-signals.md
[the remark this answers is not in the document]
nolanlawson: > reensambles

?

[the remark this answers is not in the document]
nolanlawson: > `new RecordStore(this.$api.recordId)`

How does this allow the `RecordStore` to react to changes in `recordId`? It's receiving the value, not the signal, correct? Or is `$api` an object where the props are strings and the values are signals?

caridy:
@AllanOricil @nolanlawson @mburr-salesforce I have to work on splitting this into two, bare with me as I recover, and catch up with pending work.

Great questions. All I can say at the moment is that it depends on how LWC uses the signal. One of my goals is to get to the point where signals DO NOT trigger re-rendering, at least for simple interpolation, like it is the case today. @nolanlawson has been working on this as well for a while. But clearly we haven't think through all the implications here.

Today, LWC does implement an algorithm to determine the proper order to rehydrate, but if there is no re-rendering, then what?

caridy:
Folks, I have updated the PR to split the RFC into two. This now only contains the details of how to use signals in LWC templates, without adding any new API to LWC.

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+```html
+<template>
+    <button onclick={increment}>Increment</button>
+    <p>{count.value}</p>
+</template>
+```
+
+### Compiler Enhancements
+
+The LWC compiler will be updated to recognize when a Signal's `.value` is used within a template. Instead of accessing the property directly at runtime, it will generate code to subscribe to the Signal's changes during component's template initialization and unsubscribe upon component's template disposal.
```
nolanlawson, line 63: This part is a bit nebulous to me. The HTML compiler has no access to the adjacent JS file, and even if it did, it would have no way of knowing that, in `foo.value`, `foo` is a signal. So we would have to proactively wrap every `foo.value` with some runtime helper to treat it as if it _might_ be a signal.

This doesn't sound totally outlandish to me, but I'm just calling out that there's some ambiguity here.

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Motivation
+
+The current reactivity model in LWC relies solely in the observable membrane, via proxies, which are tightly coupled with LWC internals. This coupling leads to certain limitations:
+
+1. __Limited External Store Interaction__: LWC components cannot directly interact with external state management systems without going through LWC-specific APIs. This restricts the framework's flexibility and developer freedom to integrate with the broader JavaScript ecosystem.
+1. __Complexity and Overhead__: The use of a membrane based on proxies introduces additional complexity and overhead in component design and state management, from read-and-write to read-only objects, it can be challenging for developers to debug and reason about data flow and updates.
```
nolanlawson, line 21: This is mostly about cognitive overhead for developers, not performance overhead at runtime, yes?

I assume the main issue here is how DevTools represents Proxies, having to turn on DevTools custom formatters, etc.? Or is this also about tracking how changes flow through the reactivity system? (I.e. the fact that we're missing something akin to Redux DevTools?)

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+1. __Performance Constraints__: The current model relies on components to re-render fully to reflect state changes, which can be inefficient, especially for large and complex component trees.
```
nolanlawson, line 22: I see this as orthogonal, especially because for a v1, we will probably just re-render the whole template anyway (similar to the current system). We'd have to implement [fine-grained reactivity](https://github.com/salesforce/lwc/issues/3624) to change that.

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+1. __Simplified State Management__: Signals provide a straightforward way to manage reactive data. By using signals, developers can reduce complexity, making it easier to write and maintain component logic.
+1. __Performance Optimizations__: The proposed model opens up possibilities for performance optimizations by allowing selective rehydration of the UI. Components could update the DOM directly in response to state changes without a full re-render, leading to faster updates and reduced resource consumption. This has been a goal of the LWC team for a long time, small incremental improvements has been made, but this effort could accelerate the work in this area.
+
+The motivation for this RFC is to enable LWS, and developers in general, to bring stores and signal-like objects as first class citizens into the LWC ecosystem without introducing significant overhead. To improve the ergonomics of signals in LWC Template, to align LWC with more modern JavaScript practices, simplify component development, and improve performance, making LWC a developers friendly.
```
nolanlawson, line 30:
```suggestion
The motivation for this RFC is to enable LDS, and developers in general, to bring stores and signal-like objects as first class citizens into the LWC ecosystem without introducing significant overhead. To improve the ergonomics of signals in LWC Template, to align LWC with more modern JavaScript practices, simplify component development, and improve performance, making LWC a developers friendly.
```

I assume it should be LDS, not LWS?

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+    get value(): any { /* return current value */ }
+    subscribe((newValue: any) => void): () => void { /* subscribe to changes */ }
+}
+```
+
+__NOTE:__ _We can work the details of this protocol. It could be more simple, or hopefully similar to what is out there._
+
+### Integration with Components
+
+Components should treat signals as internal reactive state mechanisms. While a signal can technically be passed down to child components, it's recommended to pass only the necessary data, typically the `.value` of the signal, to keep the child components decoupled from the parent's state management strategy.
```
nolanlawson, line 80: What do you think about formally restricting this, at least in a v1? This could be relaxed later if there are valid use cases for passing down the signal itself.

(I.e. the runtime engine could detect that an object looks like a signal and was passed down wholesale. This might be a breaking change due to false positives on objects that merely look like signals, but it could be done with API versioning.)

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+
+    move(x, y) {
+        this.position.value = { ...this.position.value, x, y };
+    }
+}
+```
+
+### Connecting to External Data Stores
+
+Components can integrate with external stores by using a pattern similar to the current `@wire`` mechanism:
```
nolanlawson, line 131:
```suggestion
Components can integrate with external stores by using a pattern similar to the current `@wire` mechanism:
```

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+}
+```
+
+```html
+<template>
+    <p>{currentRecord.value.Name}</p>
+</template>
+```
+
+__Note:__ if the currentRecord's value is not available yet, it might require some additional logic to handle that case. This is similar to what developers have to do today when using `@wire`._
```
nolanlawson, line 147:
```suggestion
__Note:__ if the currentRecord's value is not available yet, it might require some additional logic to handle that case. This is similar to what developers have to do today when using `@wire`.
```

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+The LWC compiler will be updated to recognize when a Signal's `.value` is used within a template. Instead of accessing the property directly at runtime, it will generate code to subscribe to the Signal's changes during component's template initialization and unsubscribe upon component's template disposal.
+
+### Store Protocol
+
+A store must implement the following protocol to be compatible with the new reactivity model:
+
+```ts
+class Store {
+    get value(): any { /* return current value */ }
+    subscribe((newValue: any) => void): () => void { /* subscribe to changes */ }
```
nolanlawson, line 72: Isn't there a potential for memory leaks if we don't have an `unsubscribe` method as well?

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Backward Compatibility
+
+There are two main concerns with backward compatibility:
+
+1. A component receiving an object from a parent component that contains a `value` getter and a `subscribe` method. If the object is considered a POJO by our current test, then LWC would incorrectly avoid wrapping the object with a read-only reactive process. This is a minor concern, but it is something to consider.
+2. A template using `x.value` as part of an interpolation where the value of `x` is not a signal, but contains a `value` getter and a `subscribe` method. In this case, the compiled code would attempt to call `subscribe`, if it fails, we can swallow the error and show a warning. In terms of functionality, everything remains the same, the `.value` is used to render the value into the DOM, and if it is reactive it will be updated automatically by triggering the re-rendering, which would call the `.value` getter again. Technically, this is not a non-backward compabitibility issue since we can swallow the error. But it is something to consider.
```
nolanlawson, line 158: These are both solvable with API versioning IMO – if you don't bump your `<apiVersion>`, you don't get signal reactivity magic.

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Framework Compatibility Concerns
+
+* __Integration with Existing Libraries__: While the goal is to enable the use of any store that follows the Signal protocol, real-world usage may reveal compatibility issues, necessitating additional wrapper or adapter code. As of today, there are a number of popular Signal-like libraries out there, and we should make sure that we can integrate with them.
+
+## Alternatives
+
+### 1. Adoption of Other Reactivity Models
+
+Other frameworks' reactivity models, such as starbeam, Vue's reactivity system or Svelte's store contracts, could be considered. Each comes with its own set of trade-offs and would require significant adaptation to align with LWC's architecture and principles.
```
nolanlawson, line 175: I would love to see a rundown of other frameworks' reactivity systems, and if there is anything we can learn from them, or if there is an emerging standard we can rely on, etc.

nolanlawson (commented):
Overall this looks reasonable to me, but I think this RFC is still missing a competitive analysis of what other frameworks are doing and whether we want to align or diverge from them.

Signals are a huge topic right now among JS frameworks – everyone is either adopting them or looking to adopt them. Outlining what other frameworks are doing would provide a lot of useful context for this proposal, rather than trying to design in a vacuum.

text/0000-signals-in-lwc-templates.md
nolanlawson: OK, this is already covered in the "Backward Compatibility" section.

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+
+```javascript
+import { signal } from 'signals';
+
+export default class ExampleComponent extends LightningElement {
+    position = signal({ x: 0, y: 0 });
+
+    move(x, y) {
+        this.position.value = { ...this.position.value, x, y };
+    }
```
AllanOricil, line 125: What if you allow developers to configure this behavior?

```js
position = signal({x:0, y:0}) //has to change the whole object

position = signal({x:0,y:0}, { deep: true }) //deep comparison will be made which enables signaling based on prop changes
```

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+In the template, we can bind directly to the `.value` property:
+
+```html
+<template>
+    <button onclick={increment}>Increment</button>
+    <p>{count.value}</p>
+</template>
+```
```
AllanOricil, line 59: What happens if I bind with `count`? 
I really did not like to be forced to use `.value` all the time. I recall you have already answered this question but I could not find your answer.

text/0000-signals-in-lwc-templates.md
```diff
@@ -0,0 +1,196 @@
...
+
+```javascript
+export default class ChildComponent extends LightningElement {
+    @api count;
+}
+```
+
+This ensures that child components remain agnostic of the parent's state management implementation, promoting better component encapsulation and reuse.
+
+In terms of optimizations, the compiler can detect when a signal is passed down to a child component because of the `.value` in the template interpolation, and generate code to subscribe to the signal's changes. This would allow the parent component to update the child's public property `count` without a full re-render, leading to better performance. On the receiving end, because the child component's `count` property is not a signal, but a value, it will react to changes as usual, without any special handling.
```
AllanOricil, line 111: What is the use case for being able to pass the signal ref instead of `.value` to a child component? Again, as a dev I really did not like to use `count.value`. If there is no use case, I think we could just use `count`. Then the framework can identify that it is a signal based on its type (maybe you could officially make lwc ts first to get errors at build time) and that the child has to react based on `.value`

text/0000-signals-in-lwc-templates.md
AllanOricil, line 158: Couldn't you be sure that it is a signal at build time using ts? This would help the engine to do less checks at runtime

text/0000-signals-in-lwc-templates.md
AllanOricil: What if lwc provides `ref` and `unref` like vue 3?

https://markus.oberlehner.net/blog/vue-3-composition-api-ref-vs-reactive

text/0000-signals-in-lwc-templates.md
caridy: How? today you do allow passing objects, and non-plain objects as well. Do you plan to just throw if the object has a `.value` and `.subscribe`? that seems expensive, for a very little value.

text/0000-signals-in-lwc-templates.md
caridy: @nolanlawson No, the main point here is that if you choose to use `@track`, it affects your design. Now, if the child component is expecting the object, but it should have influence over the values on that object, then you need to create a channel in user-land (probably an event), so the owner of the object has to modify the read-write version of the same object. And of course, this makes it harder to debug how the data flows.

text/0000-signals-in-lwc-templates.md
caridy: well, I think the hint of a `.value` in the template, it is certainly the difference. If we were to continue doing what we are doing, we could get far, but because we don't know what is reactive and what's not reactive, I suspect we might have perf challenges.

text/0000-signals-in-lwc-templates.md
caridy: I don't think we need to. If the library implementing the signals can take care of the first point, then we should be fine. Since this RFC doesn't cover that part, we can discuss the details on the next RFC.

caridy: @AllanOricil No, the compiler compiles one file at a time in isolation.

text/0000-signals-in-lwc-templates.md
caridy: @nolanlawson certainly. I also want to get @wycats involved here.

text/0000-signals-in-lwc-templates.md
caridy: @AllanOricil all that is possible, but not part of this RFC. Remember, this is about the template, and how to use signals in the template, how to create and update signals, that's a different RFC.

As for the specific syntax in the template, you will probably need to do: `<div>{{position.value.x}}</div>`. The `.value` is the signal to listen for changes at runtime.

text/0000-signals-in-lwc-templates.md
caridy: there are 3 ways to provide stores to a component:

1. create or import a signal object
2. use a global signal (not recommended)
3. get it via a prop

So take your poison. If you have two side by side components that are using the same store provisioned by a parent component, what else can you do other than 3?

text/0000-signals-in-lwc-templates.md
caridy: I don't understand the question @nolanlawson. In our case, the template will always subscribe, and store the unsubscribe ref (which is returned by subscribe), and call it when disconnecting or replacing the template. No memory leaks, just like it happens with our own reactivity model today.

text/0000-signals-in-lwc-templates.md
caridy: the problem of not having a hint (e.g.: `.value`), then you have two options:

1. introduce a new syntax, to hint when to use it as a signal.
2. introduce a big overhead to the template runtime execution to test every binding to see if the binding is a signal (I don't think we can afford this).

text/0000-signals-in-lwc-templates.md
nolanlawson: You know what's reactive, but that doesn't mean you necessarily want fine-grained reactivity at the level of every observable property. Even Solid batches them together in some cases.

Without Signals, I think we can batch at a very coarse level, e.g. for individual items in a `for:each` iteration, and get a lot of bang for our buck.

My main point is that I wouldn't sell Signals based on perf. I'd sell it based on DX.

text/0000-signals-in-lwc-templates.md
nolanlawson: Thanks, I had not thought this through. That makes sense!

PR #87, opened 2024-01-04
[RFC]: Template class object binding

pmdartus:
> This proposal aims to improve the Developer Experience (DX) around managing components with complex styles by enabling developers to describe the classes applied to an element using JavaScript objects.

[Rendered](text/0135-template-class-object-binding.md)

POC branch: https://github.com/salesforce/lwc/tree/pmdartus/class-shorthand

### DIFF
--- a/text/0135-template-class-object-binding.md
+++ b/text/0135-template-class-object-binding.md
@@ -1,5 +1,5 @@
 ---
-title: Comprehensive Template Style
+title: Template class object binding
 status: DRAFTED
 created_at: 2024-01-04
 updated_at: 2024-01-04
@@ -7,7 +7,7 @@ champion: Pierre-Marie Dartus (pmdartus)
 pr: None
 ---
 
-# Comprehensive Template Style
+# Template class object binding
 
 ## Summary
 
--- a/text/0135-template-class-object-binding.md
+++ b/text/0135-template-class-object-binding.md
@@ -76,6 +76,7 @@ The list of class names applied to a given element is defined as follows:
 | ------------------------------------------- | --------------- |
 | `'foo bar'`                                 | `'foo bar'`     |
 | `{foo: true, bar: false, 'fiz buz': true }` | `foo fiz buz`   |
+| `{truthy: 1, falsey: '', alsoTruthy': Symbol() }` | `truthy alsoTruthy`   |
 | `[false, 'bar', null]`                      | `'bar'`         |
 | `['foo', { bar: true }, ['baz']]`           | `'foo bar baz'` |
 
--- a/text/0135-template-class-object-binding.md
+++ b/text/0135-template-class-object-binding.md
@@ -26,8 +26,8 @@ export default class CustomButton extends LightningElement {
   get className() {
     return [
       "slds-button__icon",
-      this.variant ?? `slds-button_${this.variant}`,
-      this.position ?? `slds-button_${this.position}`,
+      this.variant && `slds-button_${this.variant}`,
+      this.position && `slds-button_${this.position}`,
       {
         "slds-button_full-width": this.fullWidth,
         "slds-button_stretch": this.stretch,
@@ -68,7 +68,7 @@ This RFC proposes to change the semantics of the template `class` attribute. On 
 The list of class names applied to a given element is defined as follows:
 
 1. If the value is a `string`, apply all the classes in the value.
-2. If the value is an `object`, iterate over its keys and apply the classes with the keys associated with a truthy value.
+2. If the value is an `object`, iterate over the enumerable string properties own by the object (excluding symbols and properties on the proto-chain) and apply the classes with the keys associated with a truthy value.
 3. If the value is an `array`, iterate over its entries and apply this algorithm recursively for each of its entries.
 4. Ignore all other values: `null`, `undefined`, other primitive values, and complex objects.
 
@@ -87,7 +87,7 @@ In theory, this proposal doesn't introduce a breaking change. However, it introd
 | Input                       | Before               | After   |
 | --------------------------- | -------------------- | ------- |
 | `{ foo: true, bar: false }` | `'[object Object]'`  | `'foo'` |
-| `[false, 'foo', null]`      | `'false,'foo',null'` | `'foo'` |
+| `[false, 'foo', null]`      | `'false,'foo','` | `'foo'` |
 
 > As a side note, it's possible to match against an element with its class attribute set to `[object Object]` using the `.\005Bobject` CSS selector. 🤯
 
--- a/text/0135-template-class-object-binding.md
+++ b/text/0135-template-class-object-binding.md
@@ -39,7 +39,7 @@ export default class CustomButton extends LightningElement {
 
 ```html
 <template>
-  <button class="{className}">
+  <button class={className}>
     <slot></slot>
   </button>
 </template>
--- a/text/0135-template-class-object-binding.md
+++ b/text/0135-template-class-object-binding.md
@@ -23,14 +23,14 @@ export default class CustomButton extends LightningElement {
   fullWidth = false;
   stretch = false;
 
-  get className() {
+  get computedClassNames() {
     return [
-      "slds-button__icon",
-      this.variant && `slds-button_${this.variant}`,
-      this.position && `slds-button_${this.position}`,
+      "button__icon",
+      this.variant && `button_${this.variant}`,
+      this.position && `button_${this.position}`,
       {
-        "slds-button_full-width": this.fullWidth,
-        "slds-button_stretch": this.stretch,
+        "button_full-width": this.fullWidth,
+        "button_stretch": this.stretch,
       },
     ];
   }
@@ -39,7 +39,7 @@ export default class CustomButton extends LightningElement {
 
 ```html
 <template>
-  <button class={className}>
+  <button class={computedClassNames}>
     <slot></slot>
   </button>
 </template>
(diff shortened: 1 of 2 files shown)


text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,105 @@
...
+        "slds-button_stretch": this.stretch,
+      },
+    ];
+  }
+}
+```
+
+```html
+<template>
+  <button class="{className}">
```
abdulsattar, line 42:
```suggestion
  <button class={className}>
```

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,105 @@
...
+| Input                                       | Output          |
+| ------------------------------------------- | --------------- |
+| `'foo bar'`                                 | `'foo bar'`     |
+| `{foo: true, bar: false, 'fiz buz': true }` | `foo fiz buz`   |
+| `[false, 'bar', null]`                      | `'bar'`         |
+| `['foo', { bar: true }, ['baz']]`           | `'foo bar baz'` |
+
+**Is this a breaking change?**
+
+In theory, this proposal doesn't introduce a breaking change. However, it introduces a user-land observable change. Even if the `class` attribute only accepts `string` values today, the LWC engine doesn't warn or throw when passing a value of a different type. When passing a non-string value, the LWC engine automatically coerces the value to a string.
```
abdulsattar, line 84: IMO let's create a custom directive, `lwc:classes={className}`, like svelte does.

abdulsattar (approved):
It's a great little utility!

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,105 @@
...
+## Detailed design
+
+This RFC proposes to change the semantics of the template `class` attribute. On top of accepting a string value, the `class` attribute would now accept objects (plain objects and arrays) representing the CSS classes applied to the elements. The LWC engine would then interpret the object to determine which class names should be applied to the element.
+
+The list of class names applied to a given element is defined as follows:
+
+1. If the value is a `string`, apply all the classes in the value.
+2. If the value is an `object`, iterate over its keys and apply the classes with the keys associated with a truthy value.
+3. If the value is an `array`, iterate over its entries and apply this algorithm recursively for each of its entries.
+4. Ignore all other values: `null`, `undefined`, other primitive values, and complex objects.
```
jmsjtu, line 73: This might be covered with the "complex objects" but wanted to ask, with template expressions component authors can use complex expressions such as arrow functions.

Is that something you think the engine should handle?

For example: 
```js
<button class={() => "slds-button__icon"}>
```

text/0135-template-class-object-binding.md
ravijayaramappa: +1 for this suggestion. The `lwc:` directives behave like macros. In this case the `lwc:class` directive can add call the getter on the bound property and add the additional transformations described in the algorithm above.

ravijayaramappa (approved):
Overall, its a very intuitive improvement.

text/0135-template-class-object-binding.md
pmdartus: > IMO let's create a custom directive, `lwc:classes={className}`, like svelte does.

As a side note, Svelte's `class:name` directive can be applied to a single class name, not a set of class names. For instance:

```svelte
<button 
  class:active={isActive} 
  class:disabled={isDisabled} 
  class:brand-variant={isBrandVariant}
></button>
``` 

The equivalent in Vue would be:

```vue
<button class={{ 
    active: isActive, 
    disabled: isDisabled, 
    "brand-variant": isBrandVariant
}}></button>
```

It is also worth pointing out that Svelte's syntax doesn't enable expressing dynamic class names.

That said @abdulsattar original question is still valid. Should we use a different directive than the standard `class` attribute for this feature? 

Outside the cognitive overhead of introducing a new directive, the interaction between the standard `class` attribute and the `lwc:class` attribute isn't really obvious. Vue established a precedence with overloading the `class` attribute semantic and developers appear to be happy with this decision. 

Given the fact that this feature will be conditionally enabled using API versioning, it isn't unreasonable for LWC to follow the same approach.

text/0135-template-class-object-binding.md
pmdartus: Great question, your example is equivalent to:

```js
  className = () =>  "slds-button__icon";
}
```

```html
<button class={className}>
```

Since the object bound to the `class` attributes isn't a string, an array, nor a plain object the generated class name ``. In the current version of the engine, the class name will be `() => &quot;slds-button__icon&quot;`  (coercing the function object to string).

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Detailed design
+
+This RFC proposes to change the semantics of the template `class` attribute. On top of accepting a string value, the `class` attribute would now accept objects (plain objects and arrays) representing the CSS classes applied to the elements. The LWC engine would then interpret the object to determine which class names should be applied to the element.
+
+The list of class names applied to a given element is defined as follows:
+
+1. If the value is a `string`, apply all the classes in the value.
+2. If the value is an `object`, iterate over its keys and apply the classes with the keys associated with a truthy value.
+3. If the value is an `array`, iterate over its entries and apply this algorithm recursively for each of its entries.
```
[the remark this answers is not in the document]
pmdartus: It's a very good question. The algo currently iterates over all the enumerable string properties of the object using a `for ... in`, ignoring properties keyed as symbols. The algo doesn't include properties attached to the proto-chain (only properties owned by the object itself). 

Let me update the RFC to clarify this.

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,105 @@
...
+| `['foo', { bar: true }, ['baz']]`           | `'foo bar baz'` |
+
+**Is this a breaking change?**
+
+In theory, this proposal doesn't introduce a breaking change. However, it introduces a user-land observable change. Even if the `class` attribute only accepts `string` values today, the LWC engine doesn't warn or throw when passing a value of a different type. When passing a non-string value, the LWC engine automatically coerces the value to a string.
+
+| Input                       | Before               | After   |
+| --------------------------- | -------------------- | ------- |
+| `{ foo: true, bar: false }` | `'[object Object]'`  | `'foo'` |
+| `[false, 'foo', null]`      | `'false,'foo',null'` | `'foo'` |
```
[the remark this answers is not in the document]
pmdartus: Good catch. It's the former.

kevinv11n (approved):

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,106 @@
...
+
+```js
+export default class CustomButton extends LightningElement {
+  variant = null;
+  position = null;
+
+  fullWidth = false;
+  stretch = false;
+
+  get className() {
```
nolanlawson, line 26: This property doesn't need to be called `className`, correct? This example may be a bit confusing, since it may look like it's important to override `Element.prototype.className`.

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,106 @@
...
+    return [
+      "slds-button__icon",
+      this.variant && `slds-button_${this.variant}`,
+      this.position && `slds-button_${this.position}`,
+      {
+        "slds-button_full-width": this.fullWidth,
+        "slds-button_stretch": this.stretch,
+      },
```
nolanlawson, line 34: Recommending SLDS classes that only work in synthetic shadow DOM may not be the best example here. 🙂

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Detailed design
+
+This RFC proposes to change the semantics of the template `class` attribute. On top of accepting a string value, the `class` attribute would now accept objects (plain objects and arrays) representing the CSS classes applied to the elements. The LWC engine would then interpret the object to determine which class names should be applied to the element.
+
+The list of class names applied to a given element is defined as follows:
+
+1. If the value is a `string`, apply all the classes in the value.
+2. If the value is an `object`, iterate over the enumerable string properties own by the object (excluding symbols and properties on the proto-chain) and apply the classes with the keys associated with a truthy value.
+3. If the value is an `array`, iterate over its entries and apply this algorithm recursively for each of its entries.
```
nolanlawson, line 72: I'd suggest switching 2 and 3 here, since an `Array` is an object. (`typeof foo === 'object'`)

text/0135-template-class-object-binding.md
```diff
@@ -0,0 +1,106 @@
...
+4. Ignore all other values: `null`, `undefined`, other primitive values, and complex objects.
+
+| Input                                       | Output          |
+| ------------------------------------------- | --------------- |
+| `'foo bar'`                                 | `'foo bar'`     |
+| `{foo: true, bar: false, 'fiz buz': true }` | `foo fiz buz`   |
+| `{truthy: 1, falsey: '', alsoTruthy': Symbol() }` | `truthy alsoTruthy`   |
+| `[false, 'bar', null]`                      | `'bar'`         |
+| `['foo', { bar: true }, ['baz']]`           | `'foo bar baz'` |
```
nolanlawson, line 81: How valuable is it to allow recursion? It does introduce additional complexity into the implementation.

nolanlawson: Question answered: `classnames` in React is recursive, plus if you are mixing arrays and objects as in the first example, it's nicer to be able to put constant values in the array and non-constant values in the object.

caridy:
I don't have a strong opinion about this particular proposal. I do remember we looked at something like this in the past, and we kinda decided not to pursue it, don't remember the details from back then.

Now, if all goes well, we favor local selectors without the need for classes, in fact the new SDS architecture doesn't use classes anymore, so I expect this to be a less of a problem in the future.

nolanlawson (approved):

PR #88, opened 2024-02-28
rfc: declarative binding for non-standard event names

ekashida:
[Rendered](text/0135-template-class-object-binding.md)


text/0000-declarative-nonstandard-event-binding.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+These are examples of some of the options that are currently on the table:
+
+```html
+<!-- Event as part of directive -->
+<third-party lwc:external lwc:on:Foo-Bar-BAZ={handleConnected}></third-party>
+
+<!-- Event as part of attribute name (special treatment for lwc:external) -->
+<third-party lwc:external onFoo-Bar-BAZ={handleConnected}></third-party>
```
nolanlawson, line 46: To me, this option is the least disruptive. Yes, it is playing fast-and-loose with HTML semantics, but it is the most similar to what we have today, avoiding the need for a special new `lwc:on` directive or other mechanism.

Scoping it to `lwc:external` also brings that advantage that we can pass 100% on Custom Elements Everywhere without changing how LWC components have worked up until today. (We can re-evaluate that later if needed, but I really think it's not a big deal – the important thing is just to support the wide world of third-party custom elements and their wacky naming conventions for events.)

text/0000-declarative-nonstandard-event-binding.md
```diff
@@ -0,0 +1,125 @@
...
+// foo.js
+class Foo extends LightningElement {
+    handlers = {
+        'kebab-case': (event) => console.log("'kebab-case' handled"),
+        camelCase: (event) => console.log("'camelCase' handled"),
+        CAPScase: (event) => console.log("'CAPScase' handled"),
+        PascalCase: (event) => console.log("'PascalCase' handled"),
+    };
+}
+```
```
nolanlawson, line 65: This to me is unnecessary since we already have `lwc:spread`, which works with `onclick`/`onblur`/etc. We may have to consider how `lwc:spread` would interact, though, with `lwc:external` if we decided to support the non-standard event names there.

text/0000-declarative-nonstandard-event-binding.md
```diff
@@ -0,0 +1,125 @@
...
+## Alternatives
+
+### lwc:spread
+
+An alternative design that was considered was to build this feature into the existing `lwc:spread`
+directive which currently only supports the setting of property values. In fact, `lwc:spread` can be
+used to listen for standard events like `click` or `focus` by assigning event listeners to the
+`onclick` and `onfocus` properties. However, adding support for non-standard event names would add
+complexity to `lwc:spread`. Introducing a new directive also makes it easier to restrict the usage
+of `lwc:on` to `lwc:external` components.
```
nolanlawson, line 104: It certainly adds complexity if we want to maintain the difference between `lwc:external` vs LWC components. We'd have to add validation to the client side (runtime) as well as at compile-time. Although we could just not support `lwc:spread` with non-standard event names entirely – we'd still score 100% on Custom Elements Everywhere, so the value is debatable.

text/0000-declarative-nonstandard-event-binding.md
```diff
@@ -0,0 +1,125 @@
...
+
+## lwc:on=`${eventName}:${listenerName}`
+
+Encoding both the event name and event handler name as a string using some delimiter would allow us
```
nolanlawson, line 108: I'm not a big fan of `lwc:on`, especially since Svelte v5 is actually moving [away from their existing `on:click` syntax](https://svelte-5-preview.vercel.app/docs/event-handlers) towards a syntax more like ours, since it plays nicer with their equivalent of `lwc:spread`.

nolanlawson (commented):
This is a great exploratory RFC, but I think we should pin it down to a particular proposal and evaluate the pros/cons of that. There are a few different options being weighed here, but I think the `onFoo-BarBaz` variant is the most viable, to me anyway.

text/0000-declarative-nonstandard-event-binding.md
ekashida: I agree that it's the least disruptive, but I also feel like it might be a source of confusion/surprise. The nice thing about introducing a new directive is that it's easy to understand...if you want to do this thing that you were never able to before, you have to use this new directive.

text/0000-declarative-nonstandard-event-binding.md
```diff
@@ -0,0 +1,125 @@
...
+The design will be based on the agreed upon API shape, which is still under discussion.
+
+## Drawbacks
+
+A drawback of this feature is that component authors of LWC components may be encouraged to
+implement non-standard event names that requires the use of `lwc:on` for declarative consumption.
+This is not desirable because it makes static analysis more difficult. Conversely, consumers might
+use this directive when they could just as easily have used a standard `on*` attribute in their
+template. These scenarios could be avoided, however, by restricting the usage of `lwc:on` to custom
+elements that use the `lwc:external` directive.
```
gaurav-rk9, line 93: Lwc can be parent or owner of components using `lwc:external`. And these events could reach them if `bubbles` or `composed` is `true`. Do we not want to support those cases?

PR #89, opened 2024-06-04
RFC: Native lifecycle mixed mode

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/native-lifecycle/text/0000-native-lifecycle-mixed-mode.md)

### DIFF
--- a/text/0000-native-lifecycle-mixed-mode.md
+++ b/text/0000-native-lifecycle-mixed-mode.md
@@ -1,0 +1,152 @@
+---
+title: Native lifecycle mixed mode
+status: DRAFTED
+created_at: 2024-06-04
+updated_at: 2024-06-04
+pr: https://github.com/salesforce/lwc-rfcs/pull/89
+---
+
+# Native lifecycle mixed mode
+
+## Summary
+
+LWC ships with a polyfill for two [custom element lifecycle hooks](https://developer.mozilla.org/en-US/docs/Web/API/Web_Components/Using_custom_elements#custom_element_lifecycle_callbacks): `connectedCallback` and `disconnectedCallback`. This polyfill was originally designed for IE11, but has persisted due to backwards compatibility concerns.
+
+The polyfill (otherwise known as "synthetic lifecycle") also affects component rendering and `renderedCallback`, since the LWC engine only does an initial render when `connectedCallback` is invoked.
+
+An attempt was made during the LWC v6 release (Salesforce Summer '24) to [upgrade components to native lifecycle based on API versioning](https://github.com/salesforce/lwc/releases/v6.0.0#native-lifecycle). This turned out to be non-viable, so it was rolled back.
+
+The reason it was non-viable is that mixing components running in two modes (synthetic vs native lifecycle) causes observable changes even for components that were not expecting their lifecycle callback timings to change, due to changes in their slot container. (An example of this is spelled out in detail [here](https://github.com/salesforce/lwc/issues/4249).)
+
+To resolve this, this RFC proposes a new "native lifecycle mixed mode," inspired by [mixed shadow DOM mode](https://github.com/salesforce/lwc-rfcs/blob/master/text/0115-mixed-shadow-mode.md). It would work in much the same way, allowing components to gradually opt-in to native custom element lifecycle, while also maintaining consistency within the descendant DOM tree for that component.
+
+## Basic example
+
+Opting in:
+
+```js
+export default class extends LightningElement {
+  static lifecycleSupportMode = 'native';
+}
+```
+
+Similar to shadow DOM mixed mode, you can also opt out:
+
+```js
+export default class extends LightningElement {
+  static lifecycleSupportMode = 'reset';
+}
+```
+
+Like shadow DOM mixed mode, this `'reset'` is mostly useful if you have a superclass that declares `'native'`, which you would like to override.
+
+## Motivation
+
+The synthetic lifecycle polyfill is technical debt. If LWC were designed from scratch today, it wouldn't ship with its own implementation of `connectedCallback` or `disconnectedCallback` – it would just defer to the native browser behavior.
+
+Unfortunately, due to subtle differences between native and synthetic lifecycle, and due to the Lightning Platform's backwards compatibility guarantees, it is not trivial to swap one implementation out for the other.
+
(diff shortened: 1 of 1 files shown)


text/0000-native-lifecycle-mixed-mode.md
```diff
@@ -0,0 +1,152 @@
...
+
+The reason for rule #2 is that slots in particular proved to be a source of observable changes for native-lifecycle components. It is simply the case that a slotted component can observe the lifecycle mode of the component it is being slotted into, due to connect/disconnect events triggering downstream callbacks in the tree.
+
+The reason for rule #1 is that, exactly like slotted content, children can observe the mode of their parents.
+
+So rather than trying to emulate synthetic-lifecycle-within-native-lifecycle (which is [a fool's errand](https://github.com/salesforce/lwc/issues/4249)), we should force the entire tree into native lifecycle mode so that `connectedCallback` and `disconnectedCallback` at least consistently fire.
+
+For component authors, this means that they will have to be careful when enabling native lifecycle mode. Similar to native shadow mode, they will have to ensure that their children are all native-lifecycle ready. But in addition, they must communicate to their consumers that any _slotted_ content provided by the consumer is also beholden to this rule.
```
ekashida, line 97: This is the part that really makes me hesitate, but I don't see a way around it.

text/0000-native-lifecycle-mixed-mode.md
```diff
@@ -0,0 +1,152 @@
...
+
+For components with a longstanding history and wide variety of consumers, this might not be feasible. Component authors will have to decide for themselves whether they are ready to ship their components in native lifecycle or not.
+
+> [!NOTE]
+> There is no scenario where LWC will "move" a component from one component's slot to another component's slot. Instead, it will unmount (destroy) and mount (create) a new component in this case. This means that LWC does not have to re-evaluate a component's lifecycle mode after the initial mount. However, if a component is manually removed/inserted, then all bets are off. (Component authors typically do not do this.)
+
+### Testing
+
+For Jest testing, we can follow the same model as the existing [`nativeShadow`](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/README.md#nativeshadow) config option:
```
ekashida, line 106: I'm not familiar with this feature. Does it force the component to use native shadow regardless of its `shadowSupportMode`?

text/0000-native-lifecycle-mixed-mode.md
```diff
@@ -0,0 +1,152 @@
...
+Like `shadowSupportMode`, `lifecycleSupportMode` has two possible values:
+
+- `'native'`
+- `'reset'` (default, i.e. synthetic mode)
+
+Any other value throws an error at runtime.
+
+### Polyfill behavior
+
+Similar to [`@lwc/synthetic-shadow`](https://www.npmjs.com/package/@lwc/synthetic-shadow) and [`@lwc/aria-reflection`](https://www.npmjs.com/package/@lwc/aria-reflection), we will isolate the code necessary for synthetic lifecycle into a standalone package: `@lwc/synthetic-lifecycle`. The polyfill can be loaded using simply:
```
ekashida, line 66: Did we consider moving synthetic lifecycle into `@lwc/synthetic-shadow`? Seems like a reasonable fit that wouldn't cause disruption.

text/0000-native-lifecycle-mixed-mode.md
```diff
@@ -0,0 +1,152 @@
...
+
+Similar to shadow DOM mixed mode, you can also opt out:
+
+```js
+export default class extends LightningElement {
+  static lifecycleSupportMode = 'reset';
+}
+```
+
+Like shadow DOM mixed mode, this `'reset'` is mostly useful if you have a superclass that declares `'native'`, which you would like to override.
```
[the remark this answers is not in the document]
nolanlawson: It's actually the `subclass` that would be doing the override.

text/0000-native-lifecycle-mixed-mode.md
nolanlawson: No, it causes `@lwc/synthetic-shadow` to not be loaded at all.

Our Jest plugin is kind of odd in that it _assumes_ synthetic shadow by default, but allows you to opt-in to native shadow. That's not how `@lwc/engine-dom` works by default in most OSS cases.

text/0000-native-lifecycle-mixed-mode.md
nolanlawson: I considered conflating the two, but I thought it might be best to keep them separate so we can test them separately. Same deal with `@lwc/aria-reflection` (yet another polyfill, loaded by default in LEX).

text/0000-native-lifecycle-mixed-mode.md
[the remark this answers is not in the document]
ekashida: Sorry, got confused by the terminology!

text/0000-native-lifecycle-mixed-mode.md
ekashida: Does it make sense to test them separately? Conflating the two would preclude an app from using `@lwc/synthetic-shadow` with default native lifecycle, but I wonder what the demand for such a configuration is.

nolanlawson: I guess I'm thinking that component authors may have more trouble with one than the other. In particular, it feels intuitively like it should be "easier" to move to native lifecycle than to native shadow. That said, we ideally want to move the needle on both.

text/0000-native-lifecycle-mixed-mode.md
jmsjtu: > For component authors, this means that they will have to be careful when enabling native lifecycle mode. Similar to native shadow mode, they will have to ensure that their children are all native-lifecycle ready. But in addition, they must communicate to their consumers that any _slotted_ content provided by the consumer is also beholden to this rule.

@nolanlawson Would it make sense to allow the customer to toggle the `lifecycleSupportMode` for an internal component?

I'm thinking of something like base components that have high external customer usage.

Something like this:

```html
<base-component lifecycle-support-mode='native'>
    <c-customer-component></c-customer-component>
</base-component>
```
This would allow internal components to move to native lifecycle but give customer an escape hatch in case of any incompatibilities.

The main drawback I can see is that the internal components would have to design for both synthetic and native lifecycles.

nolanlawson: No, I am thinking of it more like this:

- Internal components will probably have to support both modes simultaneously (due to downstream impact)
- External components will have more flexibility (due to low-to-zero downstream impact)

This is almost exactly the same situation we have with `shadowSupportMode` BTW.

text/0000-native-lifecycle-mixed-mode.md
nolanlawson: I thought about this – I'm not sure we can conflate the two together, because that would be potentially a breaking change for existing components that have `shadowSupportMode = 'native'`. Their timing would certainly change in native lifecycle.

PR #90, opened 2024-12-04
RFC: Stylesheet deduplication for SSR

nolanlawson:
[Rendered](https://github.com/salesforce/lwc-rfcs/blob/nolan/ssr-dedup/text/0000-ssr-stylesheet-deduplication.md)


nolanlawson:
Feedback from @InstyleVII:

- Need a game plan for transitioning to the native browser API. Make it future proof.
- Not doing this in `engine-server` is good – it makes it as a carrot to move to the new system.

nolanlawson:
Feedback from @divmain:

- Why allow specifying `href`? Why not generate it ourselves?

For flexibility basically – consumers will probably use a hash, but they may have complex paths depending on CDN etc.

nolanlawson:
TODO: need benchmark for light DOM as well.

text/0000-ssr-stylesheet-deduplication.md
```diff
@@ -0,0 +1,177 @@
...
+
+```js
+import './stylesheet.css' // auto-hoists
+```
+
+Needless to say, this is insufficient for our use case since LWC has first-class support for shadow DOM components.
+
+## Drawbacks
+
+This solution will add significant complexity both to the LWC engine (which will have to deal with supporting the new API in SSR, as well as dealing with hydration and mismatch detection on the client) and on the consumer side (which will have to do the bulk of the work of managing and hosting the stylesheets themselves).
```
khangsfdc, line 145: Can you explain why this change increases the possibility of hydration mismatch?

nolanlawson: Right now the hydration mechanism for `<style>`s is the same as any other DOM element. Whereas if we render `<link>`s then we will need to instruct hydration with something like "If you see a `<style>` with such-and-such text content (e.g. a checksum), then consider it the same as a `<link>` with such-and-such attributes."

text/0000-ssr-stylesheet-deduplication.md
```diff
@@ -0,0 +1,165 @@
...
+
+```js
+import './stylesheet.css' // auto-hoists
+```
+
+Needless to say, this is insufficient for our use case since LWC has first-class support for shadow DOM components.
+
+## Drawbacks
+
+This solution will add significant complexity both to the LWC engine (which will have to deal with supporting the new API in SSR, as well as potentially deduplicating styles client-side w.r.t the current CSR implementation) and on the consumer side (which will have to do the bulk of the work of managing and hosting the stylesheets themselves).
```
lpomerleau, line 139: > consumer side (which will have to do the bulk of the work of managing and hosting the stylesheets themselves)

Does the consumer need to add `<style>` or `<link>` tags to the base doc for each `stylesheetCache` item? Or does the `<lwc-style>` component handle the stylesheets (as long as `registerLwcStyleComponent()` is called on the client)?

text/0000-ssr-stylesheet-deduplication.md
```diff
@@ -0,0 +1,165 @@
...
+    }
+): void;
+```
+
+For the client side, consumers are expected to import the `registerLwcStyleComponent` function and call it once:
+
+```js
+import { registerLwcStyleComponent } from '@lwc/ssr-runtime'
+registerLwcStyleComponent()
+```
```
lpomerleau, line 111: What happens if `registerLwcStyleComponent()` is not called? Will all SSRed components be un-styled?

text/0000-ssr-stylesheet-deduplication.md
nolanlawson: No, the consumer does not need to do anything. LWC itself (and notably the `<lwc-style>` component) will 1) render a `<style>` the first time it sees it, and put it in a global cache, and then 2) render nothing the next time, because it can look up the `<style>` via the `style-id` attribute. I'd recommend reading [Rob Eisenberg's design](https://eisenbergeffect.medium.com/sharing-styles-in-declarative-shadow-dom-c5bf84ffd311) which I shamelessly stole.

text/0000-ssr-stylesheet-deduplication.md
nolanlawson: Correct.

khangsfdc: Should this call occur before `hydrateComponent`?

text/0000-ssr-stylesheet-deduplication.md
khangsfdc: >on the consumer side (which will have to do the bulk of the work of managing and hosting the stylesheets themselves).

I am confused about this statement relative to the above comments.

lpomerleau: +1. That statement triggered my original question in this thread.

text/0000-ssr-stylesheet-deduplication.md
nolanlawson: For sure, yes.

text/0000-ssr-stylesheet-deduplication.md
nolanlawson: Invalid sentence from the previous RFC. Removed.

nolanlawson, line 139:
```suggestion
This solution will add significant complexity both to the LWC engine (which will have to deal with supporting the new API in SSR, as well as potentially deduplicating styles client-side w.r.t the current CSR implementation) and on the consumer side (which will have to ensure they register the `<lwc-style>` component).

```

text/0000-ssr-stylesheet-deduplication.md
```diff
@@ -0,0 +1,165 @@
...
+This RFC proposes a new API to address this problem by deduplicating rendered `<style>`s into special
+vanilla web components that push deduplicated styles into `shadowRoot.adoptedStyleSheets`.
+
+## Basic example
+
+```ts
+import { renderComponent } from '@lwc/ssr-runtime'
+
+const html = await renderComponent('x-foo', Constructor, props, {
+    stylsheetCache: new Set()
```
khangsfdc, line 24:
```suggestion
    stylesheetCache: new Set()

```

text/0000-ssr-stylesheet-deduplication.md
```diff
@@ -0,0 +1,165 @@
...
+});
+```
+
+The above API allows the caller of `renderComponent` to specify a `Set` to be used for deduplicating styles. This `Set` can be pre-populated (e.g. across renders of multiple islands of components), or created fresh.
+
+After `renderComponent` is called, the `stylesheetCache` (if specified) will be populated with any stylesheets (as strings) that were encountered during rendering.
```
khangsfdc, line 30: >will be populated with any stylesheets (as strings)

Do you have an example of how the cache is pre-populated? It's not clear to me how the caller could pre-populate the cache.

divmain (approved):

PR #92, opened 2025-01-13
RFC: `lwc:on` Directive

gaurav-rk9:
This proposal adds a declarative mechanism to add a collection of event listeners to elements in an LWC template using a new directive `lwc:on`.

### DIFF
--- a/text/0000-dynamic-events.md
+++ b/text/0000-dynamic-events.md
@@ -1,0 +1,136 @@
+---
+title: `lwc:on` directive 
+status: DRAFTED
+created_at: 2025-01-13
+updated_at: 2025-01-13
+pr: (leave this empty until the PR is created)
+---
+
+# `lwc:on` Directive
+
+## Summary
+
+This proposal adds a declarative mechanism to add a collection of event listeners to elements in an LWC template using a new directive `lwc:on`.
+
+## Basic example
+
+```js
+// x/myComponent.js
+export default class MyComponent extends LightningElement {
+
+    childEventHandlers = {
+        foo: function (event) {console.log('foo');} ,
+        bar: function (event) {console.log('bar');}
+    };
+
+}
+```
+
+```html
+ <!-- x/myComponent.html -->
+<template>
+  <x-child lwc:on={childEventHandlers}></x-child>
+</template>
+```
+
+For each property in `childEventHandlers`, the new `lwc:on` directive adds an event listener to `x-child`. The event listener listens for the event specified by the property's key and uses the property's value, bound to the instance of `MyComponent`, as the event handler.
+
+## Motivation
+
+LWC's support for declarative event listeners is limited to scenarios where event name are known while authoring the owner component template. This is sufficient for most of the cases, however when using the `lwc:component` directive, it is common to require event listeners based on the current value of the constructor passed to `lwc:is`. In such cases, it is not possible to know the event name while authoring the owner component template.
+
+As an alternative, it is possible to add event listeners imperatively. However it requires the owner component to have a reference of the element on which it needs to add event listener. The owner component can have a reference of a element only after the element has been connected and hence typically only after the connectedCallback of corresponding component has terminated , making it impossible for the owner component to handle events dispatched by the connectedCallback of the corresponding component.
+
+
+## Detailed design
+
+A new directive `lwc:on` will be introduced that can be used to declaratively add a collection of event listeners whose name or event handlers may not be known while authoring the component.
+
+### Structure
+
+The `lwc:on` directive would accept an object. For each property of the object, it would add a event listener which would listen for the event specified by the property's key and handle it using the property's value bound to the owner component.
+
+### Caching
+
+#### For static components, i.e. components declared in template directly using its selector
+Since it is uncommon for event listeners to change after the start of Owner component's intial render, We can cache them to improve performance. Note that this is same as how `onevent` on template works currently. For consumers, the implication of this would be that any changes made to the object passed to `lwc:on` after the first `connectedCallback` of owner component would cause no effect.
+
+#### For dynamic components, i.e. components created using `lwc:component` directive
+Since it is uncommon for event listeners to change if the constructor passed to `lwc:is` doesn't change, We can skip patching of event listeners if the element doesn't change. For consumers, the implication of this would be that after the first `connectedCallback` of owner component, any changes made to the object passed to `lwc:on` would cause no effect until the constructor passed to `lwc:is` itself is changed.
+
+
+### Overriding
+
+```js
+// x/myComponent.js
+export default class MyComponent extends LightningElement {
+
+    childEventHandlers = {
+        foo: function (event) {console.log('lwc:on');} ,
+    };
+
+    fooHandler(){
+        console.log('onfoo');
+    }
+
+}
+```
+
+```html
+ <!-- x/myComponent.html -->
+<template>
+  <x-child onfoo={fooHandler} lwc:on={childEventHandlers}></x-child>
+</template>
+```
+
+`lwc:on` will always be applied last. That means it will take precedence over whatever event listeners are declared in the template directly. In the above example, `x-child`, only one listener for event `foo` would be added and `childEventHandlers.foo` would be used for event handler .
+Additionally, there can be only one `lwc:on` directive on a element 
+
+### Event names
+
+The keys of object passed to `lwc:on` should conform to the requirement set by DOM Event Specification. There would be no other constraint on the object's keys.
+
+## Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
+- the impact on teaching people Lightning Web Components
+- integration of this feature with other existing and planned features
+- cost of migrating existing Lightning Web Components applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+
+To add : Static Analyzability.
+
+## Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+To add : lwc:spread and reflecting on* properties as event listeners
+
+## Adoption strategy
+
+If we implement this proposal, how will existing Lightning Web Components developers adopt it? Is
+this a breaking change? Can we write a codemod? Should we coordinate with
+other projects or libraries?
+
+# How we teach this
+
+What names and terminology work best for these concepts and why? How is this
+idea best presented? As a continuation of existing Lightning Web Components patterns?
+
+Would the acceptance of this proposal mean the Lightning Web Components documentation must be
+re-organized or altered? Does it change how Lightning Web Components is taught to new developers
+at any level?
+
+How should this feature be taught to existing Lightning Web Components developers?
+
+# Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
+TBD?
+
+Need Input on all parts of design.


text/0000-dynamic-events.md
```diff
@@ -0,0 +1,136 @@
...
+### Event names
+
+The keys of object passed to `lwc:on` should conform to the requirement set by DOM Event Specification. There would be no other constraint on the object's keys.
+
+## Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
```
Templarian, line 98: Yes, except when used `lwc:component` is used for dynamic component creation since the events are bound after the `connectedCallback`.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,136 @@
...
+- the impact on teaching people Lightning Web Components
+- integration of this feature with other existing and planned features
+- cost of migrating existing Lightning Web Components applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+
+To add : Static Analyzability.
```
Templarian, line 106: For Static Analysis since it's a dynamic component we would run into similar issues as we do with`lwc:component` in that we cannot easily search for what components these create in the client code.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
+---
+title: `lwc:on` directive 
+status: DRAFTED
+created_at: 2025-01-13
+updated_at: 2025-01-13
+pr: (leave this empty until the PR is created)
```
nolanlawson, line 6:
```suggestion
pr: https://github.com/salesforce/lwc-rfcs/pull/92

```

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
...
+A new directive `lwc:on` will be introduced that can be used to add a collection of event listeners whose name may not be known while authoring the component.
+
+### Structure
+
+The `lwc:on` directive would accept an object. For each property of the object, it would add a event listener which would listen for the event specified by the property's key and handle it using the property's value bound to the owner component.
+
+### Caching
+
+#### For static components, i.e. components declared in template directly using its selector
+Since it is uncommon for event listeners to change after the start of Owner component's intial render, We can cache them to improve performance. Note that this is same as how `onevent` on template works currently. For consumers, the implication of this would be that any changes made to the object passed to `lwc:on` after the first `connectedCallback` of owner component would cause no effect.
```
nolanlawson, line 56:
```suggestion
Since it is uncommon for event listeners to change after the start of the owner component's intial render, We can cache them to improve performance. Note that this is same as how `onevent` in template HTML works currently. For consumers, the implication of this would be that any changes made to the object passed to `lwc:on` after the first `connectedCallback` of owner component would cause no effect.

```

This makes sense to me. I like keeping the functional behavior the same as `onclick`, `onkeyup`, etc.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
...
+
+```html
+ <!-- x/myComponent.html -->
+<template>
+  <x-child onfoo={fooHandler} lwc:on={childEventHandlers}></x-child>
+</template>
+```
+
+`lwc:on` will always be applied last. That means it will take precedence over whatever event listeners are declared in the template directly. In the above example, `x-child`, only one listener for event `foo` would be added and `childEventHandlers.foo` would be used for event handler .
+Additionally, there can be only one `lwc:on` directive on a element
```
nolanlawson, line 87:
```suggestion
Additionally, there can be only one `lwc:on` directive on an element 

```

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+```html
+ <!-- x/myComponent.html -->
+<template>
+  <x-child onfoo={fooHandler} lwc:on={childEventHandlers}></x-child>
+</template>
+```
+
+`lwc:on` will always be applied last. That means it will take precedence over whatever event listeners are declared in the template directly. In the above example, `x-child`, only one listener for event `foo` would be added and `childEventHandlers.foo` would be used for event handler .
```
nolanlawson, line 86: This makes sense to me. If something is dynamically applied, it "feels" like it should take precedence. So `onfoo` is essentially the "default" listener.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Basic example
+
+```js
+// x/myComponent.js
+export default class MyComponent extends LightningElement {
+
+    childEventHandlers = {
+        foo: function (event) {console.log('foo');} ,
+        bar: function (event) {console.log('bar');}
```
nolanlawson, line 23: It would be great if one of these examples used `this`, to demonstrate that the listeners are bound to the component instance.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
...
+  <x-child onfoo={fooHandler} lwc:on={childEventHandlers}></x-child>
+</template>
+```
+
+`lwc:on` will always be applied last. That means it will take precedence over whatever event listeners are declared in the template directly. In the above example, `x-child`, only one listener for event `foo` would be added and `childEventHandlers.foo` would be used for event handler .
+Additionally, there can be only one `lwc:on` directive on a element 
+
+### Event names
+
+The keys of object passed to `lwc:on` should conform to the requirement set by DOM Event Specification. There would be no other constraint on the object's keys.
```
nolanlawson, line 91: Currently LWC throws a compile-time error if you try to use `onfooBar` or `onfoo-bar` or other disallowed naming patterns. Are you proposing that we throw an error at runtime? Or warn? Or something else?

FWIW this was actually a point of controversy in the [Custom Elements Everywhere](https://custom-elements-everywhere.com/libraries/lwc/results/results.html) benchmark, which LWC does not fully pass because we don't support these event names.

I think there's a case to be made that we should _not_ be so strict here, and should just accept whatever event string name the user passes in. (Except maybe nonsensical values like the empty string `''`, non-strings, etc.) There are cases of web components "in the wild" that use odd naming conventions for their event names, and it would be nice to support that in `lwc:on`.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,146 @@
...
+
+Currently `lwc:spread` can be used to listen for standard events like `click` or `focus` by assigning event handlers to the `onclick` and `onfocus` properties. This works because the element on which these properties are applied inherits from HTMLElement. This implies that the event handlers would be bound to these elements itself and not the owner component as `onevent` directly on template does. Without `lwc:spread`, there is not a completely equivalent way of doing this. So modifying this behaviour would break components with no straightforward workaround. However it must also be noted that this difference is also a frequent source of confusion.
+
+Implementing of this design would cause significant degradation of performance, since we would need to parse the object passed to `lwc:spread` in each render.
```
nolanlawson, line 122: I doubt it would be significant, FWIW.

In my mind, the best argument against `lwc:spread` is 1) it's a potentially breaking change, and 2) its a potential point of confusion. `lwc:on` is much cleaner in my opinion because the intention is clear.

nolanlawson (approved):
This proposal actually looks great to me. I think using `lwc:on` is much cleaner than the previous `lwc:spread`-based proposal.

One edge case not covered: are we planning to just use a simple `Object.keys()` on the supplied object? Or `Object.getOwnPropertyNames`? This matters if someone passes in e.g. a subclass that inherits properties from a superclass.

Another thing missing from this document is a review of what other frameworks are doing – how is Svelte, Vue, Lit, Solid, React, etc. handling this case?

text/0000-dynamic-events.md
Templarian: @nolanlawson I'm fine with either approach, but was under the impression `lwc:spread` would complicate the existing compiled code. Ex: `lwc:spread={simpleProps}`,

```js
    props: {
      "name": "template",
      ...$cmp.simpleProps // <- This would need a runtime filter out .startsWith('on')
    },
```

nolanlawson: Sure, it would complicate our code, but that's not the main argument against it IMO. Yes you're right that we would need a filter there.

text/0000-dynamic-events.md
gaurav-rk9: My understanding of why this is disallowed is based on Summary of [rfc: declarative binding for non-standard event names](https://github.com/salesforce/lwc-rfcs/pull/88/files#top).
I don't think we have any of those constraints here , so in long term I don't think we must not have these restrictions.
However, I think when event names are known while authoring the component, the approach finalized in [rfc: declarative binding for non-standard event names](https://github.com/salesforce/lwc-rfcs/pull/88/files#top) should be the recommended approach due to better static analyzibility.
So maybe we should have these restrictions until [rfc: declarative binding for non-standard event names](https://github.com/salesforce/lwc-rfcs/pull/88/files#top) is implemented.

text/0000-dynamic-events.md
ekashida: Does it make sense that only one or the other would be bound to the element, when you're generally able to bind the same event more than once?
```html
<template>
    <x-counter lwc:ref="counter" onfoo={handleFoo}></x-counter>
</template>
```
```js
  handleFoo() {
    console.log('foo static')
  }
  renderedCallback() {
    this.refs.counter.addEventListener('foo', () => {
      console.log('foo dynamic');
    });
    this.refs.counter.dispatchEvent(new CustomEvent('foo'));
  }
}
```
```bash
> foo static
> foo dynamic
```

text/0000-dynamic-events.md
ekashida: I agree that we should just support whatever the spec supports (i.e., [DOMString](https://webidl.spec.whatwg.org/#idl-DOMString)s), which is any valid string. I think the only thing we'd need to look out for are Symbols, as the [spec](https://tc39.es/ecma262/#sec-object-type) allows either a string or a symbol as the key in an object.

Could you mention that we'll filter out Symbols?

text/0000-dynamic-events.md
gaurav-rk9: My thoughts on this were 

- Since we don't provide consumers with a way to explicitly remove these listeners, To execute multiple functions when an event is recieved, One can create a wrapper handler over these functions with added guarantee of order.
- [debatable] Fallback event listener seems to be of more value than an additional unconditional event listener
- This would be consistent with how `lwc:spread` works.
- This directive should not be a way for having multiple event listeners. If at some point we want to support multiple declarative event listeners, we should have a clean way to do that and it should not be a side effect of something else.

text/0000-dynamic-events.md
gaurav-rk9: Updated this section. I am not sure what qualifies as significant, I have removed it for now.

text/0000-dynamic-events.md
nolanlawson: > This would be consistent with how `lwc:spread` works.

This makes sense to me. Under the hood, it will be implemented similar to how `lwc:spread` is implemented, i.e. `lwc:spread` is:

```js
props: {
  foo: $cmp.foo,
  ...$cmp.someObject
}
```

and `lwc:on` would be:

```js
on: {
  foo: $cmp.foo,
  ...$cmp.someObject
}
```

(I'm skipping the binding and caching, but you get the idea.)

wjhsf: My initial thought was that `onfoo={handleFoo}` and `lwc:on={ /* foo: handleFoo */ }` should allow both, because having multiple event handlers is conceptually fine. But it's also reasonable to say "there's just one way of doing things, don't mix and match", to avoid user confusion and combinatorial complexity.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,151 @@
...
+A new directive `lwc:on` will be introduced that can be used to add a collection of event listeners whose name may not be known while authoring the component.
+
+### Structure
+
+The `lwc:on` directive would accept an object. For each property of the object, it would add a event listener which would listen for the event specified by the property's key and handle it using the property's value bound to the owner component.
+
+### Caching
+
+#### For static components, i.e. components declared in template directly using its selector
+Since it is uncommon for event listeners to change after the start of the owner component's intial render, We can cache them to improve performance. Note that this is same as how `onevent` in template HTML works currently. For consumers, the implication of this would be that any changes made to the object passed to `lwc:on` after the first `connectedCallback` of owner component would cause no effect.
```
wjhsf, line 57: So this is essentially just sugar for doing `addEventListener` once and moving on? We wouldn't care/explode if a user did something weird like
```js
{
  foo: function handleFoo() {
    this.removeEventListener('foo', handleFoo)
  }
}
```

gaurav-rk9: Yes, this would do `addEventListener` once immediately after construction and move on. We don't care what the function does, we take the function input, wrap it into a new function that binds it to the owner componentand and then add it as event handler to the element.

``` 
{
  foo: function handleFoo() {
    this.removeEventListener('foo', handleFoo)
  }
}
```

Here if the expectation is to remove the event listener added for event `foo`, this won't work as expected, because the event handler added won't be `handleFoo`, it will be a different function, whose reference the consumer doesn't have.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,151 @@
...
+
+Although properties and event listeners use the same HTML attribute syntax, consumers understand their differences. They recognize whether a variable in an HTML template is a property, attribute, or event listener and reason about them separately. Combining them into a single directive does not enhance the developer experience. In fact, a combined directive might lead consumers to treat them as a unified concept, causing confusion. Additionally, the usage patterns of properties and event listeners are different, and a combined directive would complicate consumer code.
```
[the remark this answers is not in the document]
gaurav-rk9: Well, we wouldn't want to confuse them more.

text/0000-dynamic-events.md
Templarian: Removing the event handler dynamically placed on the host doesn't seem like it is a required feature for these once attached.

text/0000-dynamic-events.md
gaurav-rk9: currently , when consumers use `addEventListener`, (https://github.com/salesforce/lwc/blob/8d3bc0848e77fb1a2f74019d570dbfb684010c0c/packages/%40lwc/engine-core/src/framework/restrictions.ts#L182, https://github.com/salesforce/lwc/blob/8d3bc0848e77fb1a2f74019d570dbfb684010c0c/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L336) , we don't do any checks on the `type`.  we just let the browser's `addEventListener` throw. 
Shouldn't we do the same here?

wjhsf: All of the object iterators (`Object.keys`, `for...in`, etc.) ignore symbols, so any naive implementation of `lwc:on` will likely implicitly ignore symbols. If we want to check for symbols, we'd have to do so explicitly. That seems unnecessary, though, because they're not valid event names and probably nobody will ever try to use them.

gaurav-rk9: I agree that the implementation would be simpler, my thought was what would be the consumer expectation be, but now that I think about it , it does seem reasonable to say only own enumerable string-keyed properties will be considered

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Motivation
+
+LWC's current support for declarative event listeners is limited to scenarios where event types are known at the time of authoring the owner component's HTML template. There is no way to declaratively add event listeners for event types that are determined dynamically.
+
+A common use case arises when using the `lwc:component` directive, where event listeners may depend on the current value of the constructor passed to `lwc:is`. In such cases, the event types cannot be known in advance when authoring the owner component's HTML template.
+
+As an alternative, it is possible to add event listeners imperatively. However, this requires the owner component to have a reference to the element on which the listener needs to be added. This reference is only available after the element has been connected, and hence, it is typically only available after the `connectedCallback` of the corresponding component has terminated. This makes it impossible for the owner component to handle events dispatched by the `connectedCallback` of the corresponding component.
+
+## Prior Art
```
gaurav-rk9, line 51: React, Vue and Solid segregates props,attributes and event listeners in runtime. Svelte does it in runtime if spread syntax is used. It is unclear to me what benefits this has over compile time segregation. Any thoughts anyone?

ekashida (approved):

Templarian (approved):
Let me know when this is implemented in a release. The work for `LightningPanel` / `LightningOverlay` updates will start in late 256.

text/0000-dynamic-events.md
ekashida: I think we forgot to cover priority of `lwc:spread` vs `lwc:on`. `lwc:spread` can bind several types of native listeners by setting their corresponding properties. For example, click handlers via `onclick: () => {}` and `lwc:on` via `click: () => {}`. What happens then?

wjhsf: If `lwc:on` is the new sanctioned way of adding a bag of event handlers, then I think it would take priority over `lwc:spread`.

ekashida: @wjhsf That makes sense, but is technically a breaking change. If we insist on having one thing override another for event binding, then I guess `lwc:spread` should take priority by virtue of seniority.

Side note: I'm still questioning whether overrides even make sense for event binding because it goes against my mental model of how events work. Supporting all three (`onclick` via the template, `onclick` via `lwc:spread`, and `click` via `lwc:on`) seems fine and the least confusing.

gaurav-rk9: I think event listeners from `lwc:on` and `lwc:spread` should not have any overriding behaviour, just like `onclick` directly on HTML Template and `lwc:spread`'s `onclick`.
if we have these overriding, it would be confusing when someone does something like `this.refs.something.onclick = someFn`

gaurav-rk9: > > This would be consistent with how `lwc:spread` works.
> 
`lwc:spread` is:
> 
> ```js
> props: {
>   foo: $cmp.foo,
>   ...$cmp.someObject
> }
> ```

One implication to note from this is that `foo`'s setter will be called only once, thereby it is observably different from properties getting applied by both, i.e. one overwriting another

gaurav-rk9: for `foo` from `lwc:on` and `onfoo` on HTML Template 
Also, Not technically necessary, overriding rather than having both would give customers a feel that it is similar to [Event Handler Properties](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Events#event_handler_properties)

wjhsf: > Supporting all three (`onclick` via the template, `onclick` via `lwc:spread`, and `click` via `lwc:on`) seems fine and the least confusing.

> I think event listeners from `lwc:on` and `lwc:spread` should not have any overriding behaviour, just like `onclick` directly on HTML Template and `lwc:spread`'s `onclick`. if we have these overriding, it would be confusing when someone does something like `this.refs.something.onclick = someFn`

I'ma just quote myself here:

> My initial thought was that `onfoo={handleFoo}` and `lwc:on={ /* foo: handleFoo */ }` should allow both, because having multiple event handlers is conceptually fine.

Adding to that, I think that if `onclick` + `lwc:spread={onclick}` attaches multiple listeners, then it will be the least confusing if `lwc:on` is additive, rather than overriding.

wjhsf: Slightly tangential question, will the event handler parameters be the same for `lwc:on` and `onclick`/`onfoo`? Will customers be able to change `onfoo={handler}` to `lwc:on={/* foo: handler */}` without other changes?

gaurav-rk9: > Will customers be able to change `onfoo={handler}` to `lwc:on={arg}`  where `arg={/* foo: handler */}` without other changes?
 
(repharsed for clarity)
If (both `handler` and `arg` are either [local](https://github.com/salesforce/lwc/blob/89c4656af2a0be3e7519324e49a1e9622eaab2ef/packages/%40lwc/template-compiler/src/codegen/codegen.ts#L599) or both are non-local) and (they are applied on non-dynamic component), then YES
else, there would be some differences when they are changed after first render. for dynamic component, it is intentional, see Caching section. for locality difference, it is a side effect of how caching would work, but since we have not had any problem with our caching statergy based on locality for `onevent`, we are assuming that people don't change their event listeners after first render, so we should be fine with the inconsistency.

behaviour for various cases are noted here - [lwc:on Test Plan](https://salesforce.quip.com/iqxsAiakVbDs)

gaurav-rk9: HTML template attributes not starting with on --> framework recommended way to statically add props  (ignore attrs for now)
`lwc:spread` --> framework recommended way to dynamically add props
HTML template attributes starting with on --> framework recommended way to statically add event listeners
`lwc:on` --> framework recommended way to dynamically add event listeners.

using `onclick` in `lwc:spread` is observably very much different than using `onclick` directly in HTML template, so  I think there should not be any expectations of `lwc:spread`'s `onclick` fititng neatly with framework recommended ways of adding event listeners.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Caching
+
+#### For static components, i.e. components declared in template directly using its selector
+Since it is uncommon for event listeners to change after the start of the owner component's intial render, We can cache them when first encountered, to improve performance. Note that this is same as how `on{eventType}` in template HTML works currently. For consumers, the implication of this would be that any changes made to the object passed to `lwc:on` after the first `connectedCallback` of owner component would cause no effect.
+
+#### For dynamic components, i.e. components created using `lwc:component` directive
+Since it is uncommon for event listeners to change if the constructor passed to `lwc:is` doesn't change, We can skip patching of event listeners if the `element` doesn't change. For consumers, the implication of this would be that after the first `connectedCallback` of owner component, any changes made to the object passed to `lwc:on` would cause no effect until the constructor passed to `lwc:is` itself is changed - at which point all existing listeners are removed and new ones are added based on the current value of object passed to `lwc:on`.
```
jye-sf, line 80: I mentioned this during the dev sync but we should definitely think through this edge case more, however uncommon it might be. Customers _will_ hit this scenario and have expectations on what should happen, and changing the event listeners when the constructor hasn't changed is potentially something at least 1 component author might want to do.

I think this overlaps with the overall sentiment of thinking through reactivity implications more.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Overriding
+
+There can be only one `lwc:on` directive on an element.  
+If `lwc:on` specifies a listener for the same event type as an event listener declared directly in the HTML template, only the listener from `lwc:on` will be applied, and the other listener will be ignored. In the example below, for `x-child`, only one listener for event type `foo` will be added and `childEventHandlers.foo` would be used for event handler.
```
jye-sf, line 85: During the dev sync, there was some concern around having the complexity of allowing both `lwc:on` and `onFoo` on a single component.

If it's possible to detect, it would be very helpful to have compile-time error if the component author implements something that is not allowed. In this case, if we can detect both `lwc:on` + `onXXX`, we would lean towards compile-time errors or even a runtime error if there's no compelling reason to support having both. This would be preferred over silently ignoring it.

That said, a potential valid reason for doing this is if the `lwc:on` object is received dynamically, but the component also has a static `fooHandler` they would also like to add to the component. There might not be any other way to address this use case without allowing both `lwc:on` and `onFoo`. Needs to be thought through a bit more.

jye-sf (commented):
Left some comments from dev sync.

Agree with Dale's assessment that most of our discussion is around the edge cases, but the core proposal is fine and we can approve it conditionally so your work is unblocked. We should think through the edge cases and concerns though before delivering the final implementation and make sure we have the right supporting documentation for this feature.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,148 @@
...
+        console.log('onfoo');
+    }
+
+}
+```
+
+```html
+ <!-- x/myComponent.html -->
+<template>
+  <x-child lwc:on={childEventHandlers} onfoo={fooHandler}></x-child>
```
divmain, line 105: As we talked about in our engineering discussion, the LWC team would like to avoid this scenario altogether. Here's what the behavior should be: if an `lwc:on` is provided on the same tag as `onfoo` or `onbar` or `onanything`, then an error is thrown at compilation. The message should be `It is not permitted to use declarative event listeners alongside lwc:on`.

In the future, we may revisit and relax the compile-time restriction. However, it seems like a component author using `lwc:on` should be able to accomplish everything they need to accomplish without also using `on*` declarative listeners. Without a compelling use case, we prefer the more restrictive approach.

text/0000-dynamic-events.md
```diff
@@ -0,0 +1,148 @@
...
+### Overriding
+
+There can be only one `lwc:on` directive on an element.  
+If `lwc:on` specifies a listener for the same event type as an event listener declared directly in the HTML template, only the listener from `lwc:on` will be applied, and the other listener will be ignored. In the example below, for `x-child`, only one listener for event type `foo` will be added and `childEventHandlers.foo` would be used for event handler.  
+
+```js
+// x/myComponent.js
+export default class MyComponent extends LightningElement {
+
+    childEventHandlers = {
```
divmain, line 91: To be addressed: can `childEventHandlers` be implemented as a getter?

- If so, I would expect as a component author that all normal LWC reactivity would work. In other words, if `childEventHandlers` depended on `otherProperty` and `otherProperty` changes, I would also expect `childEventHandlers` to be recomputed and re-bound.
- If not, if a component author attempts to use a getter with `lwc:on`, this should probably result in a runtime error. In other words, at the time when `lwc:on`'s bound value is evaluated at runtime, we determine if the referenced value is a getter & throw an error if so.

divmain (approved):
This RFC is approved, with some edge cases to be worked out as the implementation moves forward. Thanks for putting this together & for walking us through it!

gaurav-rk9:
@divmain @jye-sf 
Regarding reactivity, 
In Vue, Svelete and React, equivalents of both `onfoo={handleFoo}` and `lwc:on={eventHandlers}` are updated to match the current value in every render.
for `lwc:on` (scope of this RFC), A patching mechanism with early return can be implemented to get similar performance by moving [bind](https://github.com/salesforce/lwc/blob/7a761c483f0021f50741b9496a23093dc048d505/packages/%40lwc/engine-core/src/framework/api.ts#L552) from template evaluation to patching.

The only thing I am concerned is it will no longer be consistent with `onfoo={handleFoo}` which does not update on re-render (see [issue 5014](https://github.com/salesforce/lwc/issues/5014)). If LWC team is fine with that, I will update this RFC with specifics and implement accordingly.

text/0000-dynamic-events.md
ekashida: I'm not sure I buy the complexity argument in the other comment, but I'm supportive of starting conservatively and opening things up if there are valid use cases.

text/0000-dynamic-events.md
gaurav-rk9: What other frameworks do :
Vue - applies both event listeners
Svelte - applies the last event listener only
React - applies the last event listener only

text/0000-dynamic-events.md
gaurav-rk9: > There might not be any other way to address this use case without allowing both lwc:on and onFoo

`lwc:on={obj} onfoo={fooHandler}` with `obj={foo: otherHandler}` when both listeners applied is equivalent to `lwc:on={obj2}` where `obj2 = {foo : function(event){otherHandler.call(this,event); fooHandler.call(this,event)}}`

gaurav-rk9:
from dev sync:
When `lwc:on={eventHandlers}` is used, the object passed, i.e. `eventHandler` will be freezed, and any attempt to modify it shall throw a runtime error. However `eventHandlers` can be re-assigned to a different object, and this shall update the actual event listeners correspondingly, i.e. removal of previous event listeners and addition of new ones.

divmain:
🎉

PR #94, opened 2026-01-12
RFC: Async Boundaries <lwc:suspense>

AllanOricil:
[Rendered](https://github.com/AllanOricil/lwc-rfcs/blob/allanoricil/async-boundaries/text/0000-async-boundaries.md)

Similar to Vue 3 and React's built-in <Suspense> components, but with features I wish they had implemented, plus a workaround for Vue's implicit async setup() suspension behavior.

Svelte and Solid also have a built in suspense component.


wjhsf:
This is an interesting idea. We will discuss it internally next week.

PR #96, opened 2026-02-03
RFC: Private Methods in LWC

a-chabot:
RFC: Private Methods in LWC


text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+status: DRAFTED
+created_at: 2026-02-03
+updated_at: 2026-02-03
+pr: https://github.com/salesforce/lwc-rfcs/pull/96
+---
+
+# Private Methods for Lightning Web Components
+
+## Summary 
+The goal of this project is to enable [native private method support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) into LWC. Private methods are created by using a `#` prefix and cannot be referenced outside of the declared class. This provides secure encapsulation. Elements and methods marked as private are not part of the typical inheritance chain and thus cannot be accessed by subclasses. The focus is explicitly only private methods, not private properties due to the [challenges outlined here](https://docs.google.com/document/d/1MBiBU02ZZewwrigd4goxWGj6soV_wMZUaYkh5gKwSG0/edit?tab=t.0#heading=h.lfham11h83gz).
```
ekashida: Let's add an inline summary of the google doc since it's internal. The main emphasis should be that supporting private properties would require changing the way we implement reactivity for properties, and that this is extremely risky.

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+pr: https://github.com/salesforce/lwc-rfcs/pull/96
+---
+
+# Private Methods for Lightning Web Components
+
+## Summary 
+The goal of this project is to enable [native private method support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) into LWC. Private methods are created by using a `#` prefix and cannot be referenced outside of the declared class. This provides secure encapsulation. Elements and methods marked as private are not part of the typical inheritance chain and thus cannot be accessed by subclasses. The focus is explicitly only private methods, not private properties due to the [challenges outlined here](https://docs.google.com/document/d/1MBiBU02ZZewwrigd4goxWGj6soV_wMZUaYkh5gKwSG0/edit?tab=t.0#heading=h.lfham11h83gz). 
+
+## Basic Example
+To convert a method to be private, simply add a hash (`#`) before the method declaration:
```
ekashida: This RFC shouldn't be focused on conversion, just on what the authoring experience will be. So all we need is a blurb on how we declare private methods on the component class, which will be the same authoring experience as in native javascript (as opposed to the different options we've discussed).

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+Private methods can be referenced using dot notation: 
+```javascript
+this.publicMethod();
+this.#privateMethod();
+```
+
+## Motivation
+The largest motivation for this feature is improving our security approach. Private methods are inherently secure by design, and using them eliminates entire classes of potential component vulnerabilities.
```
ekashida: We can add some color to this section. Feel free to summarize the examples from [Base Components System Mode Security Risks](https://docs.google.com/document/d/1RVuTkiMgBpoecYgakc8d0-YVkbV-LUdjJV-0mGan4pc/edit?tab=t.0#heading=h.h3dlwt56q0xy)

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Detailed Design 
+The following design proposal is known as the “Trojan Horse Strategy”.
```
ekashida: Mention that this strategy allows us to avoid making sensitive changes to existing implementations.

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+
+Then it will go through the @babel/plugin-transform-class-properties. Since @babel/plugin-transform-class-properties does not see this as a private method, rather a regular method, it will not throw an error. Then the method name will be transformed back to match its original name, via another newly written babel transform: 
+```javascript
+#privateMethodCall() { ... }
+```
+
+This approach delegates to native browser behavior of private methods and is a backwards compatible solution. [You can find a POC PR here.](https://github.com/salesforce/lwc/pull/5477)
+
+**Section of Random Facts Relating to the Design**
+* We do not need to transform call expressions within a component, this is already functional
```
ekashida: I didn't get this part, can you explain?

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+* We are not going to implement private properties yet, just private methods
```
ekashida: Maybe we can remove this one since we should avoid implying anything about future plans for private properties :)

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+#privateMethodCall() { ... }
+```
+
+This approach delegates to native browser behavior of private methods and is a backwards compatible solution. [You can find a POC PR here.](https://github.com/salesforce/lwc/pull/5477)
+
+**Section of Random Facts Relating to the Design**
+* We do not need to transform call expressions within a component, this is already functional
+* We are not going to implement private properties yet, just private methods 
+* If we need to add private accessors (getters/setters), it would be implemented similar to private methods. For now, we are not going to implement this
+* You can’t use private methods in the template, this throws a linting error parsing the template expression
```
ekashida: Is this behavior specific to private methods, or does it apply to method invocations in general? Asking because complex template expressions (new feature) now supports method invocations within the template.

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+New LWCs would need to implement their own private methods. Existing LWCs would need to migrate their existing methods to be private. We intend to create a linter to ensure consistent use of private methods when needed. 
+
+The design spelled out above (the Trojan Horse Strategy) is backwards compatible and does not break existing LWC code. 
+
+A simple LWC example could be written for demonstration purposes.  
+
+## How We Teach This
+Internally, we are currently calling this design the “Trojan Horse Strategy”, as we are sneaking the private methods past @babel/plugin-transform-class-properties by transforming the method names. 
+
+For external audiences, documentation should be written announcing that private methods now work in LWCs. This documentation should link to the public MDN documentation, as we are implementing native JavaScript usage. We will want to make sure to call out all similarities and differences between the LWC private methods and native ones.
```
ekashida: Not sure where, but we should mention somewhere that we'll want to start off with internal usage before exposing the feature to external customers.

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Adoption Strategy
+New LWCs would need to implement their own private methods. Existing LWCs would need to migrate their existing methods to be private. We intend to create a linter to ensure consistent use of private methods when needed. 
+
+The design spelled out above (the Trojan Horse Strategy) is backwards compatible and does not break existing LWC code. 
+
+A simple LWC example could be written for demonstration purposes.  
+
+## How We Teach This
+Internally, we are currently calling this design the “Trojan Horse Strategy”, as we are sneaking the private methods past @babel/plugin-transform-class-properties by transforming the method names.
```
ekashida: This is an implementation detail that is irrelevant to downstream users.

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+The largest motivation for this feature is improving our security approach. Private methods are inherently secure by design, and using them eliminates entire classes of potential component vulnerabilities.
+
+## Detailed Design 
+The following design proposal is known as the “Trojan Horse Strategy”.
+
+If you were to try using a private method in an LWC today, it will fail with an error from [@babel/plugin-transform-class-properties](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/javascript.ts).
+
+<img width="851" height="266" alt="private method" src="https://github.com/user-attachments/assets/afbdb13f-4e26-4ccf-9a58-4b1fcd89dc64" />
+
+To implement private methods, the proposed solution involves transforming the method call both BEFORE and AFTER @babel/plugin-transform-class-properties.
```
ekashida: Please mention exactly where the before and after transformations need to happen during compilation. This is outside the scope of this document but I think we'll also need to check whether the folks over at Communities will need to reconfigure their compilation pipeline.

text/0000-private-methods.md
ekashida: We also need to mention that this transform only applies if we enable it using a compiler option.

text/0000-private-methods.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Adoption Strategy
+New LWCs would need to implement their own private methods. Existing LWCs would need to migrate their existing methods to be private. We intend to create a linter to ensure consistent use of private methods. 
+
+The design spelled out above is backwards compatible and does not break existing LWC code. It will be adopted internally before it is enabled and exposed for external customers.
+
+A simple LWC example could be written for demonstration purposes.  
+
+## How We Teach This
+For external audiences, documentation should be written announcing that private methods now work in LWCs. This documentation should link to the public MDN documentation, as we are implementing native JavaScript usage. We will want to make sure to call out all similarities and differences between the LWC private methods and native ones.
```
ekashida: I don't believe there will be any differences since we will be relying on native capabilities.
```suggestion
Documentation for private methods can simply link to the public MDN documentation, as we are enabling native JavaScript usage.
```

ekashida:
@caridy @ravijayaramappa can you take a look at this?

text/0000-private-methods.md
```diff
@@ -0,0 +1,100 @@
...
+
+Then it will go through the @babel/plugin-transform-class-properties. Since @babel/plugin-transform-class-properties does not see this as a private method, rather a regular method, it will not throw an error. Then the method name will be transformed back to match its original name, via another newly written babel transform: 
+```javascript
+#privateMethodCall() { ... }
+```
+
+**Section of Random Facts Relating to the Design**
+* Only method declarations need to be transformed. Call expressions, using dot notation, do not need to be transformed as they already work in LWCs
+* This feature is limited to implementing only private methods. It does not include not private properties or private accessors (getter/setters).
+* You can’t use private methods in the template, this throws a linting error when parsing the template expression
```
ekashida: The compiler should throw if private method syntax appears in the template. Let's explicitly call this out in the detailed design, even if this might already be the current behavior.

ekashida (approved):

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+The goal of this project is to enable [native private method support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) into LWC. Private methods are created by using a `#` prefix and cannot be referenced outside of the declared class. This provides secure encapsulation. Elements and methods marked as private are not part of the typical inheritance chain and thus cannot be accessed by subclasses. The focus is explicitly only private methods, not private properties due to challenges with reactivity in LWCs. 
+
+## Basic Example
+We declare private methods directly on the component class using standard JavaScript private class syntax (the `#` prefix). This matches the native JavaScript authoring experience and requires no framework-specific APIs.
+
+```javascript
+#privateMethod() { ... }
+```
+
+Private methods can be referenced using dot notation:
```
wjhsf, line 26: To highlight the utility of private methods, it might be good to include the rest of the `class` context, and to contrast that with an example of disallowed syntax (like `(arg) => arg.#privateMethod()`).

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Motivation
+The largest motivation for this feature is improving our security approach. Enabling private methods helps maintain the expected behavior of a component by restricting access to its internal methods. This will prevent unexpected context abuse and shadowing methods that alter the behavior of a component. Private methods are inherently secure by design, and using them eliminates entire classes of potential component vulnerabilities.
+
+## Detailed Design 
+The following design proposal delegates to native browser behavior of private methods and is a backwards compatible solution, allowing us to make sensitive changes without impacting existing impelementations. [You can find a POC PR here.](https://github.com/salesforce/lwc/pull/5477)
+
+If you were to try using a private method in an LWC today, it will fail with an error from [@babel/plugin-transform-class-properties](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/javascript.ts).
+
+<img width="851" height="266" alt="private method" src="https://github.com/user-attachments/assets/afbdb13f-4e26-4ccf-9a58-4b1fcd89dc64" />
```
wjhsf:
```suggestion
![IDE tooltip showing an error: "LWC1007: /Users/a.chabot/qaFor258/privateMethodExample.js: Class private methods are not enabled. Please add `@babel/plugin-transform-private-methods` to your configuration."](https://github.com/user-attachments/assets/afbdb13f-4e26-4ccf-9a58-4b1fcd89dc64)
```

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+To implement private methods, the proposed solution involves transforming the method call both BEFORE and AFTER @babel/plugin-transform-class-properties. Both of these transformations will occur during compilation.
+
+At the start, user code will look something like:
+```javascript
+#privateMethodCall() { ... }
+```
+
+We will transform that method declaration, via a newly written babel transform, to look like: 
+```javascript
+_internal_only_private_privateMethodCall()
```
wjhsf: If a user creates a component with a method named `_internal_only_private_privateMethodCall`, how do we distinguish between the weirdly named fake private and the temporarily renamed real private?

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+
+Then it will go through the @babel/plugin-transform-class-properties. Since @babel/plugin-transform-class-properties does not see this as a private method, rather a regular method, it will not throw an error. Then the method name will be transformed back to match its original name, via another newly written babel transform: 
+```javascript
+#privateMethodCall() { ... }
+```
+
+**Section of Random Facts Relating to the Design**
+* Only method declarations need to be transformed. Call expressions, using dot notation, do not need to be transformed as they already work in LWCs
+* This feature is limited to implementing only private methods. It does not include not private properties or private accessors (getter/setters).
+* Both of these transforms only get applied if enabled, using a compiler option
```
wjhsf: This is a net new feature, it doesn't seem likely to break any existing components. Why do we need a compiler option?

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+
+Private class members can only be accessed from within the class body itself, as defined by JavaScript's private fields specification. You cannot call `instance.#privateMethod()` from external code because it attempts to access the private method from outside the class context. Similarly, referencing `{#privateMethod}` in an LWC template violates this same encapsulation rule because the template exists outside the JavaScript class body. Both result in a SyntaxError caught before the code ever runs.
```
wjhsf, line 81: What are the existing errors for `instance.#privateMethod()` and `{#privateMethod}`. Particularly for the latter, should we change them to provide more clear guidance?

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+**Section of Random Facts Relating to the Design**
+* Only method declarations need to be transformed. Call expressions, using dot notation, do not need to be transformed as they already work in LWCs
+* This feature is limited to implementing only private methods. It does not include not private properties or private accessors (getter/setters).
+* Both of these transforms only get applied if enabled, using a compiler option
+
+Private class members can only be accessed from within the class body itself, as defined by JavaScript's private fields specification. You cannot call `instance.#privateMethod()` from external code because it attempts to access the private method from outside the class context. Similarly, referencing `{#privateMethod}` in an LWC template violates this same encapsulation rule because the template exists outside the JavaScript class body. Both result in a SyntaxError caught before the code ever runs.  
+
+## Drawbacks
+**Drawback #1:** Performance.
+This design requires AST traversal twice, once to rename the private functions to regular functions & then again to re-rename them as private functions. There is performance overhead to this AST traversal.
```
wjhsf: 🤔 Does it require separate traversals or just an extra plugin call within the single traversal?

text/0000-private-methods.md
ekashida: The idea was to dogfood it internally before exposing the feature publicly.

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+status: DRAFTED
+created_at: 2026-02-03
+updated_at: 2026-02-06
+pr: https://github.com/salesforce/lwc-rfcs/pull/96
+---
+
+# Private Methods for Lightning Web Components
+
+## Summary 
+The goal of this project is to enable [native private method support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) into LWC. Private methods are created by using a `#` prefix and cannot be referenced outside of the declared class. This provides secure encapsulation. Elements and methods marked as private are not part of the typical inheritance chain and thus cannot be accessed by subclasses. The focus is explicitly only private methods, not private properties due to challenges with reactivity in LWCs.
```
ravijayaramappa:
```suggestion
The goal of this project is to enable [native private method support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) into LWC. Private methods are created by using a `#` prefix and cannot be referenced outside of the declared class. This provides secure encapsulation. Elements and methods marked as private are not part of the typical inheritance chain and thus cannot be accessed by subclasses. The focus is explicitly only private methods, not private properties due to challenges with reactivity in LWCs ([ref](https://github.com/salesforce/lwc/issues/3537)) . 
```

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+pr: https://github.com/salesforce/lwc-rfcs/pull/96
+---
+
+# Private Methods for Lightning Web Components
+
+## Summary 
+The goal of this project is to enable [native private method support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) into LWC. Private methods are created by using a `#` prefix and cannot be referenced outside of the declared class. This provides secure encapsulation. Elements and methods marked as private are not part of the typical inheritance chain and thus cannot be accessed by subclasses. The focus is explicitly only private methods, not private properties due to challenges with reactivity in LWCs. 
+
+## Basic Example
+We declare private methods directly on the component class using standard JavaScript private class syntax (the `#` prefix). This matches the native JavaScript authoring experience and requires no framework-specific APIs.
```
ravijayaramappa:
```suggestion
We declare private methods directly on the component class using standard JavaScript private element syntax (the `#` prefix). This matches the native JavaScript authoring experience and requires no framework-specific APIs.
```

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+
+```javascript
+#privateMethod() { ... }
+```
```
ravijayaramappa, line 24:
```suggestion
```javascript
class MyComponent extends LightningElement {
    #privateMethod() { ... }
}
```
```

File: text/0000-private-methods.md [hunk 22]
```diff
@@ -0,0 +1,101 @@
...
+
+Private class members can only be accessed from within the class body itself, as defined by JavaScript's private fields specification. You cannot call `instance.#privateMethod()` from external code because it attempts to access the private method from outside the class context. Similarly, referencing `{#privateMethod}` in an LWC template violates this same encapsulation rule because the template exists outside the JavaScript class body. Both result in a SyntaxError caught before the code ever runs.  
+
+## Drawbacks
+**Drawback #1:** Performance.
+This design requires AST traversal twice, once to rename the private functions to regular functions & then again to re-rename them as private functions. There is performance overhead to this AST traversal. 
+
+**Drawback #2:** Complexity to LWC Compiler. 
+As with all expansions, this project adds complexity and technical debt to the LWC compiler. This project is implementing two new babel transforms that will need to be maintained. 
+
```
ravijayaramappa, line 90: I assume the transformation will apply to other classes that a component might have, for example quill library. Should we account for that risk exposure.

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+#privateMethod() { ... }
+```
+
+Private methods can be referenced using dot notation: 
+```javascript
+this.#privateMethod();
+```
+
+## Motivation
+The largest motivation for this feature is improving our security approach. Enabling private methods helps maintain the expected behavior of a component by restricting access to its internal methods. This will prevent unexpected context abuse and shadowing methods that alter the behavior of a component. Private methods are inherently secure by design, and using them eliminates entire classes of potential component vulnerabilities.
```
ravijayaramappa:
```suggestion
The largest motivation for this feature is improving our security approach. Enabling private methods helps maintain the declared API contract of a component by restricting access to its internal methods. This will prevent unexpected context abuse and shadowing methods that alter the behavior of a component. Private methods are inherently secure by design, and using them eliminates entire classes of potential component vulnerabilities.
```

ravijayaramappa: unexpected context abuse and shadowing methods


1. How much of a problem is this? Since this RFC is restricted to methods only.
2. What's our stance for customers (internally & externally) who have taken a dependency on existing behavior? It could be `we will follow this for net new components but cannot break existing exposed components.`

text/0000-private-methods.md
ravijayaramappa: Why not solve the problem by enabling `@babel/plugin-transform-private-methods`? I sense, there is a problem with enabling this plugin and that needs to be covered here.

@rax-it you might know why.

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+<img width="851" height="266" alt="private method" src="https://github.com/user-attachments/assets/afbdb13f-4e26-4ccf-9a58-4b1fcd89dc64" />
+
+To implement private methods, the proposed solution involves transforming the method call both BEFORE and AFTER @babel/plugin-transform-class-properties. Both of these transformations will occur during compilation.
+
+At the start, user code will look something like:
+```javascript
+#privateMethodCall() { ... }
+```
+
+We will transform that method declaration, via a newly written babel transform, to look like:
```
ravijayaramappa, line 58: Is renaming the method declaration sufficient? Or should all the invocations be renamed as well?
Per private method docs, the javascript is invalid at this point. Any tools that check for syntactic correctness of the Javascript will throw an error.

Image

text/0000-private-methods.md
ravijayaramappa: FWIW, I added `"@babel/plugin-transform-private-methods": "7.28.6",` to compiler package.json and a class with private method compiled perfectly fine.

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+_internal_only_private_privateMethodCall()
+```
+
+Then it will go through the @babel/plugin-transform-class-properties. Since @babel/plugin-transform-class-properties does not see this as a private method, rather a regular method, it will not throw an error. Then the method name will be transformed back to match its original name, via another newly written babel transform: 
+```javascript
+#privateMethodCall() { ... }
+```
+
+**Section of Random Facts Relating to the Design**
+* Only method declarations need to be transformed. Call expressions, using dot notation, do not need to be transformed as they already work in LWCs
```
ravijayaramappa: See above https://github.com/salesforce/lwc-rfcs/pull/96/changes#r2791141097

text/0000-private-methods.md
ravijayaramappa: If the compiler option is off and the compiler sees a private method in an LWC component, what happens? 
I suspect it will result in a build failure. What circumstance will the compiler option be toggled off via a gate, without requiring any other code changes to components that use private methods?

One possible use case is, enable for internal components only and block private methods for external components.

text/0000-private-methods.md
ravijayaramappa: As implemented in the [POC PR](https://github.com/salesforce/lwc/pull/5477/changes#diff-dfe9fd1dbaac1897abacaa6436f72829b722c6c2e04f29cef505f856a7d400bbR63), there is an extra manual traversal of the tree for the `Program` visitor.

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Alternatives 
+We have considered 4 other designs, which are explained below. 
+
+**Alternate Design \#1:** Implement [@babel/plugin-transform-private-methods](https://babeljs.io/docs/babel-plugin-transform-private-methods). 
+
+As you can see in the screenshot above, the linter suggests to “Please add @babel/plugin-transform-private methods to your configuration” when using private methods in an LWC. This Babel transform lets you use JavaScript private methods in environments that don’t support them yet by transforming them into older, compatible JavaScript. In our use case, the Babel transform would perform compile time validation on private properties (anything without LWC decorators, such as @wire, @api, and @track). 
+
+This design causes issues because class properties get renamed, thus becoming unusable in the template. While this solution does remove the error, it does not resolve the security issues within the components. It is also adding a polyfill for something that is already natively supported in browsers.
```
ravijayaramappa: Only private class properties seem to be renamed. Earlier in the RFC, we asserted that private fields are not accessible from template. So I am not quite gauging the issue here, other than the fact that we are polyfiling?

Image

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+**Alternate Design \#3:** Add a newly created @private decorator. 
+
+This alternative design is to introduce a new `@private` decorator for internal only components. Then use a custom babel plugin to convert `@private` annotated methods to be only accessible from within the class. This design offered the most promising path to unblocking internal teams, as well as allowing opportunities for growth to unblock external teams. However, it’s reiterating a design that is already native to JavaScript so feels repetitive. 
+
+**Alternate Design  \#4:** Use WeakMaps. 
+
+The final approach suggestion was to utilize WeakMaps within the internal components. This solution is a design at the component level, instead of at the compiler level. While component owners still need to migrate their LWCs once private methods are enabled, the overhead & boilerplate for the WeakMap implementation is much higher for component owners. 
+
+## Adoption Strategy
+New LWCs would need to implement their own private methods. Existing LWCs would need to migrate their existing methods to be private. We intend to create a linter to ensure consistent use of private methods.
```
ravijayaramappa: > ensure consistent use

Lets add detail to what that means. For example, `we will flag all class members that don't have a decorator to be designated as private members.`

text/0000-private-methods.md
```diff
@@ -0,0 +1,101 @@
...
+**Alternate Design  \#4:** Use WeakMaps. 
+
+The final approach suggestion was to utilize WeakMaps within the internal components. This solution is a design at the component level, instead of at the compiler level. While component owners still need to migrate their LWCs once private methods are enabled, the overhead & boilerplate for the WeakMap implementation is much higher for component owners. 
+
+## Adoption Strategy
+New LWCs would need to implement their own private methods. Existing LWCs would need to migrate their existing methods to be private. We intend to create a linter to ensure consistent use of private methods. 
+
+The design spelled out above is backwards compatible and does not break existing LWC code. It will be adopted internally before it is enabled and exposed for external customers.
+
+A simple LWC example could be written for demonstration purposes.
```
ravijayaramappa, line 123: Sounds like a TODO :)

ravijayaramappa (commented):
Overall seems like a good move. Had some questions about the implementation strategy.

text/0000-private-methods.md
wjhsf: I'm guessing that the current compiler can handle call expressions (`this.#privateMethod()`), and it only breaks on method declarations (`#privateMethod() { ... }`).

text/0000-private-methods.md
wjhsf: The transformation is really a two-step _non_-transformation. Private method in => private method out. If our transformations are `#methodName => __lwc__private_method_methodName => #methodName`, then it should be fairly low risk. The bigger risk is probably how our intermediate stage might interact with other babel plugins.

text/0000-private-methods.md
wjhsf: The internal spike doc covers it in more detail, but it's because of the class properties polyfill. IIRC there's a `loose` flag. When the flag is true, private methods aren't truly private, which defeats the purpose of this RFC. When the flag is false, it breaks reactivity due to differences between the polyfill and native class syntax.

While removing the polyfills would be nice, rewriting the fundamentals of reactivity was deemed too risky.

text/0000-private-methods.md
wjhsf: This is a risk, but only for plugins that are applied _between_ our two transformations. Provided we don't have any plugins that break on our intermediate invalid code, then in theory we should be okay. In practice, it may be safer to do the more complete transformation.

text/0000-private-methods.md
a-chabot: @wjhsf is correct

text/0000-private-methods.md
a-chabot: For `instance.#privateMethod()`: 
image

For `{#privateMethod}`:
image

Both warnings could definitely be improved to be more explicit.

text/0000-private-methods.md
ekashida: Feels yucky to have an intermediate compiled state where the javascript is intentionally broken...just me?

text/0000-private-methods.md
ekashida: Should we generate a unique hash and use that instead? For example: `_79054025255fb1a26e4bc422aef54eb4()`

text/0000-private-methods.md
ekashida: Makes sense. Run the transform for internal components only. No compiler option necessary.

text/0000-private-methods.md
a-chabot: I agree with @wjhsf . The intention is the transformations are as close together to avoid this, hoping to minimize the amount of time that the method names don't line up.

text/0000-private-methods.md
ekashida: > How much of a problem is this? Since this RFC is restricted to methods only.

We don't know the scope of the problem, but it will be gated with a flag within the component. 

> How much of a problem is this? Since this RFC is restricted to methods only.
What's our stance for customers (internally & externally) who have taken a dependency on existing behavior? It could be we will follow this for net new components but cannot break existing exposed components.

This will likely be a case-by-case basis. The guidance within base components will be exactly that--use private methods for net-new, but gate usage in existing. We didn't include this in the RFC since it's not directly relevant to the feature.

text/0000-private-methods.md
ravijayaramappa: Thanks for that clarification Will.
> IIRC there's a loose flag. When the flag is true, private methods aren't truly private, which defeats the purpose of this RFC.

@a-chabot I think that is an important detail that should added to Alternative Design 1 section. This^ is the reason that alternative was rejected.

text/0000-private-methods.md
a-chabot: @ekashida Do you mean when intermediate moment when the method declaration is `_internal_only_private_privateMethodCall()` but the call expression is still `this.#privateMethod()`?

text/0000-private-methods.md
a-chabot: I like keeping the `_internal_private_` but appending a hash to the end of that isn't a bad idea. 

So 
```javascript
#privateMethod()
```
becomes
```javascript
_internal_private_452424_privateMethod()
```

text/0000-private-methods.md
a-chabot: Updated here: https://github.com/salesforce/lwc-rfcs/pull/96/changes#diff-8917b893e94ae9c36d4d8eae6ebcdb7a7ffe0ec3b72e0a40e3efed011648d7e6R76

text/0000-private-methods.md
ekashida: @a-chabot and I consulted with our AI overlords and they correctly pointed out that it's a bad idea to have a transform with non-deterministic outputs because it will break incremental builds, introduce complexity around the coordination of two separate plugins, and result in debugging complexity.

text/0000-private-methods.md
ekashida:
```
There is definitely merit to that "yucky" feeling. The primary concern is the creation of an
invalid intermediate AST where the declaration and invocation are temporarily out of sync.

This leads to a few specific risks:

Plugin Conflicts: If a linter, minifier, or tree-shaker runs during that intermediate "window,"
it may see a call to #privateMethod with no corresponding declaration and either crash or
produce incorrect optimizations.

Shadowing & Collisions: Transforming to a standard string like _internal_only_... risks
shadowing existing properties or global variables. An intermediate transform might
misinterpret the intent of the renamed method and "optimize" it in a way that breaks the
link to the original source.

Technical Debt in the Pipeline: If the transforms aren't strictly atomic and encapsulated,
the compiler is essentially "lying" to the rest of the pipeline about the class structure,
making the build process fragile and difficult to debug when stack traces show "fake"
method names.
```

ekashida: These points are pretty convincing. Let's ensure that the output of the transform maintains valid structure? We could collect all the private method names when we enter `ClassBody`, and then use those in the visitors for `ClassPrivateMethod` and `MemberExpression` for mangling.

text/0000-private-methods.md
ekashida: The first one is completely self-explanatory but it's a typescript error. The javascript error ends up being the same as the second example: `SyntaxError: Private field '#bar' must be declared in an enclosing class`.

Not as user-friendly, but it's a native syntax error that is easily searchable. Assuming it would be simple to hook into the exception and provide an error message similar to the typescript error, but I'm ok with keeping it as-is.

text/0000-private-methods.md
```diff
@@ -0,0 +1,105 @@
...
+Private class members can only be accessed from within the class body itself, as defined by JavaScript's private fields specification. You cannot call `instance.#privateMethod()` from external code because it attempts to access the private method from outside the class context. Similarly, referencing `{#privateMethod}` in an LWC template violates this same encapsulation rule because the template exists outside the JavaScript class body. Both result in a SyntaxError caught before the code ever runs.  
+
+## Drawbacks
+**Drawback #1:** Performance.
+This design requires AST traversal twice, once to rename the private functions to regular functions & then again to re-rename them as private functions. There is also a manual traversal of the tree to find the `Program` visitor. There is performance overhead to this AST traversal. 
+
+**Drawback #2:** Complexity to LWC Compiler. 
+As with all expansions, this project adds complexity and technical debt to the LWC compiler. This project is implementing two new babel transforms that will need to be maintained. 
+
+## Alternatives
```
gaurav-rk9: Question: Why does `@babel/plugin-transform-class-properties` block private methods? seems arbitrary that a plugin that only transforms class "fields" complains when it sees a private "method".  If the reason does not actually affect us, can we fork `@babel/plugin-transform-class-properties` and modify for our needs?

gaurav-rk9: I haven’t evaluated the trade-offs or risks for this, just thinking out loud
can we use [decorators with accessors](https://github.com/tc39/proposal-decorators), to implement reactivity? That is the compiler adds `accessor` keyword and a decorator on all class fields, and the engine defines that decorator to observe for changes and re-render when needed.
Babel has a [plugin](https://babeljs.io/docs/babel-plugin-proposal-decorators) for transpilation until we get native support

ekashida: I guess it's a DX feature going off the assumption that you're using the plugin because your client doesn't support private features, and it doesn't make sense to transform private fields but not private methods?

wjhsf: Babel aligns its plugins with proposed language features. Class properties and private fields were separate proposals, so they were created as separate plugins.

wjhsf: Decorators were a bet that didn't pay off, and we've seen from other features (like synthetic shadow) that relying on non-native implementations of things is just a headache down the road.

ekashida: Sure, but they were both created by Babel so not sure we can rule out a DX assumption. We'd have to ask, I guess.

a-chabot: I agree with Eugene's comments above:

> you're using the plugin because your client doesn't support private features, and it doesn't make sense to transform private fields but not private methods?

text/0000-private-methods.md
a-chabot: I've add a section titled "Round Trip Validation" which should solve this problem. Let me know what you think of this `Set` solution

text/0000-private-methods.md
a-chabot: @ekashida Could you see if "Round-Trip Validation" section of using `Sets` to keep track of transformed method names is enough to feel secure about the intermediate steps?

wjhsf (approved):

text/0000-private-methods.md
ekashida: This is really a nice-to-have imo. Will leave it up to the LWC team to decide.

text/0000-private-methods.md
ekashida: Looks good. Can you add a few snapshot tests to try to prevent regressions?

ekashida (approved):
I think we can update the status and merge!

ravijayaramappa (approved):

